Issue #9, opened 2021-11-30
Crash on first login


rac-HH:
If you login without user and password, you get:

```
Warning: Undefined array key "username" in /var/www/docker-aio/php/src/Data/ConfigurationManager.php on line 22

[...]

PR #11, opened 2021-12-01
Fix the credentials page for port 8080 and improve the readme

szaimen:
Close #9
Signed-off-by: szaimen <szaimen@e.mail.de>

PR #12, opened 2021-12-01
#9 show setup page as long as no password is configured

pestotoast:
This should fix the problems users have during inital setup. This solution shows the setup page as long as the configuration does not yet contain a password.

### DIFF
--- a/php/src/Data/ConfigurationManager.php
+++ b/php/src/Data/ConfigurationManager.php
@@ -22,9 +22,12 @@ class ConfigurationManager
         return $this->GetConfig()['username'];
     }
 
-    public function GetPassword() : string {
-        return $this->GetConfig()['password'];
-    }
+    public function GetPassword() : ?string {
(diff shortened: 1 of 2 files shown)


szaimen:
Hi, many thanks for your PR! But I think that the problem was most likely happening bwcause the password could not get written to the config file (hence the config file could not get created) due to missing filesystem permissions. So I think this check is not needed and we should probably make sure during the initial startup that we can create a file in the data directory.

szaimen:
I've just added such a check in https://github.com/nextcloud/all-in-one/pull/11/commits/c3fc6817b5721e1770efaf80091bbaa2e9c6d84a inside #11 :)

szaimen:
But lets wait what @LukasReschke thinks about this PR :)

szaimen:
So thanks for the PR but this is really not needed as the password is never empty.
The problem will be fixed with #13 :)

PR #34, opened 2021-12-03
Remove trailing whitespace from initial nextcloud password

pestotoast:
Copy-pasting the initial nextcloud password did lead to an invalid password because of trailing whitespace in the template


szaimen (approved):
lgtm 👍
Thanks for the fix!

PR #42, opened 2021-12-03
Build docker images for multiple architectures

pestotoast:
Related to #28
Instead of manually building the images, this github action will build all the docker images on every push to the main branch and publish them to dockerhub. 

- Only pushes into the **development** tag (not the latest)
- Build the images for multiple architechtures: **amd64** and **arm64**
- Some builds for arm64 images don't work yet. I also have not tested on arm64 yet.
- Needs DOCKERHUB_USERNAME and DOCKERHUB_TOKEN configured as secret
- Here's how the jobs look like: https://github.com/pestotoast/all-in-one/actions/runs/1536966830
- Here's how it looks when pushed to dockerhub: https://hub.docker.com/u/agebhart

### DIFF
--- a/.github/workflows/build_images.yml
+++ b/.github/workflows/build_images.yml
@@ -2,6 +2,7 @@ name: Build docker images
 
 on:
   push:
+    branches: [ main ]
 
 jobs:
   build:
--- a/.github/workflows/build_images.yml
+++ b/.github/workflows/build_images.yml
@@ -1,0 +1,60 @@
+name: Build docker images
+
+on:
+  push:
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+
+    strategy:
+      fail-fast: false
+      matrix:
+        platform: [ 
+          {arch: 'linux/amd64',  tag: 'develop' },
+          {arch: 'linux/arm64', tag: 'develop-arm64' } ]
+        container: [
+          {name: 'aio-apache', context: './Containers/apache'},
+          {name: 'aio-borgbackup', context: './Containers/borgbackup'},
+          {name: 'aio-collabora', context: './Containers/collabora'},
+          {name: 'aio-domaincheck', context: './Containers/domaincheck'},
+          {name: 'all-in-one', context: './Containers/mastercontainer'},
+          {name: 'aio-nextcloud', context: './Containers/nextcloud'},
+          {name: 'aio-postgresql', context: './Containers/postgresql'},
+          {name: 'aio-redis', context: './Containers/redis'},
+          {name: 'aio-talk', context: './Containers/talk'},
+          {name: 'aio-watchtower', context: './Containers/watchtower'}]
+
+    steps:
+      - name: Check out the repo
+        uses: actions/checkout@v2
+
+      - name: Set up QEMU
+        uses: docker/setup-qemu-action@v1
+
+      - name: Set up Docker Buildx
+        uses: docker/setup-buildx-action@v1
+
+      - name: Login to DockerHub
+        uses: docker/login-action@v1 
+        with:
+          username: ${{ secrets.DOCKERHUB_USERNAME }}
+          password: ${{ secrets.DOCKERHUB_TOKEN }}
+      
+      - name: Docker metadata
+        id: metadata
+        uses: docker/metadata-action@v3
+        with:
+          images: |
+            ${{ secrets.DOCKERHUB_USERNAME }}/${{ matrix.container.name }}
+          tags: |
+            ${{ matrix.platform.tag }}
+
+      - name: Build docker image and push to dockerhub
+        uses: docker/build-push-action@v2
+        with:
+          context: ${{ matrix.container.context }}
+          platforms: ${{matrix.platform.arch}}
+          no-cache: true
+          push: true
+          tags: ${{ steps.metadata.outputs.tags }}
--- a/.github/workflows/build_images.yml
+++ b/.github/workflows/build_images.yml
@@ -47,7 +47,7 @@ jobs:
         uses: docker/metadata-action@v3
         with:
           images: |
-            ${{ secrets.DOCKERHUB_USERNAME }}/${{ matrix.container.name }}
+            nextcloud/${{ matrix.container.name }}
           tags: |
             ${{ matrix.platform.tag }}
 
--- a/develop.md
+++ b/develop.md
@@ -16,6 +16,7 @@ It will now also select the developer channel for all other containers automatic
 
 ## How to promote builds from develop to latest
 
+<!---
 You can use the Docker CLI to promote builds from develop to latest. Make sure to adjust:
 
 - $name
@@ -28,13 +29,25 @@ docker pull nextcloud/$AIO_NAME@sha256:$AIO_DIGEST
 docker tag nextcloud/$AIO_NAME@sha256:$AIO_DIGEST nextcloud/$AIO_NAME\:latest
 docker push nextcloud/$AIO_NAME\:latest
 ```
+--->
 
 To automatically promoted the latest develop version you can use the following script:
 
 **WARNING:** Make sure to verify that the latest develop tag is what you really want to deploy since someone could have pushed to main and created a new container in between.
 ```shell
+# x64
 export AIO_NAME=$name
 docker pull nextcloud/$AIO_NAME\:develop
 docker tag nextcloud/$AIO_NAME\:develop nextcloud/$AIO_NAME\:latest
 docker push nextcloud/$AIO_NAME\:latest
 ```
+
+**ATTENTION**: don't run the script below since the arm64 containers currently don't work!
+```shell
+# arm64 
+export AIO_NAME=$name
+docker pull nextcloud/$AIO_NAME\:develop-arm64
+docker tag nextcloud/$AIO_NAME\:develop-arm64 nextcloud/$AIO_NAME\:latest-arm64
+docker push nextcloud/$AIO_NAME\:latest-arm64
+```
+Later when the arm64 containers work, we can simply publish to latest and latest-arm64 in a rush by providing the name one time at the top of the script.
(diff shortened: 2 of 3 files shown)


szaimen:
Hi, thanks for you PR!
Do you think it is possible to build multiarch images? I think this is what we are heading towards with https://github.com/nextcloud/all-in-one/issues/28. Feel free to click on the links that you find there :)

pestotoast:
> Do you think it is possible to build multiarch images?

Yes, see [e895aaff](
https://github.com/pestotoast/all-in-one/commit/e895aaffbb490a8b09d86863c004fefe0943cee6#diff-87ee5504a3e25ac558b343724c905f2f7949e8cec3d92b9c4300bb922afa164f) for example.

I thought it's better to keep the images in seperate tags as long as some of arm64 builds still fail. If the arm64 build fails the amd64 image does not get pushed to dockerhub.

szaimen:
> I thought it's better to keep the images in seperate tags as long as some of arm64 builds still fail. If the arm64 build fails the amd64 image does not get pushed to dockerhub.

makes probably sense, yes

> Yes, see [e895aaff](https://github.com/pestotoast/all-in-one/commit/e895aaffbb490a8b09d86863c004fefe0943cee6#diff-87ee5504a3e25ac558b343724c905f2f7949e8cec3d92b9c4300bb922afa164f) for example.

Is there already a mulltiarch image that builds correctly in your repo? We will need it to investiage how the manifest at the docker hub registry looks like as we are using that as update check.

szaimen:
Here is an example for the shell:
`export TOKEN=$(curl "https://auth.docker.io/token?service=registry.docker.io&scope=repository:nextcloud/all-in-one:pull" | jq -r .token)`
`curl --head  -H "Authorization: Bearer $TOKEN" https://registry-1.docker.io/v2/nextcloud/all-in-one/manifests/develop -H 'Accept: application/vnd.docker.distribution.manifest.v2+json'`

pestotoast:
Okay I tested your script for my `aio-talk` image (mastercontainer does not build successfully yet). 
Here's how the `aio-talk `image looks on dockerhub: https://hub.docker.com/r/agebhart/aio-talk/tags The multiarch image is in the `development` tag.


```
ag@kronos ~
$ export TOKEN=$(curl "https://auth.docker.io/token?service=registry.docker.io&scope=repository:agebhart/aio-talk:pull" | jq -r .token)
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100  4371    0  4371    0     0   9664      0 --:--:-- --:--:-- --:--:--  9670
ag@kronos ~
$ curl --head -H "Authorization: Bearer $TOKEN" https://registry-1.docker.io/v2/agebhart/aio-talk/manifests/development -H 'Accept: application/vnd.docker.distribution.manifest.v2+json'
HTTP/1.1 200 OK
content-length: 2614
content-type: application/vnd.docker.distribution.manifest.v2+json
docker-content-digest: sha256:46d5eae25acdcd4f73f6fb220096d342f9cf631d52a6014e49a6df5fdf1d875f
docker-distribution-api-version: registry/2.0
etag: "sha256:46d5eae25acdcd4f73f6fb220096d342f9cf631d52a6014e49a6df5fdf1d875f"
date: Sat, 04 Dec 2021 16:13:12 GMT
strict-transport-security: max-age=31536000
ratelimit-limit: 100;w=21600
ratelimit-remaining: 82;w=21600
```

The `docker-content-digest` is the one for the amd64 version on dockerhub:
grafik
I'm not sure how we can get the checksum for the arm64 version though.

szaimen:
I'll think we need to investigate that before going foward. But there must be a way. Maybe by sending an additional header or so...

szaimen:
cc @LukasReschke as he is the Pro in finding such things out ;)

szaimen:
looks like it is possible that multiarch manifests are actually not supported by the docker registry api:
image
(from https://docs.docker.com/registry/spec/api/)

But lets see what @LukasReschke says :)

szaimen:
Okay, after thinking about this a bit more I think it is fine to have different tags for arm64 as we don't have to rework the updatecheck logic or actually no logic at all. The only small downsize is that people with arm64 will need to use a different tag but I guess that should be fine. Also being able to slowly fix the building problems is a big plus.

@LukasReschke what do you think about adding DOCKERHUB_USERNAME  and DOCKERHUB_TOKEN to the secrets? Also do you mind disabling the autobuilds on docker hub when we merge this? ;)

.github/workflows/build_images.yml
```diff
@@ -0,0 +1,61 @@
...
+        with:
+          username: ${{ secrets.DOCKERHUB_USERNAME }}
+          password: ${{ secrets.DOCKERHUB_TOKEN }}
+      
+      - name: Docker metadata
+        id: metadata
+        uses: docker/metadata-action@v3
+        with:
+          images: |
+            ${{ secrets.DOCKERHUB_USERNAME }}/${{ matrix.container.name }}
```
szaimen, line 50: I think it should be
```suggestion
            nextcloud/${{ matrix.container.name }}
```

LukasReschke:
As the Nextcloud organization membership is really broad I am a bit reluctant to add publishing secrets to this repository. Especially as I don't see a way to lockdown publishes on DockerHub just to a given tag. (e.g. `develop` and not `latest`) 

Even if we can lockdown the secret inside GitHub to a given branch using environments, I would still feel a bit uncomfortable.

I wonder if we can use Docker Content Trust here? If we had a way to sign the `latest` tag and enforce that on pulling, that would be quite nice. - https://docs.docker.com/engine/security/trust/

szaimen:
> Especially as I don't see a way to lockdown publishes on DockerHub just to a given tag. (e.g. `develop` and not `latest`)

Isn't that already done with https://github.com/nextcloud/all-in-one/pull/42/files#diff-afa3d7d08b5e9c094aec36d7fe957d9bdb65501dc49153cac56eacdf4ca1c55dR15-R16 ? :)

LukasReschke:
> Isn't that already done with https://github.com/nextcloud/all-in-one/pull/42/files#diff-afa3d7d08b5e9c094aec36d7fe957d9bdb65501dc49153cac56eacdf4ca1c55dR15-R16 ? :)

But that is only a limitation for the action, not the secret. Right? 

e.g. as a malicious person I could still push a malicious branch to extract the secret, or as a malicious admin modify the action.

szaimen:
> I wonder if we can use Docker Content Trust here? If we had a way to sign the `latest` tag and enforce that on pulling, that would be quite nice. - https://docs.docker.com/engine/security/trust/

Sounds interesting. I guess its worth investigating 👍

szaimen:
> or as a malicious admin modify the action.

I guess that would be possible, yes.

szaimen:
> > I wonder if we can use Docker Content Trust here? If we had a way to sign the `latest` tag and enforce that on pulling, that would be quite nice. - https://docs.docker.com/engine/security/trust/
> 
> Sounds interesting. I guess its worth investigating 👍

What would you recommend? only signing the `latest` releases or the develop releases as well?
Signing could than be added to the script that we use to publish from develop to latest, no?

LukasReschke:
> What would you recommend? only signing the latest releases or the develop releases as well?
> Signing could in the first case be added to the script that we use to publish from develop to latest, no?

Only sign the `latest` tag. We keep `develop` unsigned and will have the DockerHub secrets inside a GitHub environment.

Some open points here:

1. If someone gets access to the GitHub environment can they get a new signing key or delete the old one? I assume not but am not sure :)
2. There probably needs to be changes to make the management container only download signed containers for `latest`. Maybe that is a Docker config somewhere?

szaimen:
> * If someone gets access to the GitHub environment can they get a new signing key or delete the old one? I assume not but am not sure :)

We don't safe the signing keys on github, no?

> * There probably needs to be changes to make the management container only download signed containers for `latest`. Maybe that is a Docker config somewhere?

Based on https://docs.docker.com/engine/security/trust/#client-enforcement-with-docker-content-trust we need to enforce it on client side but not sure how to do that only by using the api...

szaimen:
@LukasReschke Can we maybe restrict publishing secrets to be only able to publish to a specific tag? This would probably already solve most concerns, no?

szaimen:
@LukasReschke I have another idea: what about actually adding an autobuild progress to docker hub that also automatically builds the arm64 image and pushes it to the develop-arm64 tag automatically? Would that be possible?

LukasReschke:
Conclusion after discussion with @szaimen will be that we can use GitHub Actions but will build it in the nextcloud-releases org (which will mirror this repository)

.github/workflows/build_images.yml
szaimen: Nope

.github/workflows/build_images.yml
```diff
@@ -0,0 +1,61 @@
+name: Build docker images
+
+on:
+  push:
+    branches: [ main ]
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
```
szaimen, line 9: Needs to skip if github_branch != nextcloud-releases
See https://docs.github.com/en/actions/learn-github-actions/environment-variables

.github/workflows/build_images.yml
szaimen: Yes

.github/workflows/build_images.yml
szaimen: not needed anymore

szaimen:
Hi @pestotoast we will build images in this repository using your workflow: https://github.com/nextcloud-releases/all-in-one/pull/3. I hope you don't mind that I am closing this PR because the workflow will only be present on https://github.com/nextcloud-releases/all-in-one
See https://github.com/nextcloud-releases/all-in-one/pull/3

My other changes will be ported to a different branch. Thank your for your patience and the workflow! :)

PR #46, opened 2021-12-04
added docker-compose file

AlexanderProd:
I've added a docker-compose file to run the container more easily.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -1,0 +1,18 @@
+version: "3.1"
+
+volumes:
+  nextcloud_aio_mastercontainer:
+    name: nextcloud_aio_mastercontainer
+
+services:
+  nextcloud:
+    image: nextcloud/all-in-one:latest
+    restart: unless-stopped
+    container_name: nextcloud-aio-mastercontainer
+    volumes:
+      - nextcloud_aio_mastercontainer:/mnt/docker-aio-config
+      - /var/run/docker.sock:/var/run/docker.sock:ro
+    ports:
+      - 80:80
+      - 8080:8080
+      - 8443:8443


szaimen:
Hi, thanks for your PR!
Did you test if everything works when you do a docker-compose up?

AlexanderProd:
> Hi, thanks for your PR! Did you test if everything works when you do a docker-compose up?

Yeah I've already used it.

docker-compose.yml
```diff
@@ -0,0 +1,14 @@
+version: "3.1"
+
+services:
+  nextcloud:
+    image: nextcloud/all-in-one
+    restart: unless-stopped
+    container_name: nextcloud-aio
+    volumes:
+      - ./config:/mnt/docker-aio-config
```
szaimen, line 9: This actually needs to be inside the nextcloud_aio_mastercontainer volume. otherwise the built-in borgbackup will not work.

docker-compose.yml
```diff
@@ -0,0 +1,14 @@
+version: "3.1"
+
+services:
+  nextcloud:
+    image: nextcloud/all-in-one
+    restart: unless-stopped
+    container_name: nextcloud-aio
```
szaimen, line 7: how does the resulting container name look like?
Could you post the output of docker ps here?
It needs to be nextcloud-aio-mastercontainer otherwise the built-in updater will not work.

docker-compose.yml
```diff
@@ -0,0 +1,14 @@
+version: "3.1"
+
+services:
+  nextcloud:
+    image: nextcloud/all-in-one
```
szaimen, line 5: please change it to 
```suggestion
    image: nextcloud/all-in-one:latest
```

docker-compose.yml
szaimen:
```suggestion
      - nextcloud_aio_mastercontainer:/mnt/docker-aio-config
```

szaimen: how does the resulting volume name look like if you put in this volume name?
could you please output a docker volume ls here? Thanks!

szaimen:
> Yeah I've already used it.

Great! Please adress the requested changes :)

docker-compose.yml
AlexanderProd:
```
CONTAINER ID   IMAGE                         COMMAND                  CREATED          STATUS          PORTS                                                                                          NAMES
ac3e7526ab75   nextcloud/all-in-one          "start.sh /usr/bin/s…"   5 hours ago      Up 11 seconds   0.0.0.0:8080->8080/tcp, :::8080->8080/tcp, 80/tcp, 0.0.0.0:8443->8443/tcp, :::8443->8443/tcp   nextcloud-aio
```

> It needs to be nextcloud-aio-mastercontainer otherwise the built-in updater will not work.

ah I didn't know that, container_name says how the container will be named.

AlexanderProd: I'll change it to `nextcloud-aio-mastercontainer`

docker-compose.yml
AlexanderProd: What do you mean exactly?

szaimen: the built-in backup cannot backup the important files that are usually stored in the docker volume if you don't store it in a volume ;)

szaimen: see https://github.com/nextcloud/all-in-one/blob/02365fdd3f91decdea6da174697fe18eedc13647/php/containers.json#L230-L234

pestotoast: > how does the resulting volume name look like if you put in this volume name?
could you please output a docker volume ls here?

If we define the volume as external we can make sure it has the exact name we need:
```
volumes:
    external: true
```

szaimen: > 

nice, thanks for the tip!

szaimen: @AlexanderProd please add that :)

AlexanderProd: `Volume nextcloud_aio_mastercontainer declared as external, but could not be found. Please create the volume manually using `docker volume create --name=nextcloud_aio_mastercontainer` and try again.` 
This is what I get when starting the container with `external: true`. Do we really want it to be necessary that the user has to create the volume manually?

szaimen: > Do we really want it to be necessary that the user has to create the volume manually?

Not necessarily... But definitely if there is no other way to name it like this.

pestotoast: Weird, it works on my machine:

```
ag@kronos ~
$ sudo docker volume ls
DRIVER    VOLUME NAME
ag@kronos ~
$ sudo docker-compose up -d
[+] Running 3/3
 ⠿ Network ag_default                       Create...                      0.0s
 ⠿ Volume "nextcloud_aio_mastercontainer"   Created                        0.0s
 ⠿ Container nextcloud-aio-mastercontainer  Started                        0.3s
ag@kronos ~
$ sudo docker volume ls
DRIVER    VOLUME NAME
local     nextcloud_aio_mastercontainer
ag@kronos ~
```

szaimen: @pestotoast whats the excact compose file that you've used? or what were the changes that you've made to the one in this PR?

pestotoast:
```
version: "3.1"

volumes:
    external: true

services:
  nextcloud:
    restart: unless-stopped
    container_name: nextcloud-aio-mastercontainer
    volumes:
      - nextcloud_aio_mastercontainer:/mnt/docker-aio-config
      - /var/run/docker.sock:/var/run/docker.sock:ro
    ports:
      - 80:80
      - 8080:8080
      - 8443:8443

```

AlexanderProd: Well, looks just like mine except I had the volumes defined at the end but putting them at the begging didn't change it either. Shall I just create it manually and move on? 

`
$ docker version
`
Client: Docker Engine - Community
 Version:           20.10.11
 API version:       1.41
 Go version:        go1.16.9
 Git commit:        dea9396
 Built:             Thu Nov 18 00:37:06 2021
 OS/Arch:           linux/amd64
 Context:           default
 Experimental:      true

Server: Docker Engine - Community
 Engine:
  Version:          20.10.11
  API version:      1.41 (minimum version 1.12)
  Go version:       go1.16.9
  Git commit:       847da18
  Built:            Thu Nov 18 00:35:15 2021
  OS/Arch:          linux/amd64
  Experimental:     false
 containerd:
  Version:          1.4.12
  GitCommit:        7b11cfaabd73bb80907dd23182b9347b4245eb5d
 runc:
  Version:          1.0.2
  GitCommit:        v1.0.2-0-g52b36a2
 docker-init:
  Version:          0.19.0
  GitCommit:        de40ad0

pestotoast: Does this work for you?
```
volumes:
    name: nextcloud_aio_mastercontainer
```

szaimen: as I said: if there is another way to provide this excact name for the volume, I am open to add that. But we definitely need this excact name. So if there is no other way, I guess we should just go with that also if it requires to create the volume first...

AlexanderProd: Using `name: nextcloud_aio_mastercontainer` fixed it for me.

szaimen: great 🎉

szaimen:
Thanks! 
@AlexanderProd do you mind squashing your commits? Aftewards I'll merge this :)

szaimen (approved):
Thanks! :)

PR #48, opened 2021-12-04
improve docs

### LINKED ISSUE #44: Add "Default Phone Region" to config.php
**Security & setup warnings displays the "missing default phone region" after initial install**

This can, of course, be added by changing the config.php in the specific "nextcloud/aio-nextcloud:latest"-container but maybe this parameter can also be implemented somehow, giving the user the change to adapt this value to match the user's personal needs?

szaimen:
Close #44


szaimen:
lgtm now :)

Issue #10, opened 2021-11-30
Docker socket is not readable by the www-data user.


pestotoast:
The container does not start if the docker group does not have the expected group id. This is what I did:

```
sudo docker run -it --name nextcloud-aio-mastercontainer --restart always -p 80:80 -p 8080:8080 -p 8443:8443 --volume nextcloud_aio_mastercontainer:/mnt/docker-aio-config --volume /var/run/docker.sock:/var/run/docker.sock:ro nextcloud/all-in-one:latest
Unable to find image 'nextcloud/all-in-one:latest' locally
latest: Pulling from nextcloud/all-in-one
eff15d958d66: Pull complete 
933427dc39f7: Pull complete 
35bb08dc7ee2: Pull complete 
58a3f26800d7: Pull complete 
bba5c0e912ac: Pull complete 
62e5c0d120c5: Pull complete 
410a26230a3b: Pull complete 
cfd99187016c: Pull complete 
c32ab7d5171e: Pull complete 
0d32eb2980c3: Pull complete 
991f816a61cc: Pull complete 
c8667ca62702: Pull complete 
6405e507a13e: Pull complete 
108fd50a39c4: Pull complete 
a4802c6a0ae4: Pull complete 
4f4fb700ef54: Pull complete 
ed2a8c5d896a: Pull complete 
918f8b879784: Pull complete 
ef811a9a19da: Pull complete 
5957cd578fac: Pull complete 
16271f6bc746: Pull complete 
1fad265ca45b: Pull complete 
d3cff4a42d6b: Pull complete 
ed82917488d5: Pull complete 
e306f9b356b0: Pull complete 
71afc1271985: Pull complete 
0ec4a207d9fd: Pull complete 
891138b323b3: Pull complete 
9ba4b052261a: Pull complete 
df2f9d124a4f: Pull complete 
48b90860d7f8: Pull complete 
2ae887bcd224: Pull complete 
2283bd5bb2d5: Pull complete 
Digest: sha256:9a47ed4b666f74027d710490ae1fac5a078275d2516ce8e671046cb2337fb88b
Status: Downloaded newer image for nextcloud/all-in-one:latest
Docker socket is not readable by the www-data user. Cannot continue.
```
[...]

PR #50, opened 2021-12-04
fix docker permissions, check if docker GID already exists 

pestotoast:
Should fix https://github.com/nextcloud/all-in-one/issues/10#issuecomment-985895684

Check if a group already exists that has the same groupid as `/var/run/docker.socket` and adjust permissions accordingly

Signed-off-by: Adrian Gebhart <adrian@pestotoast.de>

### DIFF
--- a/Containers/mastercontainer/start.sh
+++ b/Containers/mastercontainer/start.sh
@@ -15,19 +15,18 @@ elif ! mountpoint -q "/mnt/docker-aio-config"; then
     exit 1
 elif ! sudo -u www-data test -r /var/run/docker.sock; then
     echo "Trying to fix docker.sock permissions internally..."
-
     DOCKER_GROUP=$(stat -c '%G' /var/run/docker.sock)
     DOCKER_GROUP_ID=$(stat -c '%g' /var/run/docker.sock)
-    # check if a group with the same group id of /var/run/docker.socket already exists in the container
+    # Check if a group with the same group id of /var/run/docker.socket already exists in the container
     if grep -q "^$DOCKER_GROUP:" /etc/group; then
-            #if yes, add www-data to that group
-            echo "Adding internal www-data to group $DOCKER_GROUP"
-            usermod -aG "$DOCKER_GROUP" www-data
-        else
-            #if the group doesn't exist, create it
-            echo "Creating docker group internally with id $DOCKER_GROUP_ID"
-            groupadd -g "$DOCKER_GROUP_ID" docker
-            usermod -aG docker www-data
+        # If yes, add www-data to that group
+        echo "Adding internal www-data to group $DOCKER_GROUP"
+        usermod -aG "$DOCKER_GROUP" www-data
+    else
+        # If the group doesn't exist, create it
+        echo "Creating docker group internally with id $DOCKER_GROUP_ID"
+        groupadd -g "$DOCKER_GROUP_ID" docker
+        usermod -aG docker www-data
     fi
     if ! sudo -u www-data test -r /var/run/docker.sock; then
         echo "Docker socket is not readable by the www-data user. Cannot continue."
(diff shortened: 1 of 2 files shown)


Containers/mastercontainer/start.sh
```diff
@@ -15,9 +15,20 @@ elif ! mountpoint -q "/mnt/docker-aio-config"; then
     exit 1
 elif ! sudo -u www-data test -r /var/run/docker.sock; then
     echo "Trying to fix docker.sock permissions internally..."
-    GROUP="$(stat -c '%g' /var/run/docker.sock)"
-    groupadd -g "$GROUP" docker && \
-    usermod -aG docker www-data
+
+    # check if a group with the same group id of /var/run/docker.socket already exists in the container
+    if grep -q $(stat -c '%g' /var/run/docker.sock) /etc/group; then
```
szaimen, line 20: Shouldn't it be 
```suggestion
    if grep -q "^$(stat -c '%G' /var/run/docker.sock):" /etc/group; then
```
?
Also storing this in a varialbe makes probably sense since we need it at two places...

szaimen (commented):
Thanks! just a small comment :)

Also, we need to add the fix here:
https://github.com/nextcloud/all-in-one/blob/a1f00db3e3038c2d5380881590732bd75cad7cd3/Containers/watchtower/start.sh#L9-L11

Containers/mastercontainer/start.sh
szaimen: Also storing stat -c '%G' /var/run/docker.sock in a varialbe makes probably sense since we need it at two places...

pestotoast: Are you sure we need that? It looks to me like everything runs with root permissions in that watchtower container. So we shouldn't have to deal with permissions.

szaimen: if the root user doesn't have problems with different permissions set then we might consider actually removing the whole logic: https://github.com/nextcloud/all-in-one/blob/a1f00db3e3038c2d5380881590732bd75cad7cd3/Containers/watchtower/start.sh#L7-L16
But I didn't test it out.

szaimen: @pestotoast do you mind resolving https://github.com/nextcloud/all-in-one/pull/50#discussion_r762449071 and https://github.com/nextcloud/all-in-one/pull/50#discussion_r762449499 ? Otherwise I'd take this over :)

pestotoast: Regarding https://github.com/nextcloud/all-in-one/pull/50#discussion_r762449071: I was using `%g` instead of `%G` to be able to check for the group id instead of the group name.
I added some refactoring  regarding https://github.com/nextcloud/all-in-one/pull/50#discussion_r762449499

szaimen: Yes, but this might lead to problem since you only check if this gid is somewhere in the file.
This is not fail-proof enough, imho.

Containers/mastercontainer/start.sh
```diff
@@ -15,9 +15,20 @@ elif ! mountpoint -q "/mnt/docker-aio-config"; then
...
 elif ! sudo -u www-data test -r /var/run/docker.sock; then
     echo "Trying to fix docker.sock permissions internally..."
-    GROUP="$(stat -c '%g' /var/run/docker.sock)"
-    groupadd -g "$GROUP" docker && \
-    usermod -aG docker www-data
+
+    DOCKER_GROUP=$(stat -c '%G' /var/run/docker.sock)
+    DOCKER_GROUP_ID=$(stat -c '%g' /var/run/docker.sock)
+    # check if a group with the same group id of /var/run/docker.socket already exists in the container
+    if grep -q $DOCKER_GROUP_ID /etc/group; then
```
szaimen, line 22: Thanks for the refactoring! For fail-proof I'd recommend to use the following code:
```suggestion
    if grep -q "^$DOCKER_GROUP:" /etc/group; then
```

szaimen (approved):
lgtm now :)
Will merge and test it!

PR #51, opened 2021-12-04
Add permission workaround for Synology NAS to FAQ

Timmmy:
Add permission workaround for Synology NAS to FAQ


szaimen:
Hi, thanks for your Pull Request but I think we don't need it anymore with the fix in #50.
Feel free to pull the latest container and try to run the docker run ... command again. 
It should work now :)
If not, please let us know!

PR #61, opened 2021-12-07
remove the username for the aio interface

### LINKED ISSUE #25: Remove username
cc @LukasReschke if we need a username since we don't want to support multiusers and the password is authentication enough, no?

szaimen:
Close #25


szaimen:
@LukasReschke does this look good to you? :)

LukasReschke (approved):

PR #63, opened 2021-12-07
run image prune and volume prune regularly

### LINKED ISSUE #26: run docker image prune and docker volume prune regularly
probably add docker-cli into the mastercontainer and specify the socket. Then we should be able to put `sudo -u www-data docker image prune -f` and `sudo -u www-data docker volume prune -f` into the cronfile so that it runs once a day. https://github.com/nextcloud/all-in-one/blob/13360b176e58deca026e8a190ab221131ac96549/Containers/mastercontainer/cron.sh#L4-L7
https://stackoverflow.com/a/65588785
also add chmod +x

szaimen:
Close #26

### DIFF
--- a/Containers/mastercontainer/cron.sh
+++ b/Containers/mastercontainer/cron.sh
@@ -6,12 +6,5 @@ while true; do
     sudo -u www-data php /var/www/docker-aio/php/src/Cron/cron.php
     # Remove dangling images
     sudo -u www-data docker image prune -f
-    # Remove dangling volumes
-    unset DANGLING_VOLUMES
-    DANGLING_VOLUMES="$(sudo -u www-data docker volume ls --filter dangling=true | awk '{print $2}' | grep -v "nextcloud_aio_\|^VOLUME$")"
-    mapfile -t DANGLING_VOLUMES <<< "$DANGLING_VOLUMES"
-    for volume in "${DANGLING_VOLUMES[@]}"; do
-        sudo -u www-data docker volume rm "$volume"
-    done
     sleep 1d
 done
--- a/Containers/mastercontainer/cron.sh
+++ b/Containers/mastercontainer/cron.sh
@@ -2,8 +2,16 @@
 set -eux
 
 while true; do
+    # Check for updates and send notification if yes
     sudo -u www-data php /var/www/docker-aio/php/src/Cron/cron.php
+    # Remove dangling images
     sudo -u www-data docker image prune -f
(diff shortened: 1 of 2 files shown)


Containers/mastercontainer/cron.sh
```diff
@@ -1,7 +1,9 @@
 #!/bin/sh
-set -eu
+set -eux
 
 while true; do
     sudo -u www-data php /var/www/docker-aio/php/src/Cron/cron.php
+    sudo -u www-data docker image prune -f
+    sudo -u www-data docker volume prune -f
```
juliushaertl, line 7: Could there be scenarios where the cron is running while a container is recreated, so volumes might not be attached to any container during that time? I'd consider that a bit dangerous then to prune the volumes with potential data on it.

szaimen: Thats indeed possible:
https://github.com/nextcloud/all-in-one/blob/c203ad65ef32531ae700a9b2cfce6fcf7d7103ee/php/src/Controller/DockerController.php#L36-L39
Do you have any other idea how we could handle this? Without running the volume prune regularly, there will be a lot of unnecessary volumes after a while...

juliushaertl: Is there a separation between volumes that should be persistent and ones that would be only temporary per container? Maybe if all of the persistent ones are named, then they should not change during recreation and you could filter them out of a `docker volume ls` and remove the remaining ones individually.

szaimen: The thing is we don't know if the remaining ones without ours are stale. At least I don't know how to get this info without doing a docker inspect.
Perfect would be a dry-run command that we could use to get all the stale volumes and then just filter out ours and remove the rest. Unfortunately that doesnt exist yet:
https://github.com/moby/moby/issues/30623

szaimen: Okay, seems like we can simply run a docker image ls --filter dangling=true

szaimen: should be fixed now :)

juliushaertl (approved):

szaimen:
Thanks for the approval! After thinking about this a bit more I came to the conclusion that it is probably better to never run a volume prune automatically. This should be up to the admin. As most volumes are pretty small this shouldn't effect overall size that badly. Most important is that old images get cleaned up regularly which we do :)

PR #64, opened 2021-12-07
skip any update procedure if nextcloud was just restored or backup failed or database export failed

### LINKED ISSUE #23: After restore write file to nextcloud data volume so that it skips any update procedure
Skip from here to here:
https://github.com/nextcloud/all-in-one/blob/13360b176e58deca026e8a190ab221131ac96549/Containers/nextcloud/entrypoint.sh#L43-L206
and remove the file at the very end of the entrypoint script.

~~Also skip the database update in this case~~ not possible

### LINKED ISSUE #67: if backup creation fails, add file to nextcloud container so that it skips any automatic update
and remove it always if the backup was successful. (rm -f)

### LINKED ISSUE #69: make database export failproof
- add a skip-file before the export, remove it after it is successful
- if the file is present, skip any update of the database and just exit the database container -> report to github
- also cancel a backup creation if that file is present

szaimen:
Close #23
Close #67
Close #69

This is best reviewed like this: https://github.com/nextcloud/all-in-one/pull/64/files?diff=unified&w=1If 

The PR does basically two things:
1. if the database export during the database container shutdown fails, it will block new backups from getting created and will also block automatically restoring the database from database export (if an import should be needed)
2. if a backup fails or a backup was restored, it will not update the nextcloud the next time the nextcloud container starts up

### DIFF
--- a/Containers/borgbackup/backupscript.sh
+++ b/Containers/borgbackup/backupscript.sh
@@ -83,6 +83,13 @@ if [ "$BORG_MODE" = backup ]; then
         fi
     done
 
+    if [ -f "/nextcloud_aio_volumes/nextcloud_aio_database_dump/export.failed" ]; then
+        echo "Database export failed the last time. Most likely was the export time not high enough."
+        echo "Cannot create a backup now."
+        echo "Please report this to https://github.com/nextcloud/all-in-one/issues. Thanks!"
+        exit 1
+    fi
+
     # Create backup folder
     mkdir -p "$BORG_BACKUP_DIRECTORY"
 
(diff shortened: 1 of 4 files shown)


Containers/borgbackup/backupscript.sh
```diff
@@ -134,12 +134,18 @@ if [ "$BORG_MODE" = backup ]; then
     echo "Starting the backup..."
     get_start_time
     if ! borg create "${BORG_OPTS[@]}" "$BORG_BACKUP_DIRECTORY::$CURRENT_DATE-nextcloud-aio" "/nextcloud_aio_volumes/"; then
+        # Add file to Nextcloud container so that it skips any update the next time
+        touch "/nextcloud_aio_volumes/nextcloud_aio_nextcloud_data/skip.update"
+        chmod 777 "/nextcloud_aio_volumes/nextcloud_aio_nextcloud_data/skip.update"
```
szaimen, line 139: Todo: move that into start.sh as it could fail elsewhere, too. (Only create if the backupmode is create)

Also: create a fail file that lets any backup fail it the database export fails and remove that of the export is successful

szaimen: done

PR #70, opened 2021-12-08
allow to run this behind a reverse proxy

szaimen:


php/src/Data/ConfigurationManager.php
```diff
@@ -188,6 +196,15 @@ public function SetBorgBackupHostLocation(string $location) : void {
         $this->WriteConfig($config);
     }
 
+    public function GetApachePort() : string {
+        $port = getenv('APACHE_PORT');
```
szaimen, line 200: Here is basically where you set everything up. Basically when starting the mastercontainer and set the environmental variable APACHE_PORT, apache and the domaincheck container will use that.

szaimen: TODO: write it to the config file if the file is present. And overwrite the value with the one given by the environmental varialbe. If not given use the one form config.json

arvanus:
Nice! Any idea of when will it be merged? 
Thanks!

szaimen:
> Nice! Any idea of when will it be merged?

No

php/src/Data/ConfigurationManager.php
szaimen: done

szaimen:
lgtm now and will test this

PR #91, opened 2021-12-29
customize php-fpm config

### LINKED ISSUE #79: Adjust the php config to sane defaults
I would love to have the possibility to adjust the PHP settings, because I’m often hitting the 5 PHP-FPM processes limit.

`WARNING: [pool www] server reached pm.max_children setting (5), consider raising it`

I would suggest to add a config page to the AIO console, but doing it trough docker vars could work too.

szaimen:
Close #79

### DIFF
--- a/Containers/nextcloud/Dockerfile
+++ b/Containers/nextcloud/Dockerfile
@@ -217,6 +217,13 @@ RUN set -ex; \
     ; \
     rm -rf /var/lib/apt/lists/*
 
+RUN set -ex; \
+    grep -q '^pm = dynamic' /usr/local/etc/php-fpm.d/www.conf; \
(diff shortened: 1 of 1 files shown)


artonge:
I would have `COPY` the config files into their final destinations instead of inlining some sed commands.

artonge (approved):

juliushaertl (approved):
Seems good, though having fixed numbers is kind of hard without knowing the underlying available system ressources.

szaimen:
> I would have `COPY` the config files into their final destinations instead of inlining some sed commands.
> The values should be fine though.

Thanks for the feedback! Yes, usually I would copy the whole conf but since I only want to substitute those values, I opted for doing it like this. We can still change this without any problem later on.

> Seems good, though having fixed numbers is kind of hard without knowing the underlying available system ressources.

Thats right but lets do it like this for now and we can later on change it or maybe also improve it to set the value during runtime.

PR #93, opened 2021-12-30
create dependabot.yml

### LINKED ISSUE #7: Set up Dependabot

szaimen:
Close #7


.github/dependabot.yml
```diff
@@ -0,0 +1,27 @@
...
+      interval: weekly
+      day: saturday
+      time: "03:00"
+      timezone: Europe/Paris
+  open-pull-requests-limit: 10
+  labels:
+      - 3. to review
+      - dependencies
+  allow:
+      - dependency-type: direct
```
juliushaertl, line 27: What is the purpose of the second block? The first one should pick up the composer.json file in /php already

szaimen: okay nice, didn't know that. Thought we need that since the dependencies are currently stored in this repo, too. (look into the vendor directory)

juliushaertl: Right, that won't work to my knowledge. The vendor directory would always need to be updated manually or maybe through a github action. 

https://github.com/dependabot/dependabot-core/issues/386 was the feature request before dependabot was acquired by github. We have the same with the 3rdparty repo

szaimen: should be addressed with https://github.com/nextcloud/all-in-one/pull/93/commits/ce25a495927ee681fee92c6157418d6b98f702d1 :)

Containers/mastercontainer/Dockerfile
```diff
@@ -29,8 +29,16 @@ RUN curl "https://caddyserver.com/api/download?os=linux&arch=amd64" -o "/usr/bin
...
-RUN cd /var/www/docker-aio; \
-    git clone https://github.com/nextcloud/all-in-one.git .; \
+RUN set -e && \
+    curl -sS https://getcomposer.org/installer | php && \
+    mv composer.phar /usr/local/bin/composer && \
+    chmod +x /usr/local/bin/composer && \
+    cd /var/www/docker-aio; \
+    git clone https://github.com/nextcloud/all-in-one.git --depth 1 .; \
+    cd php; \
+    composer install --no-dev; \
```
juliushaertl, line 39:
```suggestion
    composer install --no-dev; \
    composer clearcache; \
```

Should avoid having any composer cache files in the image

juliushaertl (commented):
Small suggestion, but looks good otherwise :+1:

Containers/mastercontainer/Dockerfile
szaimen: Thanks, done :)

juliushaertl (approved):

szaimen:
Thanks for the review and approval! :)

PR #94, opened 2021-12-30
disable login if Nextcloud is running

### LINKED ISSUE #3: Hardening login process
The login process for the management container is currently set to six random words. There are several things that can be done to make the login more secure.

# Action Items

- [x] #4
- [ ] Don't allow login via the management container login if Apache is running (and not starting or stopped)
  - Login will happen via button inside Nextcloud (SSO)
- [ ] Add command line tool to unlock the management container login 
  - In case users the Nextcloud container isn't running properly

szaimen:
This is best reviewed like this: https://github.com/nextcloud/all-in-one/pull/94/files?diff=unified&w=1

Close #3

### DIFF
--- a/php/public/index.php
+++ b/php/public/index.php
@@ -89,9 +89,13 @@ $app->get('/containers', function ($request, $response, $args) use ($container) 
         'last_backup_time' => $configurationManager->GetLastBackupTime(),
     ]);
 })->setName('profile');
-$app->get('/login', function ($request, $response, $args) {
+$app->get('/login', function ($request, $response, $args) use ($container) {
     $view = Twig::fromRequest($request);
-    return $view->render($response, 'login.twig');
+    /** @var \AIO\Docker\DockerActionManager $dockerActionManger */
+    $dockerActionManger = $container->get(\AIO\Docker\DockerActionManager::class);
+    return $view->render($response, 'login.twig', [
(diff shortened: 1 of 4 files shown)


juliushaertl (changes requested):
Shouldn't this also be blocked then in the actual login controller? Otherwise it would just be the form that is hidden. 

https://github.com/nextcloud/all-in-one/blob/enh/3/disable-login-if-nextcloud-is-running/php/src/Controller/LoginController.php#L21

Would still be good to have a way to manually enable the login as mentioned in #3 to have a way to get back to it if something 
went wrong on the Nextcloud side.

szaimen:
Thanks for the review, julius :)

> Shouldn't this also be blocked then in the actual login controller? Otherwise it would just be the form that is hidden.

Not sure how important that is since it should fail in that case due to the csrf check?
@LukasReschke what is your take on this? :)

> Would still be good to have a way to manually enable the login as mentioned in #3 to have a way to get back to it if something
> went wrong on the Nextcloud side.

See https://github.com/nextcloud/all-in-one/pull/94/files#diff-e91b468a29aee3ff92547f02d1032086caca43138d1f9c5f693dea700f3e6058R17-R18 :)

juliushaertl (approved):

PR #112, opened 2022-01-09
add docs on how to edit the config.php file 

### LINKED ISSUE #101: Add a texteditor into Nextcloud container
How can the nextcloud config file be edited? or under which path is it located?

szaimen:
Close #101


juliushaertl:
Not really a big fan of having such tools inside of docker images, how about rather recommending to get the local volume path and rather adjust the config in there?

The following should work to get the volume path on the host:

```
docker volume inspect --format '{{ .Mountpoint }}' volumename
```

But feel free to decide of course, change looks good otherwise.

szaimen:
> how about rather recommending to get the local volume path and rather adjust the config in there?

Good idea, thanks for that!

szaimen:
@juliushaertl hopefully addressed the review and added some documentation on this :)

juliushaertl (approved):

PR #113, opened 2022-01-09
improve update instructions

### LINKED ISSUE #102: Improve update instructions
In the readme it says:

> If we push new containers to latest, you will see in the AIO interface below the containers section that new container updates were found. In this case, just press Stop containers and Start containers in order to update the containers. 

In the AIO interface it also says:

> Updates are available. Stop your containers and restart them to apply the update.

But below there is also a button:

> A new mastercontainer is available. Please click on the button below to update it.

So what is the correct procedure here: Do  we need to press the button or not - and when? 

On the basis of what I have learned so far - I am no sysadmin - by logic the procedure would be:

1. Stop the containers
2. Press update button
3. Start the containers

Am I right? 

In any case, the instructions should be revised, I think

szaimen:
Close #102


szaimen:
Hi @TheNomad11 is that clear enough? :)

szaimen:
thanks for the approval, julius :)

PR #126, opened 2022-01-10
add version number to the AIO interface

szaimen:
For possible reviewers: no logic at all is based on this version number currently so it is fine to have it hardcoded into the interface for now, imo. When I publish a new all-in-one container, I am going to create a new release then.

### DIFF
--- a/php/templates/containers.twig
+++ b/php/templates/containers.twig
@@ -16,7 +16,7 @@
     </header>
 
     <div class="content">
-        <h1>Nextcloud AIO Beta</h1>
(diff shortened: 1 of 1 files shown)


juliushaertl (approved):
In general i think the actual container versions or image hashes might be also helpful as the version here would only represent the state of the master container, but good from my side ;)

szaimen:
> In general i think the actual container versions or image hashes might be also helpful as the version here would only represent the state of the master container

Agreed but this is something for a follow-up

PR #136, opened 2022-01-14
improve setting the backup host location and allow `/var/backups`

### LINKED ISSUE #128: Borg backups : allow `/var/backups` as backup target
As stated in the doc. : Backups get created in the following directory on the host: /mnt/backup/borg 

Why do you force this dir ? Any reason ?

In my case, because I have standardised scripts on my servers where the backups dirs are not in the /mnt dir, I prefer to have the choice.

I needed to create a sym link to put the files elsewhere but it's a workaround.

szaimen:
Close #128

### DIFF
--- a/php/src/Data/ConfigurationManager.php
+++ b/php/src/Data/ConfigurationManager.php
@@ -172,14 +172,18 @@ class ConfigurationManager
 
         $isValidPath = false;
         foreach($allowedPrefixes as $allowedPrefix) {
-            if(str_starts_with($location, $allowedPrefix)) {
+            if(str_starts_with($location, $allowedPrefix) && !str_ends_with($location, '/')) {
+                $isValidPath = true;
+                break;
+            }
+            if ($location === '/var/backups') {
                 $isValidPath = true;
                 break;
             }
         }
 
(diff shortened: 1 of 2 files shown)


php/src/Data/ConfigurationManager.php
```diff
@@ -172,14 +172,18 @@ public function SetBorgBackupHostLocation(string $location) : void {
 
         $isValidPath = false;
         foreach($allowedPrefixes as $allowedPrefix) {
-            if(str_starts_with($location, $allowedPrefix)) {
+            if(str_starts_with($location, $allowedPrefix) && !str_ends_with($location, '/')) {
+                $isValidPath = true;
+                break;
+            }
+            if ($location === '/var/backups') {
```
juliushaertl, line 179: Subdirectories would also be fine from my point of view but your take I'd say :)

juliushaertl (approved):

php/src/Data/ConfigurationManager.php
szaimen: we'll only allow /var/backups for now :)

PR #145, opened 2022-01-18
fix conflicting branch names for workflows

szaimen:


szaimen:
merging because it is needed for the workflows to work correctly

PR #153, opened 2022-01-19
fix containers for arm64

szaimen:


szaimen:
Merging as it should work

PR #167, opened 2022-01-24
add docs for arm64

### LINKED ISSUE #28: Add support for ARM
- build multiarch images: https://www.docker.com/blog/multi-arch-build-and-images-the-simple-way/
- update the update check mechanism to compare with the correct image of the image list
- build the images
- make sure that all containers work on both platforms
- test if updating/update check works from current builds
- peobably more

szaimen:
Close #28

### DIFF
--- a/develop.md
+++ b/develop.md
@@ -14,40 +14,24 @@ nextcloud/all-in-one:develop
 And you are done :)
 It will now also select the developer channel for all other containers automatically.
 
+## How to build new containers
+
+Go to https://github.com/nextcloud-releases/all-in-one/actions/workflows/repo-sync.yml and run the workflow that will first sync the repo and then build new container that automatically get published to `develop` and `develop-arm64`.
+
 ## How to promote builds from develop to latest
-
-<!---
-You can use the Docker CLI to promote builds from develop to latest. Make sure to adjust:
-
-- $name
-- $digest
-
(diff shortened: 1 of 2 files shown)


develop.md
```diff
@@ -14,40 +14,24 @@ nextcloud/all-in-one:develop
...
 
-## How to promote builds from develop to latest
-
-<!---
-You can use the Docker CLI to promote builds from develop to latest. Make sure to adjust:
+## How to build new containers
 
-- $name
-- $digest
+Go to https://github.com/nextcloud-releases/all-in-one/actions/workflows/repo-sync.yml and run the workflow that will first sync the repo and then build new container that automatically get published to `develop` and `develop-arm64`.
```
juliushaertl, line 19: Just out of curiosity, is there a reason for the separate tags? It should be possible to build an image for multiple architectures using the same tag with https://docs.docker.com/buildx/working-with-buildx/

szaimen: Yes, there is one :)

See https://github.com/nextcloud/all-in-one/pull/42#issuecomment-986057315

juliushaertl: Ah I see, thanks for the quick reply :)

szaimen: no problem! Thanks for your reviews! :)

PR #218, opened 2022-02-12
fix the database connection

szaimen:
- [ ] needs #31144 so only merge if it is part of a nextcloud release (needed for new instances)


Containers/nextcloud/Dockerfile
```diff
@@ -105,7 +105,7 @@ RUN { \
 VOLUME /var/www/html
 
 
-ENV NEXTCLOUD_VERSION 23.0.1
+ENV NEXTCLOUD_VERSION 23.0.3
```
szaimen, line 108: I hope the mentioned fix in server will be reviewed, merged and backported for that release.

szaimen:
This needs a different approach due to
image

szaimen:
We should be able to create a user with less permissions and use that for the nextcloud connection. Important is that it doesn't has create role permissioins; database creation should be enough.

szaimen:
This is ready for review :)

juliushaertl (approved):
Looks good code wise 🐘

szaimen:
Thanks! will test this soon :)

Issue #205, opened 2022-02-09
migrate containers to alpine for decreased size and increased security


szaimen:
- [x] https://github.com/nextcloud/all-in-one/pull/1573  
- [x] https://github.com/nextcloud/all-in-one/pull/1578
- [x] https://github.com/nextcloud/all-in-one/pull/1577 
- [ ] https://github.com/nextcloud/all-in-one/pull/1583

szaimen:
Handling talk in https://github.com/nextcloud/all-in-one/issues/1992

PR #306, opened 2022-03-02
migrate borgbackup to alpine

szaimen:
Close part of https://github.com/nextcloud/all-in-one/issues/205


szaimen:
@juliushaertl @max-nextcloud do you know if we might run into an issue with already existing instances here as we keep the root dir between container exchanges? https://github.com/nextcloud/all-in-one/blob/853a880c9df4b994c95190c83f82950fccc66772/php/containers.json#L206-L210

juliushaertl (approved):
🐘

szaimen:
Thanks!

PR #339, opened 2022-03-09
improve error logging in dockeractionmanager

szaimen:

### DIFF
--- a/php/public/index.php
+++ b/php/public/index.php
@@ -141,4 +141,6 @@ $app->get('/', function (\Psr\Http\Message\RequestInterface $request, \Psr\Http\
     }
 });
 
+$errorMiddleware = $app->addErrorMiddleware(true, true, true);
+
 $app->run();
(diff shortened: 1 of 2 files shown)


juliushaertl:
Might be nice to actually get a error handling middleware in place, so that uncaught exceptions will get logged generally.

https://www.slimframework.com/docs/v4/middleware/error-handling.html

For that logging in a similar way we do it in the server with the `getTraceAsString()` might help for further investigations as well.

szaimen:
> Might be nice to actually get a error handling middleware in place, so that uncaught exceptions will get logged generally.
> 
> https://www.slimframework.com/docs/v4/middleware/error-handling.html
> 
> For that logging in a similar way we do it in the server with the `getTraceAsString()` might help for further investigations as well.

Thanks for the idea! I added this as an issue :)

szaimen:
for later: we'll need the crash if domaincheck fails to start...

PR #347, opened 2022-03-11
move inline javascript to its own file

szaimen:

### DIFF
--- a/php/public/automatic_reload.js
+++ b/php/public/automatic_reload.js
@@ -1,0 +1,13 @@
+if (document.hasFocus()) {
+    // hide reload button if the site reloads automatically
+    var list = document.getElementsByClassName("reload button");
+    for (var i = 0; i < list.length; i++) {
+        // list[i] is a node with the desired class name
+        list[i].style.display = 'none';
+    }
+
+    // set timeout for reload
+    setTimeout(function(){
+    window.location.reload(1);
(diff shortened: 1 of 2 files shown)


szaimen:
@max-nextcloud would be cool if you could give me a review on this :)

php/public/automatic_reload.js
```diff
@@ -0,0 +1,13 @@
+if (document.hasFocus()) {
```
max-nextcloud, line 1: My understanding is that this check will be triggered once - when the js is loaded and evaluated for the first time.

If the document does not have the focus then the rest will never be executed.
I think `setTimeout` should always be called and the if condition should live inside the function of `setTimeout`. That way we'd periodically check if the document has focus now and reload if that is the case.

max-nextcloud (commented):
The move seems fine. Commenting on the code never the less as I don't think it does what it is intended for.

php/public/automatic_reload.js
szaimen: if the document does not have focus it will display a reload button that is inserted by default. always setting the timeout did not work for me unfortunatley if I remember correctly.

szaimen:
> The move seems fine.

Thanks for the review and this comment :)
My question was basically if the move will work as intended but I'll obivously test this before releasing it 👍

PR #348, opened 2022-03-11
refactor some config logic

szaimen:

### DIFF
--- a/php/src/Data/ConfigurationManager.php
+++ b/php/src/Data/ConfigurationManager.php
@@ -265,26 +265,10 @@ class ConfigurationManager
     }
 
     public function GetApachePort() : string {
-        $port = getenv('APACHE_PORT');
-        if ($port === false) {
-            $config = $this->GetConfig();
-            if (!isset($config['apache_port']) || $config['apache_port'] === '') {
-                $config['apache_port'] = '443';
-            }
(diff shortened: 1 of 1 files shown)


php/src/Data/ConfigurationManager.php
```diff
@@ -297,6 +281,29 @@ public function WriteConfig(array $config) : void {
...
 
+    private function GetEnvironmentalVariableOrConfig(string $envVariableName, string $configName, string $defaultValue) : string {
+        $envVariableOutput = getenv($envVariableName);
+        if ($envVariableOutput === false) {
+            $config = $this->GetConfig();
+            if (!isset($config[$configName]) || $config[$configName] === '') {
+                $config[$configName] = $defaultValue;
+            }
+            return $config[$configName];
+        } else {
```
CarlSchwan, line 292: It's a code style preference but since you have a return in the if block, you don't need a block for the else statement.

CarlSchwan (approved):
Looks fine. Just a minor code style issue but feel free to ignore it 😉

php/src/Data/ConfigurationManager.php
szaimen: right, will do like suggested. thanks for the review :)

PR #395, opened 2022-03-17
add docs on how to run it on windows and improve macOS docs

szaimen:

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -79,7 +79,26 @@ Nextcloud AIO is inspired by projects like Portainer that allow to manage the do
 Yes. Please refer to the following documentation on this: [reverse-proxy.md](https://github.com/nextcloud/all-in-one/blob/main/reverse-proxy.md)
 
 ### How to run it on macOS?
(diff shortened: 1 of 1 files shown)


szaimen:
I figured it out and tested it successfully. With that docs it runs nearly anywhere now I'd say :)

szaimen:
Thouh this probably needs to allow `/host_mnt/*` as backup path for windows e.g. `/host_mnt/c` for the c drive though I couldn't make it work with that. Seems like it behaves differently on windows:
Even manually setting produces this config:
```
 "Mounts": [
            {
                "Type": "bind",
                "Source": "/host_mnt/c/TEMP/website1",
                "Destination": "/mnt/borgbackup",
                "Mode": "",
                "RW": true,
                "Propagation": "rprivate"
            },
]
```
but it should instead generate this:
```
"Mounts": [
            {
                "Type": "bind",
                "Source": "/host_mnt/c/TEMP/website1",
                "Destination": "/mnt/borgbackup",
                "Mode": "rw",
                "RW": true,
                "Propagation": "rprivate"
            }
]
```

szaimen:
tracked in https://github.com/nextcloud/all-in-one/issues/396

Issue #400, opened 2022-03-18
Increase Nextcloud HPB limit?


muellerm01:
### How to use GitHub

* Please use the 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to show that you are affected by the same issue.
* Please don't comment if you have no relevant information to add. It's just extra noise for everyone subscribed to this issue.
* Subscribe to receive notifications on status change and new comments. 

### Steps to reproduce
Software
Install Nextcloud AIO 0.8 like documented

Hardware:
Azure North Europe Standard D4as v4 (2 vcpus, 8 GiB Arbeitsspeicher)
8 Gbit download (measured)
1,8 Gbit upload (measured)

Steps:
1. start a meeting with allowed guestes
2. the 7th person will take a little bit longer to join
3. the 8th person is not able to join the meeting

### Expected behaviour 
Nextcloud talk participant limit should be releated to CPU, RAM and upload/download speed

### Actual behaviour 
CPU runs on 0,5 cores
RAM 2,5 GB used
Upload/Download ~20Mbit
and participant limit is between 7 and 8

---

#### Nextcloud AIO version 
AIO 0.8
#### Current channel 
latest channel

szaimen:
Hi, thanks for your report. I fear though that there is not much we can do here because we install everything based on the docs.

Feel free to read the following docs though: https://github.com/nextcloud/spreed#how-to-have-the-maximum-number-of-participants-in-a-call

szaimen:
Also it seems like your cpu is pretty limitied. Feel free to try to increase the cpu count

muellerm01:
OK, sorry I thought HPB was part of AIO because it is configured.
I requested your sales team but think the minimum user count is around 100 to get enterpise support and search for a solution below that. Let´s see what i will get from them.

I also tried it with a larger CPU, but that doesn't seem to change anything.

I think in the range of 50 participants and an on premise installation. But I could not find a purchasable solution so far. 

Hosting is not an option for me. 
Then I could just take teams.

szaimen:
> OK, sorry I thought HPB was part of AIO because it is configured.

yes, the HPB is part of AIO.

szaimen:
It seems like it is just not scaling as well as promised...

szaimen:
> But than HPB with a limit of 7-8 thems to be confusing.

yes, it is

muellerm01:
Is something configured incorrectly?

[...]

PR #402, opened 2022-03-18
adjust talk hpb config

szaimen:
Close https://github.com/nextcloud/all-in-one/issues/400


szaimen:
@muellerm01 does this look good to you?

szaimen:
@morph027 it looks like the hpb config might not apply anymore with new releases of the signaling server. See 
https://github.com/strukturag/nextcloud-spreed-signaling/blob/master/backend_configuration.go
vs
https://github.com/nextcloud/all-in-one/blob/main/Containers/talk/start.sh

Do you maybe have some ideas/pointers what we need to change?

szaimen:
@morph027 here is a logfile: https://github.com/nextcloud/all-in-one/issues/400#issuecomment-1072434711

szaimen:
Thanks for having a look! :)

PR #494, opened 2022-04-15
Adding apache port environment variable to docker-compose.yml

Arkonos:
We compared the startup command from https://github.com/nextcloud/all-in-one/blob/main/reverse-proxy.md to this file and noticed that "-e APACHE_PORT=11000" is not reflected here.
After adding the line, we could progress with our setup.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -16,3 +16,5 @@ services:
       - 80:80
       - 8080:8080
       - 8443:8443
+    environment:
+      - APACHE_PORT=11000


szaimen:
Hi, thanks for your PR but this will config will not work for everyone because not everyone runs a reverse proxy. I see two options: either add a second docker-compose.yml to the root of this repository for using it behind a reverse proxy or adding it as an example to the documentation: https://github.com/nextcloud/all-in-one/blob/main/reverse-proxy.md

What do you think? :)

acetousk:
I would say that just having a docker compose file that is named something like `docker-compose-reverse-proxy.yml` with additional documentation on the `readme.md` would be great.

Also having a reverse proxy example with a docker compose file for nginx / caddy would be great also

Arkonos:
I wanted to reserve my judgement until we progressed with our setup, but that might take a while.
File or documentation doesn't matter to us specifically. We rtfm and as long as it's easy enough to follow we'll manage.
A separate config file that can just be pulled directly sounds easier to use, but if it is more bothersome to maintain I'd understand documentation only.
Thank you

szaimen:
Lets add all possibilities to the default docker-compose file and simply add comments for what other values are supported. I'll make a POC

szaimen:
Made a POC in https://github.com/nextcloud/all-in-one/pull/563

PR #499, opened 2022-04-17
docs: provide context around reverse proxy config

ForestJohnson:
I just spent a couple hours learning about `nextcloud-aio` in order to help someone debug their nextcloud-aio deployment. they were trying to use `nextcloud-aio` with a caddy reverse proxy docker container on a custom docker network.  It wasn't working because nextcloud-aio spawns an auxiliary `domaincheck` container on the default docker network and maps the APACHE_PORT to the host.  

But the user's reverse proxy docker container cannot dial the  `domaincheck` container because its on a different docker network, and the  reverse proxy docker container can't dial the host (or, at least, the user never figured out how) 

The docs need to provide context around the instructions, if the user is allowed to make assumptions about what contexts the instructions are valid in, they will try to do all kinds of stuff which wont work, resulting in failure, frustrated users.


ForestJohnson:
missing sign off

PR #593, opened 2022-05-06
improve the reverse-proxy documentation another time

szaimen:
Close https://github.com/nextcloud/all-in-one/discussions/568


szaimen:
@peracchi I finally see where the issue comes from. Mostly my wrong idea that all servers have always also a private ip-address.

Do you think this PR is ready good to go? :)

szaimen:
Ah I see. lets discuss this in the thread further...

Issue #660, opened 2022-05-19
Digitalocean AIO install cannot receive port 80 HTTP traffic


Eeemil:
### How to use GitHub

* Please use the 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to show that you are affected by the same issue.
* Please don't comment if you have no relevant information to add. It's just extra noise for everyone subscribed to this issue.
* Subscribe to receive notifications on status change and new comments. 

### Steps to reproduce
1. install nextcloud from https://marketplace.digitalocean.com/apps/nextcloud
2. follow installation steps, wait for containers to start
3. curl `server.name.tld`, get "connection refused on port 80"

### Expected behavior 
A HTTPS redirect
### Actual behavior 
Connection refused

### Host OS 
Ubuntu 20.04 via https://marketplace.digitalocean.com/apps/nextcloud

#### Nextcloud AIO version 
v1.2.0
#### Current channel 
latest
#### Other valuable info 

I "resolved" the issue by `docker exec -it nextcloud-aio-mastercontainer` and added the following:
```
Listen 80

<VirtualHost *:80>
    RewriteEngine On
    RewriteCond %{HTTPS} off
    RewriteRule ^ https://%{HTTP_HOST}%{REQUEST_URI}
</VirtualHost>
```
to the top of /etc/apache2/sites-enabled/mastercontainer.conf and this resolved the issue. However, I'm afraid this will either be overwritten during updates or that the letsencrypt certificate will be unable to update.

PR #670, opened 2022-05-19
Fix https redirects

Eeemil:
Trying a request to a new AIO install with an ordinary HTTP request on port 80 would just give "connection refused". Some browsers (Firefox) automatically tries HTTPS if HTTP fails, but not all.

This PR adds a `http://:80` block to caddy to automatically send a https redirect. Removing the `auto_https disable_redirects`-directive did not properly resolve the issue, this created redirects to `https://{host}:8443{uri}` (note port 8443) directs the traffic to the AIO Installer, rather than the Nextcloud instance (which is probably the most common use case). The redirect to `:8443` is probably due to the `8443`-block being the only known service to Caddy

See https://github.com/nextcloud/all-in-one/issues/660 and https://github.com/nextcloud/all-in-one/discussions/667 for more context


szaimen:
Great, thanks for the PR! :)
@Eeemil can you please fix DCO? See https://github.com/nextcloud/all-in-one/pull/670/checks?check_run_id=6509421470 how to fix it. If that is too complicated for you, feel free to create a new PR with a commit that contains the sign-off. Thank you very much for your effort!

Eeemil:
Done!

szaimen:
Thank you! :)

szaimen:
Hey @Eeemil , before merging this, I though about adding something like this:
```
http://:8443 {
  redir https://{host}{uri}:8443
}
```
Do you think this would allow to redirect to the https connection if you forget to enter the correct protocol for port 8443? :)

Eeemil:
Yes I think that would probably work, e.g. http://sitename:8443 would be redirected to https://sitename:8443

I'm unable to test it at this moment though, maybe later this week if you are not able to test it yourself. Should I add it or you?

Containers/mastercontainer/Caddyfile
```diff
@@ -1,11 +1,17 @@
...
 
     storage file_system {
         root /mnt/docker-aio-config/caddy/
     }
 }
 
+http://:80 {
+  redir https://{host}{uri}
+}
+
```
Eeemil, line 14:
```suggestion
http://:8443 {
  redir https://{host}:8443{uri}
}
```

Something like this?

szaimen:
> Yes I think that would probably work, e.g. http://sitename:8443 would be redirected to https://sitename:8443

Yes, that is the idea :)

> I'm unable to test it at this moment though, maybe later this week if you are not able to test it yourself. Should I add it or you?

I'll add it :)

Containers/mastercontainer/Caddyfile
szaimen: done

szaimen:
> maybe later this week if you are not able to test it yourself

yes, would be cool if you could test it, thank you for your effort! :)

szaimen (approved):

Issue #673, opened 2022-05-19
Wrong redirect from Nextcloud settings to aio if the aio is behind reverse proxy


2020Sanoj:
### How to use GitHub

* Please use the 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to show that you are affected by the same issue.
* Please don't comment if you have no relevant information to add. It's just extra noise for everyone subscribed to this issue.
* Subscribe to receive notifications on status change and new comments. 

### Steps to reproduce
1. set up Nextcloud AIO & set up Reverse Proxy over Nginx Proxy Manager (works)
2. set up a reverse proxy with https for the aio interface (to port 8080)
3. open the aio interface over the reverse proxy
4. start the conatiner
5. open nextcloud
6. open from the nextcloud settings the aio interface
7. you will get send to https://aio.domain:8000, so it places the correct domain and set the protocol to https, but it adds a port, which is not need and since the port is only open to localhost, it will time out, if I manually copy the URL and remove the port it works

### Expected behavior 
- It redirects to https://aio.domain:port (or no port if the port would be 443)

### Actual behavior 
- It redirects to https://aio.domain:8000

### Host OS 
- Debian 11

#### Nextcloud AIO version 
v1.2.1

#### Current channel 
beta

#### Other valuable info 
- removing :8000 in /var/lib/docker/volumes/nextcloud_aio_mastercontainer/_data/data/configuration.json, does not help because after a restart :8000 is added again

szaimen:
Hello, does this help?
![Screenshot_20220519-184043_Brave](https://user-images.githubusercontent.com/42591237/169353437-993bb326-a3e8-4b5f-ab53-039ef1f9b6a8.png)

2020Sanoj:
No, I tested changing the domain, this worked, so in the mastercontainer configuration.json was the new domain and also on the Nextcloud overview, after restarting all containers over the aio interface from this domain on port 443. But the port :8000 was still in the config and in the url of the button in the Nextcloud overview, so both got the new domain, but kept the port :8000, but I've never opened the aio interface over port 8000

szaimen:
Do I understand correctly that you are running the AIO interface behind a reverse proxy?

2020Sanoj:
Yes, also Behind the Nginx Proxy Manager

szaimen:
All right. Let me investigate

szaimen:
[...]

PR #677, opened 2022-05-20
set AIO_URL port to 443 if port 8000 was discovered

szaimen:
Fix https://github.com/nextcloud/all-in-one/issues/673


szaimen:
@2020Sanoj does this look good to you? :)

2020Sanoj:
Looks good, but there should maybe also a note in the documentation  that the aio panel should not be  on port 8000, or?

szaimen:
> Looks good, but there should maybe also a note in the documentation that the aio panel should not be on port 8000, or?

I think it is not that important for now. If really needed we can add it later.

Thanks for the review! :)

Issue #665, opened 2022-05-19
Add docs on how to reset instance


szaimen:
(Maybe also document how to backup/restore in that case?)

PR #678, opened 2022-05-20
document how to reset the instance properly

szaimen:
Close https://github.com/nextcloud/all-in-one/issues/665

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -120,6 +120,23 @@ Additionally, there is a cronjob that runs once a day that checks for container 
 ### How to easily log in to the AIO interface?
 If your Nextcloud is running and you are logged in as admin in your Nextcloud, you can easily log in to the AIO interface by opening `https://yourdomain.tld/settings/admin/overview` which will show a button on top that enables you to log in to the AIO interface by just clicking on this button. **Note:** You can change the domain/ip-address/port of the button by simply stopping the containers, visiting the AIO interface from the correct and desired domain/ip-address/port and clicking once on `Start containers`.
 
+### How to properly reset the instance?
+If something goes unexpected routes during the initial installation, you might want to reset the AIO installation to be able to start from scratch.
+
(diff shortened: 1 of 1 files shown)


szaimen:
@Joly0 does this look good to you? :)

Joly0:
In my opinion the commands should be adjusted so not all containers and images not running get deleted.

Here is an example command, that might work (not tested)

Docker container prune --filter name=nextcloud_aio_*

And the volume equivalent should be used, though i havent tested this command.

Otherwise these commands should work

docker rm -f $(docker ps -a -q --filter name=nextcloud_aio_*)

docker volume rm $(docker volume ls -q --filter name=nextcloud_aio_*)


I could test these later today, if needed.

szaimen:
Thanks, good suggestion! I'll adjust the documentation accordingly!

szaimen:
@Joly0 does this look good to you now? :)

Joly0:
As far as i know, there is the possibility to filter for names though i would have to test that.
So if you could wait a bit longer, than i would test that and try to figure out a way how this could work.

szaimen:
> As far as i know, there is the possibility to filter for names though i would have to test that. So if you could wait a bit longer, than i would test that and try to figure out a way how this could work.

All right, I'll wait then until you report back if you found a solution 👍 thanks for investigating! :)

Joly0:
I have found the correct command to remove the containers and their volumes (though the images are not getting deleted)

docker rm -v $(docker ps -a -q --filter name=nextcloud)

Would be great, if someone could test and verify that command. If i find a command for deleting the images aswell, then i will add it here

szaimen:
@Joly0 thanks for investigating but I honestly would not use the command substitution here because it will not work on all supported platforms so I adjusted the commands to show which things will be removed before executing the specific commands. Is that fine for you, too?

Joly0:
Alright, i didnt thought about every usecase and you are right, some platforms dont or might not support this way. So if it is well documented, what the commands do and what to take care of, i am fine with it

szaimen:
All right! Thanks for your review! Merging then :)

PR #682, opened 2022-05-20
add nginx-proxy-manager to the reverse proxy documentation

szaimen:
Close #681

### DIFF
--- a/reverse-proxy.md
+++ b/reverse-proxy.md
@@ -29,7 +29,7 @@ https://<your-nc-domain>:443 {
 }
 ```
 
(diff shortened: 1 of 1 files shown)


szaimen:
@2020Sanoj does this look good to you? :)

2020Sanoj:
Maybe these pictures are better:

grafik
grafik
grafik
(to this picture, they can also use custom uploaded / existing certificates in the selection or use wildcard certificates they generated over a DNS challenge)
grafik


And what about also mention that the nginx proxy manager can also be started in host mode with adding, this to the compose file, if the people use docker compose for the npm:
`    network_mode: host`

And I think you won't will add this, but this would be the details page I use for the aio dashboard, the other config pages in the npm are the same. (it is working, only there is currently #673 which is not working). Maybe this will be useful someday.
grafik

2020Sanoj:
Yes, but if you mention that `<your-nc-domain>` needs to be changed, why not also mention, that `<you>@<your-mail-provider-domain>` needs to be changed?
But beside this it looks good and should be understandable for everyone using the npm.

2020Sanoj:
Looks good now for me

szaimen:
> Looks good now for me

thanks! I'll merge as soon as CI is done :)

Issue #753, opened 2022-05-27
Update notification spam


2020Sanoj:
### How to use GitHub

* Please use the 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to show that you are affected by the same issue.
* Please don't comment if you have no relevant information to add. It's just extra noise for everyone subscribed to this issue.
* Subscribe to receive notifications on status change and new comments. 

### Steps to reproduce
1. wait for an aio update
2. wait for update notifications

### Expected behavior 
one notification
### Actual behavior 
many notifications

### Host OS 
debian 11

#### Nextcloud AIO version 
v1.3.0
#### Current channel 
beta
#### Other valuable info 
![grafik](https://user-images.githubusercontent.com/75573284/170776393-ddd53138-8e97-4b37-9e25-c835f9d9d7ba.png)

szaimen:
Thanks for the report! Can you please post the output of `sudo docker exec -it nextcloud-aio-nextcloud php /var/www/html/occ user:list`?

2020Sanoj:
```
  - David: David *** ***
  - Jonlp00: *** ***
  - Sanoj: Jonas L***
  - Theo: Theo ***
```

2020Sanoj:
Just a note, all 14 notifications are about the aio containers, so two notifications per aio container

szaimen:
Hi, can you maybe post the mastercontainer logs of when this problem happens? This will help understanding why this happens. Thank you! :)

2020Sanoj:
Yes, when the next update gets released

szaimen:
I think I found the problem. Will be fixed in 759.

PR #759, opened 2022-05-29
fix the notification spam

szaimen:
Fix https://github.com/nextcloud/all-in-one/issues/753


szaimen:
https://github.com/nextcloud/all-in-one/pull/759/files#diff-8134a42e79b37f710bb0edb05165e2c2ac6197e1d6bfc80db88c8d39b03eb97dR22
This is the problem: we only check for minutes here... Maybe we should check every 30sec and if found wait 1min...

Issue #813, opened 2022-06-11
Slim application error when not using Docker tag 'latest' in imagename


dantefromhell:
Ansible applies the 'latest' tag per default if not supplied in the image name [1].
If the tag was not supplied in the image name, it is not shown in the `Config.Image` output of `docker inspect $CONTAINER_NAME`.

After some digging into `docker inspect` it seems a more reliable way to retrieve the image tag could be:
1) Grab the `Image` hash from `docker inspect nextcloud-aio-mastercontainer`
2) Read the `RepoTags` variable from `docker image inspect $IMAGE_HASH`

The benefit of this solution would be that the tag is directly read from the image that is in use [2] 

[1] https://docs.ansible.com/ansible/latest/collections/community/docker/docker_container_module.html#parameter-image
[2] https://containers.gitbook.io/build-containers-the-hard-way/#manifest

### Steps to reproduce
1. Deploy nextcloud-aio using the following ansible snippet
```
  tasks:
    - name: Start 'nextcloud-all-in-one' container
      community.docker.docker_container:
        name:  nextcloud-aio-mastercontainer
        image: nextcloud/all-in-one
        state: started
        restart_policy: unless-stopped
        volumes:
          - nextcloud_aio_mastercontainer:/mnt/docker-aio-config
          - /var/run/docker.sock:/var/run/docker.sock:ro
        ports:
          - "{{ ansible_default_ipv4.address }}:80:80"
          - "{{ ansible_default_ipv4.address }}:8443:8443"
          - "{{ ansible_default_ipv4.address }}:443:443"
```
3. Browse to `https://server-name:8443` and grab password
4. Log in, get error message

### Expected behavior 
The setup configuration dialog is shown

### Actual behavior 
After login, setup fails with the following stracktrace
```
Slim Application Error
Type: TypeError
Code: 0
Message: AIO\Docker\DockerActionManager::GetCurrentChannel(): Return value must be of type string, null returned
/var/www/docker-aio/php/src/Docker/DockerActionManager.php
Line: 434
Trace: #0 /var/www/docker-aio/php/src/Docker/DockerActionManager.php(51): AIO\Docker\DockerActionManager->GetCurrentChannel()
#1 /var/www/docker-aio/php/src/Docker/DockerActionManager.php(343): AIO\Docker\DockerActionManager->BuildImageName(Object(AIO\Container\Container))
[...]

PR #814, opened 2022-06-11
allow to use it even if no channel was provided

szaimen:
Close https://github.com/nextcloud/all-in-one/issues/813


LukasReschke (commented):
`isset` would probably help avoiding the Lint warning :)

szaimen:
> `isset` would probably help avoiding the Lint warning :)

thanks for the idea but the following does not seem to work...

```
            // /**
            //  * @psalm-suppress TypeDoesNotContainNull
            //  */
            if (!isset($tag)) {
                error_log("No tag was found when getting the current channel. You probably did not follow the documentation correctly. Changing the channel to the default 'latest'.");
                $tag = 'latest';
            }
```

Issue #823, opened 2022-06-13
Docker rootless - Permission issue?


Kirk1984:
Hello,

today i tried setting up AiO (1.4.0) under Ubuntu 22.04 and ran into a permission issue. i setup a clean Debian 11 install with the same result. It is the same result with and without the "NEXTCLOUD_DATADIR" env variable.
The used compose file worked before during testing. I guess it started with 1.3.1, but sadly i didn't pay much attention because the old test microsd was corrupt.

This is the error the aio-nextcloud container puts out:
```
Configuring Redis as session handler
touch: /mnt/ncdata/this-is-a-test-file: Permission denied
The www-data user doesn't seem to have access rights in /mnt/ncdata.
Did you maybe change the datadir and did forget to apply the correct permissions?
```

And the Apache container waits for nextcloud:
```
nc: getaddrinfo for host "nextcloud-aio-nextcloud" port 9000: Name or service not known
Waiting for Nextcloud to start...
```

szaimen:
Hi, can you show me the docker compose file?

Also did you follow https://github.com/nextcloud/all-in-one#how-to-change-the-default-location-of-nextclouds-datadir and adjusted the permissions correctly?

Kirk1984:
My current docker-compose:
```
version: "3.8"

volumes:
  nextcloud_aio_mastercontainer:
    name: nextcloud_aio_mastercontainer

services:
  nextcloud:
    image: nextcloud/all-in-one:latest-arm64
    restart: unless-stopped
    container_name: nextcloud-aio-mastercontainer
    volumes:
      - nextcloud_aio_mastercontainer:/mnt/docker-aio-config
      - /$XDG_RUNTIME_DIR/docker.sock:/var/run/docker.sock:ro
    environment:
      - APACHE_PORT=11000
      - DOCKER_SOCKET_PATH=$XDG_RUNTIME_DIR/docker.sock
      - NEXTCLOUD_DATADIR=/mnt/ncdata
    ports:
      - 8080:8080
```

I also did the permissions you mentioned before starting the container. neither /mnt/ncdata nor any other location (a folder in /home for example) worked so far.

And i was wrong, it works without NEXTCLOUD_DATADIR, it just takes its sweet time on my Pi.

szaimen:
> And i was wrong, it works without NEXTCLOUD_DATADIR, it just takes its sweet time on my Pi.

So then it works now and we can close this?

szaimen:
[...]

PR #829, opened 2022-06-15
add a note on docker-rootless regarding permissions outside the conta…

szaimen:
…iners

Close https://github.com/nextcloud/all-in-one/issues/823


szaimen:
@Kirk1984 does this look good to you?

Kirk1984:
I don't think that will work.
In my test the user pi is 1000:1000 and the Folder the rootless docker should use is pi:pi. It didn't work. It only worked when i setup Nextcloud All in One with a volume and copied the value that volume had (1000032:1000032) onto the Folder i want to Bind. Only then it worked. But "systemctl --user status docker" shows only group 1000 for cgroup and containerd.

"ps -u pi" shows that the user pi runs docker (dockerd, containerd, etc). The same user that created the folder that should be used as a bind volume.

szaimen:
Okay, so the folders outside the containers do not created with the user that is running the docker daemon?

Kirk1984:
> Okay, so the folders outside the containers do not created with the user that is running the docker daemon?

No, the volume was 100032:100032. Which seems to be normal (what i found googling). Looks like the values are always over 100000. 

 grep ^$(whoami): /etc/subuid shows also 100000 as the first value.

szaimen:
Hm... then this is maybe really a bug?

Or could `100032:100032` be applied to make it work?

Kirk1984:
When i'm i will check if maybe values over100000 work in general. My older system doesn't use 32 but 999.

PR #830, opened 2022-06-15
fix trusted_proxies for caddy in case of a reverse proxy situation

### LINKED ISSUE #828: trusted proxies does not work as intended
### How to use GitHub

* Please use the 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to show that you are affected by the same issue.
* Please don't comment if you have no relevant information to add. It's just extra noise for everyone subscribed to this issue.
* Subscribe to receive notifications on status change and new comments. 

### Steps to reproduce
1. Install nextcloud aio with custom apache port
2. Nextcloud brute force thinks every access comes from ::1 and the log thinks every wrong login comes from ::1 / `docker logs nextcloud-aio-nextcloud` shows 172.25.0.9
3. Set ::1 as trusted proxy in the config.php
4. With apache and caddy fd00:dead:beef:c0::1 is now shown / with the Nginx Proxy Manager 172.17.0.1, both are the gateways of the docker bridge network (X-Forwarded-For header is set in all the configurations)
5. Setting fd00:dead:beef:c0::1/172.17.0.1/172.25.0.9 as trusted proxy changes nothing.

### Expected behavior 
The real IP is shown (which is sent in the X-Forwarded-For header)
### Actual behavior 
::1 is shown as IP to access / `docker logs nextcloud-aio-nextcloud` shows 172.25.0.9
With ::1 set as trusted proxy, it shows fd00:dead:beef:c0::1  /  `docker logs nextcloud-aio-nextcloud` shows 172.25.0.9
Setting fd00:dead:beef:c0::1/172.17.0.1/172.25.0.9 as trusted proxy changes nothing
This makes the brute force protection useless and makes it block everything, instead of a certain IP
If my cloud gets blocked, I need to run `[sudo] docker exec -it nextcloud-aio-nextcloud php occ security:bruteforce:reset ::1` to make it possible for me to login, (Whitelisting ::1 in brute force settings app also works)

### Host OS 
Debian GNU/Linux 11 (bullseye)

#### Nextcloud AIO version 
v1.4.0  and v1.4.1
#### Current channel 
Beta
#### Other valuable info 
Default Nextcloud config.php (Apache/Caddy and NPM):
grafik
NPM with ::1 as set trusted proxy:
grafik
Apache2/Caddy with ::1 set as trusted proxy:
grafik

fd00:dead:beef:c0::1 and 172.17.0.1 are the gateways of the docker bridge network and 172.25.0.9 the IP of the nextcloud-aio-apache container
The X-Forwarded-For header is set by all three proxies to my real IP (`185.***.**.**`)

Edit:
Also, this block of trusted proxies does not work (And yes, some entries in this block are nonsense)
```
  array (
    0 => '127.0.0.1',
    1 => '172.25.0.9:*',
    2 => '172.17.0.1:*',
    3 => '0.0.0.0/0:*',
    4 => '0.0.0.0/0',
    5 => '0.0.0.0',
    6 => '::1/0:*',
    7 => '::1/0',
    8 => '::1',
  ),
```

szaimen:
Close #828

### DIFF
--- a/Containers/apache/Caddyfile
+++ b/Containers/apache/Caddyfile
@@ -11,24 +11,34 @@
     # Notify Push
     route /push/* {
         uri strip_prefix /push
-        reverse_proxy {$NEXTCLOUD_HOST}:7867
+        reverse_proxy {$NEXTCLOUD_HOST}:7867 {
+            # trusted_proxies placeholder
+        }
     }
 
     # Talk
     route /standalone-signaling/* {
         uri strip_prefix /standalone-signaling
-        reverse_proxy {$TALK_HOST}:8081
+        reverse_proxy {$TALK_HOST}:8081 {
+            # trusted_proxies placeholder
+        }
     }
 
     # Collabora
     route /browser/* {
-        reverse_proxy {$COLLABORA_HOST}:9980
+        reverse_proxy {$COLLABORA_HOST}:9980 {
+            # trusted_proxies placeholder
+        }
     }
     route /hosting/* {
-        reverse_proxy {$COLLABORA_HOST}:9980
+        reverse_proxy {$COLLABORA_HOST}:9980 {
+            # trusted_proxies placeholder
+        }
     }
     route /cool/* {
-        reverse_proxy {$COLLABORA_HOST}:9980
+        reverse_proxy {$COLLABORA_HOST}:9980 {
+            # trusted_proxies placeholder
+        }
     }
 
     # Onlyoffice
@@ -37,6 +47,7 @@
         reverse_proxy {$ONLYOFFICE_HOST}:80 {
             header_up X-Forwarded-Host {http.request.host}/onlyoffice
             header_up X-Forwarded-Proto https
+            # trusted_proxies placeholder
         }
     }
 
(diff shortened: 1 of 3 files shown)


Containers/apache/Caddyfile
```diff
@@ -4,6 +4,9 @@
     storage file_system {
         root /mnt/data/caddy
     }
+
+    # See https://github.com/nextcloud/all-in-one/issues/828
+    # trusted_proxies placeholder
```
2020Sanoj, line 9: grafik
I think this does not work here

szaimen: I see, that is unfortune but okay if it works

szaimen: hm.... this is getting worse then I thought since we need brackets now which are hard to parse and modify via sed...

szaimen: okay, seems like empty brackets are still accepted so not as problematic as I thought.

2020Sanoj (approved):
Should work (If the proxy connects over localhost / an IP from the private IP range)

I've tested manipulating the X-forwarded-for header using a local reverse proxy and a local DNS rewrite (`dig mydomain` showed the DNS rewrite and my browser was inside the logs of the local proxy)
And the manipulating didn't work, Nextcloud showed me the external IP, which my local proxy used to connect to my Nextcloud. So it should be not possible to trick Caddy with a manipulated X-forwarded-for header. 
But this also means that if nextcloud-aio is used behind a reverse proxy, which is not connecting to nextcloud-aio using a local / private range IP, the X-forwarded-for header wouldn't work.

szaimen:
Thanks for the review and approval! :)

Issue #834, opened 2022-06-18
Make collabora work behind Apache and Nginx


charredchar:
### How to use GitHub

* Please use the 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to show that you are affected by the same issue.
* Please don't comment if you have no relevant information to add. It's just extra noise for everyone subscribed to this issue.
* Subscribe to receive notifications on status change and new comments. 

### Steps to reproduce
1. Install Nextcloud AIO following the steps in `readme.md`. Enable the Collabora container from the AIO Interface.
2. Open any .docx or .odt file.

### Expected behavior 
Collarbora/Nextcloud Office loads the file in the browser

### Actual behavior 
I receive an error, "Document Loading failed. Failed to load Nextcloud Office - please try again later"

### Host OS 
Ubuntu Server 20.04 in a Windows Server Hyper-V VM.

#### Nextcloud AIO version 
Nextcloud AIO v1.4.0

#### Current channel 
Latest, attempted Beta with the same results.

#### Other valuable info 
I am currently running Nextcloud through an Apache Reverse Proxy located on another VM. I followed the setup under `reverse-proxy.md`  including the VirtualHosts and startup command. Everything except for Collabora works perfectly as expected.
I wouldn't mind testing without a reverse proxy at all, using just the server IP, for testing but I couldn't get through the AIO setup without a domain entered.
I've tried it in Edge, Chrome and Firefox. Adblock disabled.

AIO Apache Log; https://pastebin.com/7VNKQw29
AIO Nextcloud Log; https://pastebin.com/NcnBpz4C
Chrome Console while loading file; https://pastebin.com/YRxkR55L
AIO Collabora log is too large for Pastebin.

szaimen:
@Cr0iX am I correct that you are running AIO behind Apache? If yes, is it possible for you to reproduce this issue?

Cr0iX:
> @Cr0iX am I correct that you are running AIO behind Apache? If yes, is it possible for you to reproduce this issue?

I have just tried to run nextcloud-aio with collabora instead of onlyoffice, and i can confirm that the error "Document loading failed" is back. I've tried both .odt and .docx, and both have the same error.

Pulling the updates to see if stuff changes.
Still same error with latest docker versions.

szaimen:
I see. Thanks for providing the log!

Can you check if adding `ProxyPreserveHost On` makes it work? 
Thank you!

[...]

PR #835, opened 2022-06-20
apache documentation - make the push proxy work

szaimen:
Close part of https://github.com/nextcloud/all-in-one/issues/834

### DIFF
--- a/reverse-proxy.md
+++ b/reverse-proxy.md
@@ -20,7 +20,7 @@ In order to run Nextcloud behind a reverse proxy, you need to specify the port t
 
 <summary>click here to expand</summary>
 
-**Disclaimer:** It might be possible that the config below is not working 100% correctly, yet. Improvements to it are very welcome!
+**Disclaimer:** It might be possible that the config below is not working 100% correctly, yet. See e.g. https://github.com/nextcloud/all-in-one/issues/834. Improvements to it are very welcome!
 
 Add this as a new Apache site config:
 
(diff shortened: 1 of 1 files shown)


2020Sanoj (commented):
For me, collabora still doesn't work with this change.

szaimen:
> For me, collabora still doesn't work with this change.

So you are running it behind apache, too?

Or is this happening with other reverse proxies, too?

2020Sanoj:
> > For me, collabora still doesn't work with this change.
> 
> So you are running it behind apache, too?
> 
> Or is this happening with other reverse proxies, too?

I've stopped my NPM and updated my apache config with this change, started apache server and tested it. So I use the NPM, but have apache2 for testing also installed.
With the NPM it works. But with this apache config not.

szaimen:
> With the NPM it works. But with this apache config not.

Thanks for letting me know!

Yeah, we need to check what we can adjust so that websocket connections for collabora work behind apache, too.

szaimen:
https://httpd.apache.org/docs/current/mod/mod_rewrite.html#rewritecond

2020Sanoj:
> https://httpd.apache.org/docs/current/mod/mod_rewrite.html#rewritecond
> 
> Check what happens if these get removed:
> 
> https://github.com/nextcloud/all-in-one/blob/56948a81407b6d0c56cbb9b731a4555a9d448e3c/reverse-proxy.md?plain=1#L46-L47

Still same problem

2020Sanoj:
But Websocket in general works
grafik

szaimen:
Should check and test some things of https://github.com/nextcloud/integration_whiteboard#optionally-make-spacedeck-accessible-via-a-reverse-proxy-in-a-virtual-host

szaimen:
Push notifications do work now with this config but I was unfortunately not abel to make collabora work even after trying for a few hours. I think it does not work because collabora seemingly does not send the upgrade variable nor the connection variable...

szaimen:
merging this since it seems to fix at least this one issue...

PR #935, opened 2022-07-14
add docs on fail2ban config

HedgeShot:
fail2ban and docker don't play nice by default, need to add a rule in the config file to make sure docker doesn't bypass fail2ban jail.

Signed-off-by: HedgeShot <HedgeShot@users.noreply.github.com>


szaimen (approved):
Perfect, thanks! :)

PR #941, opened 2022-07-16
Improve docs for Cloudflare Tunnel setup

MiguelGuthridge:
I ran into an issue during my setup where I set up my Cloudflare Tunnel's configuration to point to `https://localhost:11000` instead of `http://localhost:11000` (due to another program I'm using requiring the opposite). This created a multitude of issues which I found challenging to debug. 

As such, I am proposing that the configuration specifies to make the tunnel use `http` instead of `https` to avoid confusion for other users.

If maintainers think that the project would benefit from it, I can also add in an example `config.json` that others can adapt to mk the setup process simpler for those who are newer to self-hosting.


szaimen (approved):
Thanks!

Issue #967, opened 2022-07-21
fix database import for cases where the owner is not one of the expected ones


eonrider:
### How to use GitHub

* Please use the 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to show that you are affected by the same issue.
* Please don't comment if you have no relevant information to add. It's just extra noise for everyone subscribed to this issue.
* Subscribe to receive notifications on status change and new comments. 

### Steps to reproduce
1. Perform step 12 of migration instructions. https://github.com/nextcloud/all-in-one/blob/v1.5.0/migration.md
2. See `ERROR:  role "ncadmin" does not exist` everywhere during database dump import

### Expected behavior 
The import should be successful.

### Actual behavior 
The import errors out because the database role does not exist.

### Host OS 
Raspberry Pi OS Lite (64-bit)

#### Nextcloud AIO version 
Nextcloud AIO v1.5.0

#### Current channel 
`latest-arm64`

#### Other valuable info 
It seems to me that the expected role is actually `nextcloud` as seen here https://github.com/nextcloud/all-in-one/blob/v1.5.0/php/containers.json#L72. Instead, my script tries to use `ncadmin` as that's what I used when I converted my MariaDB database to Postgres using `export PG_USER="ncadmin"` as suggested by the migration document.

szaimen:
Hi, thnaks for your report. Can you please run `grep 'Owner:'` on your database export and post the output here?

I think we rather need to adjust the things here: https://github.com/nextcloud/all-in-one/blob/e6a9fcd3c87f0ca055688b67c14c1d3651ee56b2/Containers/postgresql/start.sh#L68-L84

szaimen:
I suppose it should return `Owner: nextcloud` but not really sure. So that is why I am asking...

eonrider:
Yes. Here's just a snippet:
```
-- Name: oc_accounts; Type: TABLE; Schema: public; Owner: ncadmin
-- Name: oc_accounts_data; Type: TABLE; Schema: public; Owner: ncadmin
-- Name: oc_accounts_data_id_seq; Type: SEQUENCE; Schema: public; Owner: ncadmin
-- Name: oc_accounts_data_id_seq; Type: SEQUENCE OWNED BY; Schema: public; Owner: ncadmin
-- Name: oc_activity; Type: TABLE; Schema: public; Owner: ncadmin
[...]

PR #970, opened 2022-07-23
fix database import for cases where the owner is not one of the expec…

szaimen:
…ted ones

Fix https://github.com/nextcloud/all-in-one/issues/967


juliushaertl (approved):
Looks sane

PR #1010, opened 2022-08-14
Fix sed line in start.sh

flortsch:
Fix sed line in start.sh. Missing space throws an sed expression error and does not change the janus config.


juliushaertl (approved):

juliushaertl:
Nice catch, thanks for your pull request :+1:

PR #1021, opened 2022-08-15
Grammar fix: Replace "loose" with "lose"

andygeorge:
This replaces a few instances of "loose" with "lose", which is grammatically correct in these contexts.


szaimen (approved):
Thanks for your PR! Looks good to me!

PR #1022, opened 2022-08-15
Set collabora home option

juliushaertl:


szaimen:
Thanks for the PR Julius! What exactly does this do? Can you point me to some docs on this option?

szaimen (approved):

PR #1027, opened 2022-08-17
add docker healthchecks to the docker images

Zoey2936:
Added [Docker Health checks](https://docs.docker.com/engine/reference/builder/#healthcheck) to some containers.
[clamav](https://github.com/Cisco-Talos/clamav/blob/4a0382cf7a8e86e617b8f2e24f5ccb0e649f7f81/Dockerfile#L115) container already includes this.
And for the watchtower & borgbackup container I don't know how this could be done, and I think, because they only run for a short time, it wouldn't make much sense to add this check to them. 
In my test (I connected using bash, for example: `docker exec -it nextcloud-aio-collabora bash`, the executed the health check commands and then checked with `echo $?`) all checks reported Exit Code 0.


szaimen:
@Z0ey2022 Thanks for the PR! I'll have a look soon!

szaimen:
Short question: why do you need these health checks?

Z0ey2022:
> Short question: why do you/we need these health checks?

I don't need them directly, but I like it when I see that my containers report not only that they are running, but also show that they are healthy, so that I could see if the service inside the container is stopped or crashed or something like this
grafik

Maybe the mastercontainer could also check the status of the container from the docker socket and show the status also inside the aio panel or do something else with this information, but I don't know if this could work or how this could be implemented

Zoey2936:
I will wait for #1044 to be merged, so I can directly add a Healthcheck for elastic search

Containers/apache/Dockerfile
```diff
@@ -19,6 +19,7 @@ RUN set -ex; \
         openssl \
         netcat \
         dpkg-dev \
+        curl \
```
szaimen, line 22: isn't curl already in the image?

Containers/imaginary/Dockerfile
```diff
@@ -1,2 +1,15 @@
 # From https://github.com/h2non/imaginary/blob/master/Dockerfile
 FROM nextcloud/imaginary:20220822
+
+USER root
+RUN set -ex; \
+    \
+    apt-get update; \
+    apt-get install -y --no-install-recommends \
+        ca-certificates \
+        curl \
```
szaimen, line 10: So curl is not in the image?

Containers/mastercontainer/Dockerfile
```diff
@@ -27,6 +27,7 @@ RUN apt-get update; \
         sudo \
         dpkg-dev \
         netcat \
+        curl \
```
szaimen, line 30: here as well?

Containers/nextcloud/Dockerfile
```diff
@@ -113,6 +113,7 @@ RUN set -ex; \
     apk add --no-cache --virtual .fetch-deps \
         bzip2 \
         gnupg \
+        fcgi \
```
szaimen, line 116: do we really need this? doesn't curl work for localhost:9000?

Containers/talk/Dockerfile
```diff
@@ -67,3 +68,5 @@ ENV TALK_PORT=3478
 USER talk
 ENTRYPOINT ["start.sh"]
 CMD ["/usr/bin/supervisord", "-c", "/supervisord.conf"]
+
+HEALTHCHECK CMD (curl -skI localhost:8081 && curl -skI localhost:8188 && curl -skf --http0.9 localhost:4222 && nc -z localhost $TALK_PORT) || exit 1
```
szaimen, line 72: maybe?
```suggestion
HEALTHCHECK CMD (curl -skI localhost:8081 && curl -skI localhost:8188 && curl -skf --http0.9 localhost:4222 && curl -skI localhost:$TALK_PORT) || exit 1
```

szaimen (commented):
Thanks! LGTM in general! 
Just a few questions/regards.

Containers/apache/Dockerfile
Zoey2936: grafik

Containers/imaginary/Dockerfile
Zoey2936: grafik

Containers/talk/Dockerfile
Zoey2936: grafik

Containers/apache/Dockerfile
szaimen: all right!

Containers/imaginary/Dockerfile
szaimen: all right!

Containers/mastercontainer/Dockerfile
szaimen: cool!

Containers/talk/Dockerfile
szaimen: all right!

Containers/nextcloud/Dockerfile
```diff
@@ -252,3 +253,5 @@ RUN echo "root:$(openssl rand -base64 12)" | chpasswd
 
 USER www-data
 ENTRYPOINT ["/start.sh"]
+
+HEALTHCHECK CMD (cgi-fcgi -bind -connect localhost:9000 && curl -skI localhost:7867) || exit 1
```
szaimen, line 257: maybe? Or is that not safe enough?

```suggestion
HEALTHCHECK CMD (nc -z localhost 9000 && curl -skI localhost:7867) || exit 1
```

Zoey2936: This also works
I don't know if the cgi-fcgi check is better, what's your opinion?

szaimen: I would use netcat for now. We already have too much dependencies inside the nextcloud container...

Containers/fulltextsearch/Dockerfile
```diff
@@ -2,3 +2,5 @@
 FROM elasticsearch:7.17.6
 
 RUN elasticsearch-plugin install --batch ingest-attachment
+
+HEALTHCHECK CMD curl -skfI localhost:9200 || exit 1
```
szaimen, line 6: we don't need this as the base image already includes the healthcheck. See https://github.com/elastic/elasticsearch/blob/8b3293ddeb008459bf795cb4082beab3d654cd88/distribution/docker/src/docker/Dockerfile#L295

Containers/nextcloud/Dockerfile
Zoey2936: done

szaimen:
LGTM now! Please rebase on top of master and squash your changes. Afterwards I'll merge!

Zoey2936:
> LGTM now! Please rebase on top of master and squash your changes. Afterwards I'll merge!

Sorry, but I don't get them rebased, should I create a new PR with these changes?

szaimen:
> Sorry, but I don't get them rebased, should I create a new PR with these changes?

Fine by me if that is easier for you 👍

Containers/fulltextsearch/Dockerfile
Zoey2936: grafik
I think it is needed

szaimen: Strange! Seems like you are right.

Zoey2936: grafik
I think the reason is this if condition

szaimen: Can you add this back?

PR #1030, opened 2022-08-17
allow to adjust the max upload time

szaimen:

Close https://github.com/nextcloud/all-in-one/discussions/901


szaimen:
Seems like this will only work with https://github.com/nextcloud/server/pull/32216 which is currently only available for Nextcloud 25. Triggered a backport to stable24.

Issue #1011, opened 2022-08-14
Suggestion: Fallback to default port 3478 if TALK_PORT envvar is not set


flortsch:
Currently, the start.sh script in the Talk container generates an invalid janus config, if the TALK_PORT envvar is not set. This results in the janus process not starting within the container. I suggest to fallback to default port 3478, if this envvar is not set. This would also increase compatibility with older setups, where setting the TALK_PORT was not required.

Edit: the start.sh exits with error code 1 if e.g. NC_DOMAIN or other envvars are not set. If no default talk port is used, a check for TALK_PORT should also be included here.

szaimen:
We actually have fallback in place: https://github.com/nextcloud/all-in-one/blob/177e115b2bc5d180a06cfcba2e82811ed1e0aed2/php/src/Data/ConfigurationManager.php#L398-L403

Why do you need a fallback inside the container?

flortsch:
I have a setup of Nextcloud that I configured before the all-in-one suite existed. So I am using the official Docker Nextcloud container in combination with a standalone Talk container of the all-in-one-suite. The last update that added the TALK_PORT envvar broke my setup, because I didn't have it set and the container used a fixed port back then. This might not be a supported setup, but I thought I would report this here as a suggestion. Perhaps others have a similar setup, and when they update they will encounter the same problem.

szaimen:
I see. This is indeed not a supported setup. However I am willing to accept a PR that adds this check if you create it.

szaimen:
This is now fixed with v2.0.0 Beta. Testing and feedback is welcome! See https://github.com/nextcloud/all-in-one#how-to-switch-the-channel

PR #1035, opened 2022-08-19
Set default talk port inside Dockerfile

Zoey2936:
Should close https://github.com/nextcloud/all-in-one/issues/1011, because so the default env is set and when the talk container is launched by the mastercontainer, the env should be overridden by the env send from the mastercontainer.
If I have an image with an env (myenv=456) set inside the Dockerfile and then launch it with "-e myenv=123" the env is set to 123 inside the container,
I hope it's the same when the container is launched through the docker socket. But if yes, this change should help.

### DIFF
--- a/Containers/talk/Dockerfile
+++ b/Containers/talk/Dockerfile
@@ -61,6 +61,9 @@ RUN mkdir -p /etc/nats; \
 # Give root a random password
 RUN echo "root:$(openssl rand -base64 12)" | chpasswd
 
+# Set default talk port https://github.com/nextcloud/all-in-one/issues/1011
+ENV TALK_PORT=3478
+
 USER talk
 ENTRYPOINT ["start.sh"]
 CMD ["/usr/bin/supervisord", "-c", "/supervisord.conf"]


szaimen:
Thanks! This is indeed a valid way to close the mentioned issue. Will merge this after 1.7.1 stable is released!

Containers/talk/Dockerfile
```diff
@@ -61,6 +61,9 @@ RUN mkdir -p /etc/nats; \
 # Give root a random password
 RUN echo "root:$(openssl rand -base64 12)" | chpasswd
 
+# Set default talk port https://github.com/nextcloud/all-in-one/issues/1011
+ENV TALK_PORT=3478
```
szaimen, line 65: Can you please adjust this to this? (I think it is required to write it like this)

```suggestion
ENV TALK_PORT 3478
```

Zoey2936: Oh, I see both is possible

szaimen: all right! then fine by me!

szaimen (approved):

PR #1038, opened 2022-08-21
add section about local setup without public domain to local-instance.md

woetenwondla:
As I had some difficulties setting up nextcloud locally without public domain and without public internet access, 
I added a section to local-instance.md, which explains the key-points for this kind of setup

Signed-off-by: Jan Aigner <jan.aigner7@gmail.com>


szaimen:
Hello, thanks for your PR!

Unfortunately are self-signed certificates [not supported](https://github.com/nextcloud/all-in-one#are-self-signed-certificates-supported-for-nextcloud) so we will not add this to this documentation as you will most likely run into issues with at least some of the components due to the usage of self-signed certificates.

However feel free to add your guide as a new entry to this discussion: https://github.com/nextcloud/all-in-one/discussions/588

Issue #1036, opened 2022-08-20
Feature Req: Add Elasticsearch container for Full Text Search


revr3nd:
Elasticsearch is used by the Full Text Search addon and is extremely useful to provide content searching.
Given that Elastic search is already easily run within Docker, I think it's a fairly easy container app to add to the list for integration with Nextcloud AIO in single node setups.

The custom docker files make Elasticsearch deployment even easier as it does require an additional plugin to be installed, which requires an additional docker file command to run: '/usr/share/elasticsearch/bin/elasticsearch-plugin install --batch ingest-attachment'

I am currently running elasticsearch alongside the AIO deployment with:
1. docker run -d --name elasticsearch --net nextcloud-aio -p 9200:9200 -e "discovery.type=single-node" elasticsearch:7.17.5
2. docker exec -t elasticsearch /usr/share/elasticsearch/bin/elasticsearch-plugin install --batch ingest-attachment
3. Open Nextcloud and install the Full text search - Elasticsearch Platform Addon
4. Open Nextcloud settings, under Full Test search, enable Elasticsearch as the search platform, servlet address, and index name

szaimen:
Hey, thanks for the suggestion! 
Just two small questions:
- do you know if it works with elastisearch 8 as well?
- which directory contains the elastisearch cache and should be made persistant?

Z0ey2022:
>     * do you know if it works with elastisearch 8 as well?

https://github.com/nextcloud/fulltextsearch_elasticsearch/issues/178

Z0ey2022:
>     * which directory contains the elastisearch cache and should be made persistant?

If I have read correctly, its `/usr/share/elasticsearch/data`

revr3nd:
Ok, I had to go into the wayback on how this worked as it had been a long time since I had set this up. There actually does need to be a command that is run on startup every reboot in order to ensure the indexing continues indefinitely. The Fulltextsearch docs gives a better breakdown of what is required here: https://github.com/nextcloud/fulltextsearch/wiki/Basic-Installation

1. Create elasticsearch container: docker run -d --name elasticsearch --net nextcloud-aio -p 9200:9200 -e "discovery.type=single-node" -v esdata:/usr/share/elasticsearch/data elasticsearch:7.17.5
2. Install ingest plugin: docker exec -t elasticsearch /usr/share/elasticsearch/bin/elasticsearch-plugin install --batch ingest-attachment
3. Install the following addons in Nextcloud
    - Full text search
[...]

Issue #1041, opened 2022-08-22
publish new images after recent PRs are merged and tested


szaimen:
### Follow up to https://github.com/nextcloud/all-in-one/issues/1012
(add the most recent ticket for a new release here)

### Tested:
(add PRs below that are potentially breaking and need to be well tested)
- [x] https://github.com/nextcloud/all-in-one/commit/ae29ce409e0afad7babeac5f9b16d848c9dd4783
- [x] https://github.com/nextcloud/all-in-one/pull/1024 (fixed with https://github.com/nextcloud/all-in-one/commit/20892c2d40c9bc9633ffaeffeb1c076ef8cf1ff5)
- [x] https://github.com/nextcloud/all-in-one/pull/1026 (fixed with https://github.com/nextcloud/all-in-one/commit/096be7107a4a4e708e120d10ddd00db75eff0532)
- [x] https://github.com/nextcloud/all-in-one/pull/1058 (fixed with https://github.com/nextcloud/all-in-one/pull/1060 and https://github.com/nextcloud/all-in-one/pull/1059)
- [x] https://github.com/nextcloud/all-in-one/pull/1028
- [x] https://github.com/nextcloud/all-in-one/pull/1029 (fixed with https://github.com/nextcloud/all-in-one/commit/2e1e53e9d3ce11d0b390c9fe2977d6b00746666b  and https://github.com/nextcloud/all-in-one/commit/7b4876c681abdbd6e828bbcd7c82b68b47e0a6f9)
- [x] https://github.com/nextcloud/all-in-one/pull/1030 (Currently not working due to https://github.com/nextcloud/all-in-one/pull/1030#issuecomment-1223110690 will be postponed in https://github.com/nextcloud/all-in-one/issues/1048; fixed with https://github.com/nextcloud/all-in-one/commit/c90ee1c4dba4a8fdaa232ba8f86cd0e2fda00fb2)
- [x] https://github.com/nextcloud/all-in-one/pull/1035
- [x] https://github.com/nextcloud/all-in-one/pull/1042
- [x] https://github.com/nextcloud/all-in-one/commit/09003fcce19a871797ee6629aa69b6b6a3b9ca19
- [x] https://github.com/nextcloud/all-in-one/pull/1044 (fixed with https://github.com/nextcloud/all-in-one/pull/1065)
[...]

PR #1044, opened 2022-08-22
add fulltextsearch as option

szaimen:
Close https://github.com/nextcloud/all-in-one/issues/1036

- [x] needs https://github.com/nextcloud-releases/all-in-one/pull/28

### DIFF
--- a/Containers/fulltextsearch/Dockerfile
+++ b/Containers/fulltextsearch/Dockerfile
@@ -1,0 +1,4 @@
+# Probably from here https://github.com/elastic/elasticsearch/blob/main/distribution/docker/src/docker/Dockerfile
+FROM elasticsearch:7.17.5
+
+RUN elasticsearch /usr/share/elasticsearch/bin/elasticsearch-plugin install --batch ingest-attachment
--- a/Containers/nextcloud/activate-fulltextsearch.sh
+++ b/Containers/nextcloud/activate-fulltextsearch.sh
@@ -2,8 +2,8 @@
 
 if [ "$FULLTEXTSEARCH_ENABLED" != yes ]; then
     # Basically sleep for forever if fulltextsearch is not enabled
-    sleep 365d
+    sleep inf
 fi
 echo "Activating fulltextsearch..."
 php /var/www/html/occ fulltextsearch:live -q
-sleep 365d
+sleep inf
--- a/Containers/nextcloud/activate-fulltextsearch.sh
+++ b/Containers/nextcloud/activate-fulltextsearch.sh
@@ -1,0 +1,9 @@
+#!/bin/bash
+
+if [ "$FULLTEXTSEARCH_ENABLED" != yes ]; then
+    # Basically sleep for forever if fulltextsearch is not enabled
+    sleep 365d
+fi
+echo "Activating fulltextsearch..."
+php /var/www/html/occ fulltextsearch:live -q
+sleep 365d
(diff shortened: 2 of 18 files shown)


Containers/nextcloud/entrypoint.sh
```diff
@@ -423,5 +423,47 @@ if version_greater "24.0.0.0" "$installed_version"; then
     fi
 fi
 
+# Fulltextsearch
+if [ "$FULLTEXTSEARCH_ENABLED" = 'yes' ]; then
+    while ! nc -z "$FULLTEXTSEARCH_HOST" 9200; do
+        echo "waiting for OnlyOffice to become available..."
```
Zoey2936, line 429: I think this is not OnlyOffice

szaimen: fixed

Containers/nextcloud/activate-fulltextsearch.sh
```diff
@@ -0,0 +1,9 @@
+#!/bin/bash
+
+if [ "$FULLTEXTSEARCH_ENABLED" != yes ]; then
+    # Basically sleep for forever if fulltextsearch is not enabled
+    sleep 365d
+fi
+echo "Activating fulltextsearch..."
+php /var/www/html/occ fulltextsearch:live -q
+sleep 365d
```
Zoey2936, line 9:
```suggestion
    sleep inf
fi
echo "Activating fulltextsearch..."
php /var/www/html/occ fulltextsearch:live -q
sleep inf
```
Why not directly sleep forever?
See: https://www.gnu.org/software/coreutils/manual/html_node/sleep-invocation.html

szaimen: good point! didn't know that this is possible...

szaimen: fixed

Containers/fulltextsearch/Dockerfile
```diff
@@ -0,0 +1,4 @@
+# Probably from here https://github.com/elastic/elasticsearch/blob/main/distribution/docker/src/docker/Dockerfile
+FROM elasticsearch:7.17.5
+
+RUN elasticsearch /usr/share/elasticsearch/bin/elasticsearch-plugin install --batch ingest-attachment
```
Zoey2936, line 4: I think this does not build
grafik

If I replace the RUN command with: `elasticsearch-plugin install --batch ingest-attachment`
It outputs:
grafik

szaimen: you are right. Fixed with https://github.com/nextcloud/all-in-one/commit/6c63adc37ff0f755375cea2a518bb0adc11560f4

szaimen:
@Zoey2936 This is the last remaning thing that I was not yet able to confirm that it works from [this list](https://github.com/nextcloud/all-in-one/issues/1041). Did you tried it out already? I am not sure what I am currently missing why it doesn't work...

Zoey2936:
> @Zoey2936 This is the last remaning thing that I was not yet able to confirm that it works from [this list](https://github.com/nextcloud/all-in-one/issues/1041). Did you tried it out already? I am not sure what I am currently missing why it doesn't work...

I tried but it seems that it does not work
Running `docker exec -it nextcloud-aio-nextcloud php occ fulltextsearch:index` did something, but I don't know what exactly

Zoey2936:
grafik

szaimen:
Thanks for checking! I've also asked a colleague that has more knowledge regarding this. Let's see.

Zoey2936:
grafik
Now it reports this error

szaimen:
Yeah, seems like this is not really stable and usable currently... Will see on monday if I am able to fix this with the help of a colleague...

szaimen:
Fixed with https://github.com/nextcloud/all-in-one/pull/1065

PR #1052, opened 2022-08-23
Fix bad error message in start.sh

flortsch:
Fix bad error message in start.sh which seems to be a copy/paste error from the previous if check.

Signed-off-by: Florian Latifi <mail@florian-latifi.at>


szaimen (approved):
thanks, LGTM!

PR #1259, opened 2022-10-13
Update nginx section of reverse-proxy.md

Klaus15:
added nginx.conf code snippet to get a working websocket config

Signed-off-by: Klaus15 <le_kluus@web.de>


szaimen:
Thanks Klaus! :)

---

@Zoey2936 what do you think about also adding something about lets encrypt which is required anyway to the section like we already do with the Apache documentation?

szaimen (approved):
LGTM

PR #1271, opened 2022-10-16
Added documentation for Traefik reverse proxy using docker-compose.ym…

Lorthiz:
Extended documentation for Traefik reverse proxy

### DIFF
--- a/reverse-proxy.md
+++ b/reverse-proxy.md
@@ -274,6 +274,54 @@ Of course you need to modify `<your-nc-domain>` in the nextcloud.toml to the dom
 
 </details>
 
+### Traefik 2.x(via docker-compose.yml labels)
+
+<details>
+
+<summary>click here to expand</summary>
+
+The `docker-compose.yml` file prepared by the team works pretty well out-of-the box, but some additional configuration is needed to make it work with Traefik.
+
+Add your configuration using labels in deploy section. The `deploy` section is at the same level as `image` and `environment`
+
+   ```yaml
+    deploy:
+      labels:
+        # enable traefik for this service
+        - traefik.enable=true
+   
+        # needed for Traefik to communicate with Nextcloud instance - mostly needed in docker swarm mode to pass overlay network or pass custom networks
+        # if you are running simple config and encounter some problems try removing this property
+        - traefik.docker.network=<your-network> 
+        
+        # you can pass your own domain here as Host(`nextcloud.example.com`)
+        - traefik.http.routers.nextcloud-rtr.rule=Host(`<your-nc-domain>`)
+   
+        # here is just some standard Traefik labels port will stay the same, the entrypoints you can find in your static configuration
+        - traefik.http.services.nextcloud.loadbalancer.server.port=80
+        - traefik.http.routers.nextcloud-rtr.entrypoints=<entrypoint(s)>
+        
+        # labels below are used for configuring TLS(access via HTTPS) - cert resolvers you will find in static configuration
+        - traefik.http.routers.nextcloud-rtr.tls=true
+        - traefik.http.routers.nextcloud-rtr.tls.certresolver=<your-cert-resolver>
+   ```
+   
+---
+
+### Known issues
+
+1. Strict mode, no HTTP connection allowed!
+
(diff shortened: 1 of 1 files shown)


szaimen:
Hi, did you test this already and verified that you get a working Nextcloud with that?

szaimen:
Also did you see this?
Screenshot_20221016-213915_Brave

Lorthiz:
> Hi, did you test this already and verified that you get a working Nextcloud with that?

Yes it's running fine on my cluster atm - i did not had much problems with it, but I also did not spend much time in nextcloud yet. But things like changing settings, adding users, uploading files work fine with this config.

I kind of don't understand the second question since post is mentioning Apache being spawned by 'mastercontainer' but the way I see it we are adding the labels to the master container itself and apache should not have anything to do with it. 

The way traefik should work here are:

We are sending requests to Traefik using DNS mapping - all the requests to our domain and it's subdomains are pointed at machine/cluster running traefik. Then traefik using host matching is seeking the best route/service, when it finds it it asks the proper url(or other docker container) as a user and can add some stuff like additional headers, apply some security to the endpoints etc, but it's still just doing the same HTTP requests as the normal user would do and then it passes it back to the user rewrapped in HTTPS or with some additional headers.

If there is possibility that my config is in truth incorrect I will try to check it and get back to you. Meaby I even just prepare whole project having only DB + Nextcloud + Traefik and if that will be working we may be able to provide users with config ready to go for nextcloud with traefik.

szaimen:
And you tested these labels with AIO or with a different cluster which is not based on AIO?

szaimen:
For me it is still unclear how that should work technically... As the apache container does not have the labels in that case... Or can you somehow point traefik towards a sepcific port like the chosen apache_port on the host?

reverse-proxy.md
```diff
@@ -274,6 +274,54 @@ Of course you need to modify `<your-nc-domain>` in the nextcloud.toml to the dom
...
+   
+        # needed for Traefik to communicate with Nextcloud instance - mostly needed in docker swarm mode to pass overlay network or pass custom networks
+        # if you are running simple config and encounter some problems try removing this property
+        - traefik.docker.network=<your-network> 
+        
+        # you can pass your own domain here as Host(`nextcloud.example.com`)
+        - traefik.http.routers.nextcloud-rtr.rule=Host(`<your-nc-domain>`)
+   
+        # here is just some standard Traefik labels port will stay the same, the entrypoints you can find in your static configuration
+        - traefik.http.services.nextcloud.loadbalancer.server.port=80
```
szaimen, line 301: If yoi use port 11000 as in all the other examples. Does it make it work?

reverse-proxy.md
```diff
@@ -274,6 +274,54 @@ Of course you need to modify `<your-nc-domain>` in the nextcloud.toml to the dom
...
+
+1. Strict mode, no HTTP connection allowed!
+
+This problem will manifest itself during logins on Android via HTTPS endpoint and it will not allow you to login. To fix it simply add `OVERWRITEPROTOCOL` variable to your `docker-compose.yml` file in `environment` section: 
+
+   ```yaml
+     environment:
+        # your properties like MYSQL_PASSWORD MYSQL_DATABASE MYSQL_USER etc. will also be here
+        - OVERWRITEPROTOCOL=https
+   ```
```
szaimen, line 321: As you may noticed are we not adding these to the mastercontainer itself since nextcloud and co are not included in this one container but in additonal ones that get launched by it.

szaimen:
To me it looks like you havent read this: https://github.com/nextcloud/all-in-one#how-does-it-work

reverse-proxy.md
Lorthiz: But why would I want to use that port? Nextcloud as a service exposes port 80 and the goal is to put nextcloud behind Traefik soo...

szaimen: Where do you got the idea that nextcloud exposes port 80?

Lorthiz:
Okay I see... I must have been searching for fix of my problem >no HTTP connection allowed!< and I landed at this documentation... The docker-compose.yml I used initially was from here:

https://github.com/docker-library/docs/tree/master/nextcloud#base-version---apache

And now I see that those are 2 different projects...
I am really sorry for the confusion @szaimen. I am pretty new to Nextcloud and I was not even awere there were more repos with different configs etc. 

So I am closing this PR and yeah... Sorry again for this mistake ;/

PR #1274, opened 2022-10-16
rework session deduplication

szaimen:
Fix https://github.com/nextcloud/all-in-one/discussions/1272

### DIFF
--- a/php/src/Auth/AuthManager.php
+++ b/php/src/Auth/AuthManager.php
@@ -3,6 +3,8 @@
 namespace AIO\Auth;
 
 use AIO\Data\ConfigurationManager;
+use AIO\Data\DataConst;
+use \DateTime;
 
 class AuthManager {
     private const SESSION_KEY = 'aio_authenticated';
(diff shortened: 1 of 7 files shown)


php/src/Auth/AuthManager.php
```diff
@@ -21,6 +23,14 @@ public function CheckToken(string $token) : bool {
     }
 
     public function SetAuthState(bool $isLoggedIn) : void {
+
+        if (!$this->IsAuthenticated() && $isLoggedIn === true) {
+            $date = new DateTime();
+            $dateTime = $date->getTimestamp();
+            $_SESSION['date_time'] = $dateTime;
+            file_put_contents(DataConst::GetSessionDateFile(), (string)$dateTime);
```
juliushaertl, line 31: The session cleanup mechanism seems quite complex, but looks good to me for now.

In general, how about using an in memory cache for this with APCu. You could use this to store the current session id and then handle error messages and logout of other sessions all from the PHP side. (Just a suggestion, not a blocker of course)

juliushaertl (approved):

php/src/Auth/AuthManager.php
szaimen: you are right. However I don't know enough about apcu to do this on my own. So I rather stick with this solution for now 👍

PR #1297, opened 2022-10-23
Update traefik documentation

QuentinLemCode:
- Use redirectScheme instead of redirectSSL deprecated from traefik v2.5
- fixed middleware name in routers configuration
- added redirect regex for CalDAV or CardDAV  (https://docs.nextcloud.com/server/25/admin_manual/configuration_server/reverse_proxy_configuration.html)
- updated middleware names for better consistency with the docs
- added some security headers
- added docker-compose configuration

Signed-off-by: Quentin Lemeasle <quentinlemcode@gmail.com>

### DIFF
--- a/reverse-proxy.md
+++ b/reverse-proxy.md
@@ -229,6 +229,10 @@ Of course you need to modify `<your-nc-domain>` to the domain on which you want 
 
 **Disclaimer:** It might be possible that the config below is not working 100% correctly, yet. Improvements to it are very welcome!
 
+ 
+#### Toml File configuration
+    
+    
 1. Add a `nextcloud.toml` to the Treafik rules folder with the following content:
 
     ```toml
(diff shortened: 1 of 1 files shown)


Zoey2936:
The redirect regex is not needed, because it is handled by Caddy.
Working with labels will only work in manual mode, so maybe add a note.
The security headers are already handled by caddy and Nextcloud itself, so I think it is not a good Idea to overwrite them.
Some notes from me, at the moment I can't review/test this further.

QuentinLemCode:
> The redirect regex is not needed, because it is handled by Caddy.
> Working with labels will only work in manual mode, so maybe add a note.
> The security headers are already handled by caddy and Nextcloud itself, so I think it is not a good Idea to overwrite them.
> Some notes from me, at the moment I can't review/test this further.

Well, it's a configuration file for traefik.
IDK if traefik and caddy can work together, but this configuration will work for a traefik reverse proxy

QuentinLemCode:
Ohh 🤦
I haven't noticed that I was in the all-in-one repo and thought it was configuration for general NextCloud
I close this one

PR #1307, opened 2022-10-24
use the add commands to add talk settings

szaimen:
Close https://github.com/nextcloud/all-in-one/discussions/1088

### DIFF
--- a/Containers/nextcloud/entrypoint.sh
+++ b/Containers/nextcloud/entrypoint.sh
@@ -399,14 +399,15 @@ if [ "$TALK_ENABLED" = 'yes' ]; then
         php /var/www/html/occ app:update spreed
     fi
     # Based on https://github.com/nextcloud/spreed/issues/960#issuecomment-416993435
-    if ! php /var/www/html/occ talk:turn:list --output="plain" | grep -q "$NC_DOMAIN:$TALK_PORT"; then
+    if [ -z "$(php /var/www/html/occ talk:turn:list --output="plain")" ]; then
         php /var/www/html/occ talk:turn:add "$NC_DOMAIN:$TALK_PORT" "udp,tcp" --secret="$TURN_SECRET"
     fi
-    if ! php /var/www/html/occ talk:stun:list --output="plain" | grep -q "$NC_DOMAIN:$TALK_PORT"; then
+    if php /var/www/html/occ talk:stun:list --output="plain" | grep -oP '[a-zA-Z.:0-9]+' | grep -q "^stun.nextcloud.com:443$"; then
         php /var/www/html/occ talk:stun:add "$NC_DOMAIN:$TALK_PORT"
+        php /var/www/html/occ talk:stun:delete "stun.nextcloud.com:443"
     fi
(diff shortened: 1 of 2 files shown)


Containers/nextcloud/entrypoint.sh
```diff
@@ -396,12 +396,16 @@ if [ "$TALK_ENABLED" = 'yes' ]; then
...
-    php /var/www/html/occ config:app:set spreed signaling_servers --value="$SIGNALING_SERVERS" --output json
+    # Based on https://github.com/nextcloud/spreed/issues/960#issuecomment-416993435
+    if ! php /var/www/html/occ talk:turn:list --output="plain" | grep -q "$NC_DOMAIN:$TALK_PORT"; then
+        php /var/www/html/occ talk:turn:add "$NC_DOMAIN:$TALK_PORT" "udp,tcp" --secret="$TURN_SECRET"
+    fi
+    if ! php /var/www/html/occ talk:stun:list --output="plain" | grep -q "$NC_DOMAIN:$TALK_PORT"; then
+        php /var/www/html/occ talk:stun:add "$NC_DOMAIN:$TALK_PORT"
+    fi
+    if ! php /var/www/html/occ talk:signaling:list --output="plain" | grep -q "https://$NC_DOMAIN/standalone-signaling/"; then
+        php /var/www/html/occ talk:signaling:add "https://$NC_DOMAIN/standalone-signaling/" "$SIGNALING_SECRET" --validate-ssh-certificate
```
szaimen, line 407: Question: what will happen if a specific backend is already addedd if I simply add backends without checking if they are already added?

nickvergessen: It will end up twice in the list. Also note that providing multiple servers for each might have a negative performance impact when starting a call, I would even argue when a server is already specified don't do anything. Because if someone configured an external turn server having a second turn that is not as well configured might be disturbing to debug.

Also you misnamed the verify parameter?

```suggestion
    if ! php /var/www/html/occ talk:turn:list --output="plain" | grep -q "$NC_DOMAIN:$TALK_PORT"; then
        php /var/www/html/occ talk:turn:add "$NC_DOMAIN:$TALK_PORT" "udp,tcp" --secret="$TURN_SECRET"
    fi
    if ! php /var/www/html/occ talk:stun:list --output="plain" | grep -q "$NC_DOMAIN:$TALK_PORT"; then
        php /var/www/html/occ talk:stun:add "$NC_DOMAIN:$TALK_PORT"
    fi
    if ! php /var/www/html/occ talk:signaling:list --output="plain" | grep -q "https://$NC_DOMAIN/standalone-signaling/"; then
        php /var/www/html/occ talk:signaling:add "https://$NC_DOMAIN/standalone-signaling/" "$SIGNALING_SECRET" --verify
```

szaimen: good point! All fixed with https://github.com/nextcloud/all-in-one/pull/1307/commits/836ff799b063bd3f76f004b60fc480978f47ff8d

szaimen:
@nickvergessen does this look good to now? :)
Testing is not needed as I am doing that 👍

nickvergessen (approved):
🐘

szaimen:
Thanks Joas! :)

PR #1336, opened 2022-10-28
optimize nginx reverse proxy documentation

Zoey2936:
With this change, it is not needed to change the main http block (tested on plain nginx install from buster-backports)


szaimen (approved):
Thanks! :)

PR #1344, opened 2022-10-31
Make sure that the sites are really gone

szaimen:


Zoey2936 (approved):
running a2dissite already deletes the link in /etc/apache2/sites-enabled, which is a link to the same file in the /etc/apache2/sites-available folder and this folder is ignored by apache, so I think deleting this files will change nothing. 

But I think it is not a problem to double delete them, for the case that a2dissite does not work

PR #1370, opened 2022-11-07
session should not get cleared automatically before 24h

szaimen:


php/public/index.php
```diff
@@ -22,6 +22,9 @@
 // Auto logout on browser close
 ini_set('session.cookie_lifetime', '0');
 
+# Keep session for 24h max
+ini_set('session.gc_maxlifetime', '86400');
```
szaimen, line 26: this helps keeping the session open however not sure if it is he desired behaviour. Need to think about it...

juliushaertl (approved):

Issue #1162, opened 2022-09-16
Think about a solution for adding external dependencies into the Nextcloud container


szaimen:
E.g. for https://github.com/nextcloud/all-in-one/discussions/751 and https://github.com/nextcloud/all-in-one/discussions/1159

szaimen:
Thanks for the input! Agree, no of them is perfrect unfortunately...

szaimen:
@Zoey2936 What do you think about https://github.com/nextcloud/all-in-one/tree/enh/1162/allow-to-add-dependencies#how-to-install-external-dependencies-into-the-nextcloud-container-permanently?

Zoey2936:
I think it should work, but how do you want to check that the container first starts, when the apks are installed?

szaimen:
Simple: we create a file that the install_dependencies script checks for and crwates after running the first time. If the file is there we exit directly

Zoey2936:
And how do you want to handle the first point? (That the entrypoint first continues if the packages are installed?)

szaimen:
The entrypoint then checks for this as well...

szaimen:
I will think about the implementation details later but good to hear that you think it would work!

Zoey2936:
> The entrypoint then checks for this as well...

So again:
If the deps are successfully installed, it created a file.
If the file exists and cron gets executed again, it will exit.
And the entrypoint will only continue if this file exists.
1. But what if the file will never be created, because the user don't want to install additional deps, what will the entrypoint then do? Because if it waits at least five minutes this could be annoying, also the installation of the packages could take longer than five minutes.
2. Or what will the entrypoint do if the installation of the deps aren't installed successfully? (This could also be done using a lock file or something like this)

I know it is still WIP, but I don't understand how at least the first question can be resolved with the current logic of this implementation? Or did I misunderstand the logic?

szaimen:
> * But what if the file will never be created, because the user don't want to install additional deps, what will the entrypoint then do? Because if it waits at least five minutes this could be annoying, 

I guess we need to create a file in a persistant storage somewhere after running thr scripr successfully one time. If that is present we trigger the logic. If not we dont.

> also the installation of the packages could take longer than five minutes.

We need two files for that: one that is created while we install the dependencies and removed adter it is succeasful. The other one gets created when it was successful. Then entrypoint checks if the successful file is there and waits max 5min but waits indefinitely for the install file. 

[...]

PR #1377, opened 2022-11-08
allow to add dependencies and php extensions into the Nextcloud container

szaimen:
Close https://github.com/nextcloud/all-in-one/issues/1162


szaimen:
@Zoey2936 WDYT? :)

Containers/nextcloud/start.sh
```diff
@@ -36,6 +36,34 @@ if ! [ -f "$NEXTCLOUD_DATA_DIR/this-is-a-test-file" ]; then
 fi
 sudo -u www-data rm -f "$NEXTCLOUD_DATA_DIR/this-is-a-test-file"
 
+# Install additional dependencies
+if [ -n "$ADDITIONAL_APKS" ]; then
+    if ! [ -f "/additional-apks-are-installed" ]; then
+        read -ra ADDITIONAL_APKS_ARRAY <<< "$ADDITIONAL_APKS"
+        for app in "${ADDITIONAL_APKS_ARRAY[@]}"; do
+            apk add "$app"
```
Zoey2936, line 44: Why not also:
```suggestion
            apk add --no-cache "$app"
```

Zoey2936 (commented):
I think it should work, but I think it would also be good to have an option to install packages using
```
npm install --global <package> - requieres to have npm installed using apk
yarn global add <package> - requieres to have yarn installed using apk
pip install <packages> - requieres to have py3-pip installed using apk
```
Because for example the NCDownloader app requires installing yt-dlp using pip (which requires gcc, musl-dev and py3-pip to be installed using apk)
I currently don't know if any npm/yarn packages is needed by any nextcloud app existing, but I think at some point in the future it could be also needed

szaimen:
> Because for example the NCDownloader app requires installing yt-dlp using pip (which requires gcc, musl-dev and py3-pip to be installed using apk)

https://pkgs.alpinelinux.org/packages?name=youtube-dl&branch=v3.16&repo=&arch=&maintainer=

Zoey2936:
> > Because for example the NCDownloader app requires installing yt-dlp using pip (which requires gcc, musl-dev and py3-pip to be installed using apk)
> 
> https://pkgs.alpinelinux.org/packages?name=youtube-dl&branch=v3.16&repo=&arch=&maintainer=

There is a difference between youtube-dl and yt-dlp

szaimen:
> There is a difference between [youtube-dl](https://github.com/ytdl-org/youtube-dl) and [yt-dlp](https://github.com/yt-dlp/yt-dlp)
> And [NCDownloader referces to yt-dlp](https://github.com/shiningw/ncdownloader#installing-aria2-and-yt-dlp-in-ubuntu)

https://pkgs.alpinelinux.org/packages?name=yt-dlp&branch=v3.16&repo=&arch=&maintainer=

Zoey2936:
Then apk packages and php extension should be enough, then there is only the `--no-cache` question, if it is needed

Containers/nextcloud/start.sh
szaimen: sounds good, will apply that later!

Containers/nextcloud/start.sh
```diff
@@ -36,6 +36,34 @@ if ! [ -f "$NEXTCLOUD_DATA_DIR/this-is-a-test-file" ]; then
...
+if [ -n "$ADDITIONAL_PHP_EXTENSIONS" ]; then
+    if ! [ -f "/additional-php-extensions-are-installed" ]; then
+        read -ra ADDITIONAL_PHP_EXTENSIONS_ARRAY <<< "$ADDITIONAL_PHP_EXTENSIONS"
+        for app in "${ADDITIONAL_PHP_EXTENSIONS_ARRAY[@]}"; do
+            if [ "$app" = imagick ]; then
+                pecl install imagick-3.7.0
+                docker-php-ext-enable imagick
+            else
+                pecl install "$app"
+                docker-php-ext-enable "$app"
```
szaimen, line 60: It seems like that some extensions are not on pecl but instead delivered via core. If pecl fails, we need to run something like https://github.com/nextcloud/all-in-one/blob/b027cdd2933ede2f072fb4259147d32cc0276572/Containers/nextcloud/Dockerfile#L49 instead

szaimen: or maybe try to install it via core and afterwards via pecl. Need to figure out the best order.

szaimen: I guess best is to check if part of core by looking into `/usr/src/php/ext/`. If yes, install using docker-php-ext-install, if not, install using pecl.

docker-compose.yml
```diff
@@ -28,6 +28,8 @@ services:
       # - NEXTCLOUD_MAX_TIME=3600 # Can be adjusted if you need more. See https://github.com/nextcloud/all-in-one#how-to-adjust-the-max-execution-time-for-nextcloud
       # - TRUSTED_CACERTS_DIR=/path/to/my/cacerts # CA certificates in this directory will be trusted by the OS of the nexcloud container (Useful e.g. for LDAPS) See See https://github.com/nextcloud/all-in-one#how-to-trust-user-defiend-certification-authorities-ca
       # - COLLABORA_SECCOMP_DISABLED=false # Setting this to true allows to disable Collabora's Seccomp feature. See https://github.com/nextcloud/all-in-one#how-to-disable-collaboras-seccomp-feature
+      # - NEXTCLOUD_ADDITIONAL_APKS=imagick # This allows to add additional packages to the Nextcloud container permanently.
+      # - NEXTCLOUD_ADDITIONAL_PHP_EXTENSIONS=imagick # This allows to add additional php extensions to the Nextcloud container permanently. Default is imagick but can be overwritten by modifying this value.
```
szaimen, line 32: we coudl add a link to the readme here

Containers/nextcloud/start.sh
```diff
@@ -36,6 +36,34 @@ if ! [ -f "$NEXTCLOUD_DATA_DIR/this-is-a-test-file" ]; then
...
+fi
+
+# Install additional php extensions
+if [ -n "$ADDITIONAL_PHP_EXTENSIONS" ]; then
+    if ! [ -f "/additional-php-extensions-are-installed" ]; then
+        read -ra ADDITIONAL_PHP_EXTENSIONS_ARRAY <<< "$ADDITIONAL_PHP_EXTENSIONS"
+        for app in "${ADDITIONAL_PHP_EXTENSIONS_ARRAY[@]}"; do
+            if [ "$app" = imagick ]; then
+                pecl install imagick-3.7.0
+                docker-php-ext-enable imagick
```
szaimen, line 57:
```suggestion
                echo "Installing Imagick via PECL..."
                pecl install imagick-3.7.0 >/dev/null
                docker-php-ext-enable imagick >/dev/null
```

Containers/nextcloud/start.sh
szaimen, line 60:
```suggestion
                echo "Installing $app via PECL..."
                pecl install "$app" >/dev/null
                docker-php-ext-enable "$app" >/dev/null
```

Containers/nextcloud/start.sh
szaimen:
```suggestion
            echo "Installing $app via apk..."
            apk add --no-cache "$app" >/dev/null
```

.github/workflows/nextcloud-update.yml
```diff
@@ -57,8 +57,8 @@ jobs:
             | sort -V \
             | tail -1
         )"
-        sed -i "s|pecl install imagick.*\;|pecl install imagick-$imagick_version\;|" ./Containers/nextcloud/Dockerfile
-        
+        sed -i "s|pecl install imagick.*|pecl install imagick-$imagick_version|" ./Containers/nextcloud/start.sh
```
szaimen, line 60:
```suggestion
        sed -i "s|pecl install imagick.*|pecl install imagick-$imagick_version  >/dev/null|" ./Containers/nextcloud/start.sh
```

php/src/Docker/DockerActionManager.php
```diff
@@ -332,6 +332,10 @@ public function CreateContainer(Container $container) : void {
                     $replacements[1] = $this->configurationManager->GetCollaboraSeccompPolicy();
                 } elseif ($out[1] === '%NEXTCLOUD_STARTUP_APPS%') {
                     $replacements[1] = $this->configurationManager->GetNextcloudStartupApps();
+                } elseif ($out[1] === '%NEXTCLOUD_ADDITIONAL_APKS%') {
+                    $replacements[1] = $this->configurationManager->GetNextcloudAdditionalApks();
+                } elseif ($out[1] === '%NEXTCLOUD_ADDITIONAL_PHP_EXTENSIONS%') {
```
szaimen, line 337:
```suggestion
                } elseif ($out[1] === 'NEXTCLOUD_ADDITIONAL_APKS') {
                    $replacements[1] = $this->configurationManager->GetNextcloudAdditionalApks();
                } elseif ($out[1] === 'NEXTCLOUD_ADDITIONAL_PHP_EXTENSIONS') {
```

readme.md
```diff
@@ -467,6 +467,16 @@ You can run AIO also with docker rootless. How to do this is documented here: [d
...
 
+### How to add packets permanently to the Nextcloud container?
+Some Nextcloud apps require additional external dependencies that must be bundled within Nextcloud container in order to work correctly. As we cannot put each and every dependency for all apps into the container - as this would make the project very fast unmaintainable - there is an official way how you can add additional dependencies into the Nextcloud container. However note that doing this is not recommended since we do not test Nextcloud apps that require external dependencies. 
+
+You can do so by adding `-e NEXTCLOUD_ADDITIONAL_APKS=dependency1 dependency2` to the docker run command of the mastercontainer and customize the value to your fitting. It must be a string with small letters a-z, spaces and hyphens or '_'. You can find available packages here: https://pkgs.alpinelinux.org/packages?name=&branch=v3.16&repo=&arch=&maintainer=
```
szaimen, line 473:
```suggestion
You can do so by adding `-e NEXTCLOUD_ADDITIONAL_APKS="dependency1 dependency2"` to the docker run command of the mastercontainer and customize the value to your fitting. It must be a string with small letters a-z, spaces and hyphens or '_'. You can find available packages here: https://pkgs.alpinelinux.org/packages?name=&branch=v3.16&repo=&arch=&maintainer=
```

Containers/nextcloud/start.sh
szaimen: okay, just tested and it is not possible to check this beforehand based on a folder.

szaimen: found a solution

Zoey2936 (approved):
Looks good

szaimen:
Thanks Zoey! :)

Zoey2936:
I think numbers should also be allowed
grafik

szaimen:
good catch. Let me add this...

szaimen:
to review: https://github.com/nextcloud/all-in-one/pull/1387

PR #1381, opened 2022-11-09
rename TRUSTED_CACERTS_DIR to NEXTCLOUD_TRUSTED_CACERTS_DIR

szaimen:


szaimen:
@lollo0296 This is in order to always have the pattern `CONTAINER_VARIABLE_NAME`

Since we are not renaming the internally used name https://github.com/nextcloud/all-in-one/pull/1381/files#diff-7da19c38d7098fcaf5399e1f3ee7c1d2757065e2238b8b814a81b15ef79a6d77R552, it should be backwards compatible.

lollo0296:
@szaimen I got you. I will update the variable name in my `docker-compose.yml`, as soon as the next Release comes out.

Do you think backward compatibility is needed? This is in such an early stage, I doubt someone other then me is already using it in production. But at the end, I cannot say it for sure.

szaimen:
> Do you think backward compatibility is needed? This is in such an early stage, I doubt someone other then me is already using it in production. But at the end, I cannot say it for sure.

I'll rather make sure that it works for everyone :)

manual-install/update-yaml.sh
```diff
@@ -65,7 +65,7 @@ sed -i 's|NEXTCLOUD_MOUNT=|NEXTCLOUD_MOUNT=/mnt/          # This allows the Next
 sed -i 's|NEXTCLOUD_UPLOAD_LIMIT=|NEXTCLOUD_UPLOAD_LIMIT=10G          # This allows to change the upload limit of the Nextcloud container|' sample.conf
 sed -i 's|APACHE_MAX_SIZE=|APACHE_MAX_SIZE=10737418240          # This needs to be an integer and in sync with NEXTCLOUD_UPLOAD_LIMIT|' sample.conf
 sed -i 's|NEXTCLOUD_MAX_TIME=|NEXTCLOUD_MAX_TIME=3600          # This allows to change the upload time limit of the Nextcloud container|' sample.conf
-sed -i 's|TRUSTED_CACERTS_DIR=|TRUSTED_CACERTS_DIR=/path/to/my/cacerts          # Nextcloud container will trust all the Certification Authorities, whose certificates are included in the given directory.|' sample.conf
+sed -i 's|NEXTCLOUD_TRUSTED_CACERTS_DIR=|NEXTCLOUD_TRUSTED_CACERTS_DIR=/path/to/my/cacerts          # Nextcloud container will trust all the Certification Authorities, whose certificates are included in the given directory.|' sample.conf
```
szaimen, line 68: should use an existing path as example...

szaimen: done

szaimen:
This is now released with v3.0.0 Beta. Testing and feedback is welcome! See https://github.com/nextcloud/all-in-one#how-to-switch-the-channel

PR #1451, opened 2022-11-28
Fix reverse proxy documentation for nginx

Zoey2936:

### DIFF
--- a/reverse-proxy.md
+++ b/reverse-proxy.md
@@ -243,18 +243,19 @@ map $http_upgrade $connection_upgrade {
 
 server {
     listen 80;
-#    listen [::]:80;            # uncomment to use IPv6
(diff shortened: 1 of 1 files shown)


szaimen (approved):
Thanks, great fix! :)

SpriteOvO:
Hmmm, kinda strange. The `resolver localhost;` doesn't work for me, I still have to change `proxy_pass http://localhost:11000$request_uri;` to `proxy_pass http://127.0.0.1:11000$request_uri;` to fix the 502 Bad Gateway whatever the `resolver localhost;` is commented out or not.

Zoey2936:
That's because you have no resolver running on localhost, you could try 94.140.14.14 instead of localhost as resolver (adguard dns servers) or 1.1.1.1 for cloudflare or 8.8.8.8 for google

PR #1471, opened 2022-12-01
check if overlay2 is used

szaimen:
Hopefully make debugging of something like https://github.com/nextcloud/all-in-one/discussions/1467 in the future easier


iGadget:
Looking at the [Docker docs](https://docs.docker.com/storage/storagedriver/select-storage-driver/) it seems that besides overlay2, also [zfs](https://docs.docker.com/storage/storagedriver/zfs-driver/) and [btrfs](https://docs.docker.com/storage/storagedriver/btrfs-driver/) are 'legit' options when set up properly. Especially since overlay2 only supports [xfs (with ftype=1) and ext4](https://docs.docker.com/storage/storagedriver/select-storage-driver/#supported-backing-filesystems) as backing filesystems.
IMHO this check should really be to avoid accidental use of vfs, which is only meant for testing purposes in situations where any of the other Docker storage driver options are unavailable, or deprecated storage drivers.

szaimen:
Thanks for the review! Please re-review!

iGadget (commented):
Much better. As per [this thread](https://forum.proxmox.com/threads/docker-lxc-unprivileged-container-on-proxmox-7-with-zfs.99796/page-3), I wonder if fuse-overlayfs should be dis-recommended (small typo there in the warning) as well?
On the other hand, perhaps it's better/easier to put a section about Docker storage drivers (especially in combination with nested containers / LXC on top of i.e. ZFS) in the docs? This could also be linked the solution I found.

szaimen:
> As per [this thread](https://forum.proxmox.com/threads/docker-lxc-unprivileged-container-on-proxmox-7-with-zfs.99796/page-3), I wonder if fuse-overlayfs should be dis-recommended (small typo there in the warning) as well?

Fine by me and adjusted. Please check again!

st3iny (approved):

Zoey2936 (approved):

Issue #1032, opened 2022-08-18
with NC25 it will be possible to add fonts to Nextcloud Office. Add a volume for this


szaimen:
implemented by https://github.com/nextcloud/richdocuments/pull/2032,

a volume should be added to /opt/cool/systemplate/tmpfonts inside the collabora container

karlemilnikka:
Do you, @szaimen, know if there are any similar plans regarding support for custom dictionaries? Currently, it seems like users cannot add new words to a dictionary of their own in Nextcloud Office.

szaimen:
> Do you, @szaimen, know if there are any similar plans regarding support for custom dictionaries? Currently, it seems like users cannot add new words to a dictionary of their own in Nextcloud Office.

No idea, sry!

szaimen:
> This will work with OnlyOffice too?

No

szaimen:
This is now released with v4.0.0 Beta. Testing and feedback is welcome! See https://github.com/nextcloud/all-in-one#how-to-switch-the-channel

PR #1498, opened 2022-12-09
add a volume for collabora fonts

szaimen:
Close https://github.com/nextcloud/all-in-one/issues/1032

- [x] TODO: maybe we should use a different name in case other locations need inside the container need to be added later on?


Zoey2936 (approved):

szaimen:
Thanks for all the reviews @Zoey2936 ! :)

PR #1499, opened 2022-12-09
make collabora more safe

szaimen:
- [x] TODO: change it to allow private ip-ranges only which should make it work. Wait for Julius' response on this.

### DIFF
--- a/Containers/nextcloud/entrypoint.sh
+++ b/Containers/nextcloud/entrypoint.sh
@@ -400,6 +400,44 @@ if [ "$COLLABORA_ENABLED" = 'yes' ]; then
     php /var/www/html/occ config:app:set richdocuments wopi_url --value="https://$NC_DOMAIN/"
     # Fix https://github.com/nextcloud/all-in-one/issues/188:
     php /var/www/html/occ config:system:set allow_local_remote_servers --type=bool --value=true
+    # Make collabora more save
+    COLLABORA_IPv4_ADDRESS="$(echo "<?php echo gethostbyname('$NC_DOMAIN');" | php | head -1)"
+    COLLABORA_IPv6_Address="<?php \$record = dns_get_record('$NC_DOMAIN', DNS_AAAA);"
+    # shellcheck disable=SC2016
+    COLLABORA_IPv6_Address+='if (!empty($record)) {echo $record[0]["ipv6"];}'
+    COLLABORA_IPv6_Address="$(echo "$COLLABORA_IPv6_Address" | php | head -1)"
+    COLLABORA_ALLOW_LIST="$(php /var/www/html/occ config:app:get richdocuments wopi_allowlist)"
+    if [ -n "$COLLABORA_IPv4_ADDRESS" ]; then
+        if ! echo "$COLLABORA_ALLOW_LIST" | grep -q "$COLLABORA_IPv4_ADDRESS"; then
+            if [ -z "$COLLABORA_ALLOW_LIST" ]; then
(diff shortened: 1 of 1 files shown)


szaimen:
Looks like this makes it work, however I am really unsure if that will work everywhere. 
image

Better would be from my perspective to just allow all private ip-ranges:
`127.0.0.1/8,192.168.0.0/16,172.16.0.0/12,10.0.0.0/8`

Containers/nextcloud/entrypoint.sh
```diff
@@ -400,6 +400,38 @@ if [ "$COLLABORA_ENABLED" = 'yes' ]; then
     php /var/www/html/occ config:app:set richdocuments wopi_url --value="https://$NC_DOMAIN/"
     # Fix https://github.com/nextcloud/all-in-one/issues/188:
     php /var/www/html/occ config:system:set allow_local_remote_servers --type=bool --value=true
+    # Make collabora more save
+    COLLABORA_IPv4_ADDRESS="$(echo "<?php echo gethostbyname('$COLLABORA_HOST');" | php | head -1)"
+    # shellcheck disable=SC2034
+    COLLABORA_IPv6_Address="$(echo "<?php \$record = dns_get_record('$COLLABORA_HOST', DNS_AAAA); if (\!empty(\$record)) {echo \$record[0]['ipv6'];}" | sed 's|\\!empty|!empty|' | php | head -1)"
```
szaimen, line 406:
```suggestion
    COLLABORA_IPv4_ADDRESS="$(echo "<?php echo gethostbyname('$NC_DOMAIN');" | php | head -1)"
    # shellcheck disable=SC2034
    COLLABORA_IPv6_Address="$(echo "<?php \$record = dns_get_record('$NC_DOMAIN', DNS_AAAA); if (\!empty(\$record)) {echo \$record[0]['ipv6'];}" | sed 's|\\!empty|!empty|' | php | head -1)"
```

Containers/nextcloud/entrypoint.sh
```diff
@@ -400,6 +400,38 @@ if [ "$COLLABORA_ENABLED" = 'yes' ]; then
...
+                COLLABORA_ALLOW_LIST="$COLLABORA_IPv6_ADDRESS"
+            else
+                COLLABORA_ALLOW_LIST+=",$COLLABORA_IPv6_ADDRESS"
+            fi
+        fi
+    else
+        echo "No ipv6-address found for the collabora container."
+    fi
+    if [ -n "$COLLABORA_ALLOW_LIST" ]; then
+        php /var/www/html/occ config:app:set richdocuments wopi_allowlist --value="$COLLABORA_ALLOW_LIST"
```
szaimen, line 431: should also insert `127.0.0.1/8,192.168.0.0/16,172.16.0.0/12,10.0.0.0/8,fd00::/8,::1` here.

szaimen: done

juliushaertl:
It should only require the collabora internal ip address here (assuming that the reverse proxy is properly accepted as a trusted proxy and forwarding IP in the X-Forwarded-For header). Any reason why you would prefer the full private ranges?

juliushaertl (approved):
Looks good, so not blocking, but ideally only the relevant ip ranges are added to the allow list.

szaimen:
> Looks good, so not blocking, but ideally only the relevant ip ranges are added to the allow list.

Thanks Julius! However since I am really not sure if the needed ip-address will always be the gateway ip of the docker bridge network, I am rather making sure that it will work always and have a bit less secure config. For more details check my comments on Talk on c.nc.c :)

PR #1539, opened 2022-12-17
reload in case of undefined response

szaimen:

### DIFF
--- a/php/public/forms.js
+++ b/php/public/forms.js
@@ -19,12 +19,13 @@
     const xhr = e.target;
     if (xhr.status === 201) {
       window.location.replace(xhr.getResponseHeader('Location'));
-    }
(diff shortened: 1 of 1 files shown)


php/public/forms.js
```diff
@@ -19,12 +19,13 @@
...
-    if (xhr.status === 422) {
+    } else if (xhr.status === 422) {
       showError(xhr.response);
-    }
-    if (xhr.status === 500) {
-      showError("Server error. Please see the logs for details.");
+    } else if (xhr.status === 500) {
+      showError("Server error. Please check the mastercontainer logs for details.");
+    } else {
+      // If the responose is not defined, we should still reload.
```
szaimen, line 27:
```suggestion
      // If the responose is not one of the above, we should still reload.
```

PR #1569, opened 2022-12-21
add gcompat for the recognize app

szaimen:
It is needed for the recognize app IIRC.

### DIFF
--- a/Containers/nextcloud/Dockerfile
+++ b/Containers/nextcloud/Dockerfile
@@ -199,6 +199,7 @@ RUN set -ex; \
         sudo \
         grep \
         coreutils \
+        gcompat \
     ; \
     rm -rf /var/lib/apt/lists/*
 


szaimen:
@marcelklehr I read that gcompat is required for the recognize app to work. Is this correct? :)

thomasverelst:
Hi, this is based on my gist https://gist.github.com/thomasverelst/60403f22e7a9f5200b0dc43615c54720
where I included it. 
I'll repeat here: 
"I'm running my setup on ARM64 architecture for now (oracle free tier), so maybe that's why I needed gcompat. Recognize seems to include a nodejs binary, but it doesnt run in the docker (tried bash'ing into the docker and executing it), and gcompat was required. I also tried to include nodejs and npm is apks in the master docker command but - even when pointing the path in Recognize to the right nodejs, it failed to execute. Maybe gcompat is not needed for x86-64, I did not test."

marcelklehr:
> I read that gcompat is required for the recognize app to work.

Huh, I never read that anywhere. I've been running it fine inside docker on an x86_64 machine (until the setup destroyed itself, but that's another story...).

marcelklehr (approved):

szaimen:
This is now released with v4.0.1 Beta. Testing and feedback is welcome! See https://github.com/nextcloud/all-in-one#how-to-switch-the-channel

erwanlpfr:
Hello, 

With the actual beta tag from docker hub, I still have the issue mentioned here : https://github.com/nextcloud/recognize/issues/604

Do you have any idea how to debug it? 
Since I can't run `npm ci`.

Thanks !

marcelklehr:
@erwanlpfr Hello! Can you try running `.../recognize/bin/node .../recognize/src/test_libtensorflow.js` where `.../recognize` is the path to recognize on your server?

erwanlpfr:
Hmm, I waited one day and it seems to work.
Maybe it was slow for some reasons.

Thanks for your reply!

marcelklehr:
Great! I'm happy it works now :)

marcelklehr:
In that case I'm out of ideas.

szaimen:
@fabianbees thanks for testing! Can you check if adding `libc6-compat` additionally makes it work?

szaimen:
No, moving back to debian is not an option for AIO. So we need to find a way to make it work with alpine linux.

szaimen:
What I wonder: wasnt recognize already running in wasm mode in the past? Why is it not running at all anymore? Should we remove gcompat again?

szaimen:
Wow, is this the final answer? https://github.com/tensorflow/tfjs/issues/1425 I thought recognize could was able to work in wasm mode at some point?

marcelklehr:
> wasnt recognize already running in wasm mode on alpine linux in the past? Why is it not running at all anymore? Should we remove gcompat again?

I believe it still works in WASM mode. WIth gcompat the check whether libtensorflow can be loaded just never ends. WIthout gcompat it simply fails.

szaimen:
I see. Then removing gcompat again probably makes sense since it at least finishes the test then and allows to enable the wasm mode.

szaimen:
Will be done in https://github.com/nextcloud/all-in-one/pull/1816

szaimen:
This is now released with v4.3.0 Beta. Testing and feedback is welcome! See https://github.com/nextcloud/all-in-one#how-to-switch-the-channel

thomasverelst:
I think the beta update (Nextcloud AIO v4.3.1) broke my recognize (reinstalled v3.4.0) (WASM enabled)

`Could not execute the Node.js binary. You may need to set the path to a working binary manually.`

In logs:
```
Failed to install Tensorflow.js: sh: /var/www/html/custom_apps/recognize/bin/node: not found
Exception: Failed to install Tensorflow.js: sh: /var/www/html/custom_apps/recognize/bin/node: not found

```

It used to work before (in WASM mode on ARM).
So it seems that some setups require gcompat (luckily its easy to add manually)

szaimen:
There seems to be no way to make it work for everyone so we removed gcompat again (as it breaks recognize on other instances). But if you need it you can add it back as you did 👍

szaimen:
Looks like this should fix the requirement on gcompat for arm64 in a future update: https://github.com/nextcloud/recognize/commit/30602dcfa972257f4fae1534fb2419ef319acdb9

PR #1573, opened 2022-12-22
migrate apache container to alpine

Zoey2936:
- [x] TODO: check if the www-data user has the same id
- [x] needs #1619

### DIFF
--- a/Containers/apache/Dockerfile
+++ b/Containers/apache/Dockerfile
@@ -30,26 +30,25 @@ COPY --from=caddy /usr/bin/caddy /usr/bin/
 RUN chmod +x /usr/bin/caddy
 
 RUN sed -i \
-		-e 's/^#\(LoadModule .*mod_rewrite.so\)/\1/' \
-		-e 's/^#\(LoadModule .*mod_headers.so\)/\1/' \
-		-e 's/^#\(LoadModule .*mod_proxy.so\)/\1/' \
-		-e 's/^#\(LoadModule .*mod_proxy_fcgi.so\)/\1/' \
-		-e 's/^#\(LoadModule .*mod_setenvif.so\)/\1/' \
-		-e 's/^#\(LoadModule .*mod_env.so\)/\1/' \
-		-e 's/^#\(LoadModule .*mod_mime.so\)/\1/' \
-		-e 's/^#\(LoadModule .*mod_dir.so\)/\1/' \
-		-e 's/^#\(LoadModule .*mod_authz_core.so\)/\1/' \
-		-e 's/^#\(LoadModule .*mod_alias.so\)/\1/' \
-		-e 's/^#\(LoadModule .*mod_mpm_event.so\)/\1/' \
-		-e 's/\(LoadModule .*mod_mpm_worker.so\)/#\1/' \
-		-e 's/\(LoadModule .*mod_mpm_prefork.so\)/#\1/' \
-		conf/httpd.conf
-
+                -e '/^Listen /d' \
+                -e 's/^#\(LoadModule .*mod_rewrite.so\)/\1/' \
+                -e 's/^#\(LoadModule .*mod_headers.so\)/\1/' \
+                -e 's/^#\(LoadModule .*mod_proxy.so\)/\1/' \
+                -e 's/^#\(LoadModule .*mod_proxy_fcgi.so\)/\1/' \
+                -e 's/^#\(LoadModule .*mod_setenvif.so\)/\1/' \
+                -e 's/^#\(LoadModule .*mod_env.so\)/\1/' \
+                -e 's/^#\(LoadModule .*mod_mime.so\)/\1/' \
+                -e 's/^#\(LoadModule .*mod_dir.so\)/\1/' \
+                -e 's/^#\(LoadModule .*mod_authz_core.so\)/\1/' \
+                -e 's/^#\(LoadModule .*mod_alias.so\)/\1/' \
+                -e 's/^#\(LoadModule .*mod_mpm_event.so\)/\1/' \
+                -e 's/\(LoadModule .*mod_mpm_worker.so\)/#\1/' \
+                -e 's/\(LoadModule .*mod_mpm_prefork.so\)/#\1/' \
+                conf/httpd.conf; \
+		echo "Include conf/nextcloud.conf" | tee -a conf/httpd.conf; \
+                echo "ServerName localhost" | tee -a conf/httpd.conf
+		
 COPY nextcloud.conf conf
-RUN echo Include conf/nextcloud.conf | tee -a conf/httpd.conf
-
-RUN sed -i "/^Listen /d" conf/httpd.conf
-echo "ServerName localhost" | tee -a conf/httpd.conf
 
 RUN set -ex; \
     rm -rf conf/original conf/original && \
--- a/Containers/apache/Dockerfile
+++ b/Containers/apache/Dockerfile
@@ -1,59 +1,61 @@
 # Caddy is a requirement
 FROM caddy:2.6.2-alpine as caddy
 
-FROM debian:bullseye-20230109-slim
+FROM httpd:2.4.54-alpine3.17
 
 RUN mkdir -p /mnt/data; \
     chown www-data:www-data /mnt/data;
 
 VOLUME /mnt/data
 
+RUN apk add --no-cache bash \
+		       wget \
+		       tzdata \
+		       coreutils \
+		       supervisor \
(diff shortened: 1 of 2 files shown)


szaimen:
Wow, thanks a lot Zoey! I will include this in the release after 4.0.1 :)

Containers/apache/Dockerfile
```diff
@@ -1,60 +1,40 @@
 # Caddy is a requirement
 FROM caddy:2.6.2-alpine as caddy
 
-FROM debian:bullseye-20221219-slim
+FROM httpd:2.4.54-alpine3.17
```
szaimen, line 4: the image needs to be change to the php-fpm-alpine

Zoey2936: why?

szaimen: ah sorry, wrong PR 😅

Containers/apache/Dockerfile
```diff
@@ -1,59 +1,44 @@
...
```
szaimen, line 4: lets wait for alpine 3.17.1 as well

Containers/apache/Dockerfile
```diff
@@ -1,59 +1,44 @@
...
+		-e 's/^#\(LoadModule .*mod_proxy_fcgi.so\)/\1/' \
+		-e 's/^#\(LoadModule .*mod_setenvif.so\)/\1/' \
+		-e 's/^#\(LoadModule .*mod_env.so\)/\1/' \
+		-e 's/^#\(LoadModule .*mod_mime.so\)/\1/' \
+		-e 's/^#\(LoadModule .*mod_dir.so\)/\1/' \
+		-e 's/^#\(LoadModule .*mod_authz_core.so\)/\1/' \
+		-e 's/^#\(LoadModule .*mod_alias.so\)/\1/' \
+		-e 's/^#\(LoadModule .*mod_mpm_event.so\)/\1/' \
+		-e 's/\(LoadModule .*mod_mpm_worker.so\)/#\1/' \
+		-e 's/\(LoadModule .*mod_mpm_prefork.so\)/#\1/' \
```
szaimen, line 29: I think prefork is not supported. See https://github.com/nextcloud/spreed/pull/8509/files

Zoey2936: Thats why we disable mpm_worker and mpm_prefork and enable mpm_event

szaimen: Ah I see. Sorry, didnt read correctly 😅

Containers/apache/Dockerfile
Zoey2936: the httpd image was now pushed to alpine v3.17.1

szaimen: It looks like I will not get to this due to time problems before the next beta release which is planned for coming tuesday. Sorry for that! I'll try to include it in the version afterwards.

szaimen:
I wonder: would it make sense to enable http2 in Apache? If I understand correctly is caddy enabling http2 and http3 out of the box for outgoing connections so our internal apache does not need it? Or WDYT?

Zoey2936:
> I wonder: would it make sense to enable http2 in Apache? If I understand correctly is caddy enabling http2 and http3 out of the box for outgoing connections so our internal apache does not need it? Or WDYT?

It would make the connection between caddy and apache uses http2

We would need to configure it inside the caddy file: https://caddyserver.com/docs/caddyfile/directives/reverse_proxy#upstream-addresses (`reverse_proxy h2c://localhost:8000 {` instead of `reverse_proxy localhost:8000 {`)

But:
> versions allows customizing which versions of HTTP to support. As a special case, "h2c" is a valid value which will enable cleartext HTTP/2 connections to the upstream (however, this is a non-standard feature that does not use Go's default HTTP transport, so it is exclusive of other features; subject to change or removal). Default: 1.1 2, or if scheme is h2c://, h2c 2

we can only use h2c and not h2 since caddy connects to apache unencrypted if I understand correctly

[Also nginx hasn't implemented this because they don't see any sense in this.](https://mailman.nginx.org/pipermail/nginx/2015-December/049445.html)

szaimen:
Ah I see but would this bring any performance advantage?

Zoey2936:
> Ah I see but would this bring any performance advantage?

If I read the nginx mail, it could even degrade the performance

szaimen:
> > Ah I see but would this bring any performance advantage?
> 
> If I read the nginx mail, it could even degrade the performance (maybe read it yourself)

I see. Okay good, then I'd say lets leave it as it is without http2 enabled in apache :)

szaimen:
I'd like to keep the userid and maybe also the user and group-name. Can we add something like this here as well? https://github.com/nextcloud/all-in-one/blob/b249e02d53f063e1b75df96f50007a329d94717e/Containers/nextcloud/Dockerfile#L4-L11

Containers/apache/Dockerfile
```diff
@@ -1,59 +1,44 @@
...
-        apache2 \
-        supervisor \
-        wget \
-        ca-certificates \
-        openssl \
-        netcat \
-        dpkg-dev \
-    ; \
-    rm -rf /var/lib/apt/lists/*
+RUN apk add --no-cache bash tzdata coreutils supervisor wget ca-certificates openssl netcat-openbsd
```
szaimen, line 11: can we keep the former formatting?

Zoey2936: done

Zoey2936:
> I'd like to keep the userid and maybe also the user and group-name. Can we add something like this here as well?
> 
> https://github.com/nextcloud/all-in-one/blob/b249e02d53f063e1b75df96f50007a329d94717e/Containers/nextcloud/Dockerfile#L4-L11

we just need the www-data user/group with id 33?

Containers/apache/Dockerfile
```diff
@@ -1,59 +1,60 @@
...
+		-e 's/^#\(LoadModule .*mod_mpm_event.so\)/\1/' \
+		-e 's/\(LoadModule .*mod_mpm_worker.so\)/#\1/' \
+		-e 's/\(LoadModule .*mod_mpm_prefork.so\)/#\1/' \
+		conf/httpd.conf
+
+COPY nextcloud.conf conf
+RUN echo Include conf/nextcloud.conf | tee -a conf/httpd.conf
+
+RUN sed -i "/^Listen /d" conf/httpd.conf
+echo "ServerName localhost" | tee -a conf/httpd.conf
```
szaimen, line 52: same as https://github.com/nextcloud/all-in-one/pull/1577/files#r1085253864 and below

Zoey2936: done

Zoey2936:
@szaimen should be finished now

Containers/apache/start.sh
```diff
@@ -48,7 +48,7 @@ echo "$CADDYFILE" > /Caddyfile
 # Add caddy path
 mkdir -p /mnt/data/caddy/
 
-# Fix apache sturtup
-rm -f /var/run/apache2/apache2.pid
+# Fix apache startup
+rm -f logs/httpd.pid
```
szaimen, line 52: Can we use an absloute path here?

Zoey2936: only there or also on all other locations?

szaimen: for now only here I'd say

Zoey2936:
```suggestion
rm -f /usr/local/apache2/logs/httpd.pid
```

Containers/apache/start.sh
```diff
@@ -49,6 +49,6 @@ echo "$CADDYFILE" > /Caddyfile
 mkdir -p /mnt/data/caddy/
 
 # Fix apache startup
-rm -f logs/httpd.pid
+rm -f /usr/local/apache2/logs/httpd.pid
```
szaimen, line 52: btw, why is the pid located in logs?

Zoey2936: good question:

```console
root@mx:~# docker run --rm -it httpd:2.4.54-alpine3.17 httpd -V | grep pid
 -D DEFAULT_PIDLOG="logs/httpd.pid"
```

szaimen:
LGTM now but will try to test this locally first 👍

szaimen (approved):
Builds correctly. Testing will show if everything works. Thank you very much! :)

szaimen:
It works 🎉🎉🎉🎉

Zoey2936:
> It works 🎉🎉🎉🎉

Perfect!

PR #1574, opened 2022-12-22
Create Helm-chart

szaimen:
Close https://github.com/nextcloud/all-in-one/discussions/1582

### TODO
- [x] fix SC
- [x] fix review comments
- [x]  ~~should this maybe get pushed to a new repo or is it still possible to import it even if it is not located in the root dir of the repo?~~ should be fine based on Richard

---

### follow-ups: (tracked in https://github.com/nextcloud/all-in-one/discussions/1582)
- [ ]  test if it really works
- [ ] implement releases https://medium.com/@stefanprodan/automate-helm-chart-repository-publishing-with-github-actions-and-pages-8a374ce24cf4 and https://github.com/nextcloud/helm/blob/master/.github/workflows/release.yaml
- [ ] think about different default sizes for different volumes...

### DIFF
--- a/.github/workflows/update-helm.yml
+++ b/.github/workflows/update-helm.yml
@@ -2,8 +2,8 @@ name: Update Helm Chart
 
 on:
   workflow_dispatch:
-  # schedule:
-  # - cron:  '00 12 * * *'
+  schedule:
+  - cron:  '00 12 * * *'
 
 jobs:
   psalm:
@@ -14,7 +14,12 @@ jobs:
         uses: actions/checkout@v3
       - name: update helm chart
         run: |
-          sudo bash helm-chart/update-helm.sh
+          DOCKER_TAG="$(curl -L -s 'https://registry.hub.docker.com/v2/repositories/nextcloud/all-in-one/tags?page_size=1024' | jq '."results"[]["name"]' | sed 's|"||g' | grep '^20' | sort -r | head -1)"
+          DOCKER_TAG="${DOCKER_TAG%%latest*}"
+          export DOCKER_TAG
+          if [ -n "$DOCKER_TAG" ] && ! grep -q "$DOCKER_TAG" ./helm-chart/templates/nextcloud-aio-nextcloud-deployment.yaml; then
+            sudo bash helm-chart/update-helm.sh "$DOCKER_TAG"
+          fi
       - name: Create Pull Request
         uses: peter-evans/create-pull-request@v4
         with:
@@ -25,4 +30,4 @@ jobs:
           labels: dependencies
           milestone: next
           branch: aio-helm-update
-          github_token: ${{ secrets.GITHUB_TOKEN }}
+          token: ${{ secrets.GITHUB_TOKEN }}
--- a/helm-chart/Chart.yaml
+++ b/helm-chart/Chart.yaml
@@ -1,0 +1,8 @@
+name: Nextcloud AIO Helm Chart
+description: A generated Helm Chart for Nextcloud AIO from Skippbox Kompose
+version: 4.1.0
+apiVersion: v1
+keywords:
+  - latest
+sources: https://github.com/nextcloud/all-in-one/tree/main/helm-chart
+home: https://github.com/nextcloud/all-in-one/tree/main/helm-chart
--- a/helm-chart/readme.md
+++ b/helm-chart/readme.md
@@ -1,0 +1,3 @@
+# You can also install the AIO containers on Kubernetes using this Helm Chart
+
+This is currently beta and not ready yet.
--- a/helm-chart/templates/nextcloud-aio-apache-deployment.yaml
+++ b/helm-chart/templates/nextcloud-aio-apache-deployment.yaml
@@ -1,0 +1,61 @@
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  annotations:
+    kompose.cmd: kompose convert -c -f latest.yml
+    kompose.version: 1.27.0 (b0ed6a2c9)
+  labels:
+    io.kompose.service: nextcloud-aio-apache
+  name: nextcloud-aio-apache
+spec:
+  replicas: 1
+  selector:
+    matchLabels:
+      io.kompose.service: nextcloud-aio-apache
+  template:
+    metadata:
+      annotations:
+        kompose.cmd: kompose convert -c -f latest.yml
+        kompose.version: 1.27.0 (b0ed6a2c9)
+      labels:
+        io.kompose.network/nextcloud-aio: "true"
+        io.kompose.service: nextcloud-aio-apache
+    spec:
+      containers:
+        - env:
+            - name: APACHE_MAX_SIZE
+              value: "{{ .Values.APACHE_MAX_SIZE }}"
+            - name: APACHE_MAX_TIME
+              value: "{{ .Values.NEXTCLOUD_MAX_TIME }}"
+            - name: APACHE_PORT
+              value: "{{ .Values.APACHE_PORT }}"
+            - name: COLLABORA_HOST
+              value: nextcloud-aio-collabora
+            - name: NC_DOMAIN
+              value: "{{ .Values.NC_DOMAIN }}"
+            - name: NEXTCLOUD_HOST
+              value: nextcloud-aio-nextcloud
+            - name: ONLYOFFICE_HOST
+              value: nextcloud-aio-onlyoffice
+            - name: TALK_HOST
+              value: nextcloud-aio-talk
+            - name: TZ
+              value: "{{ .Values.TIMEZONE }}"
+          image: nextcloud/aio-apache:20221229_091124-{{ .Values.IMAGE_TAG }}
+          name: nextcloud-aio-apache
+          ports:
+            - containerPort: {{ .Values.APACHE_PORT }}
+          volumeMounts:
+            - mountPath: /var/www/html
+              name: nextcloud-aio-nextcloud
+              readOnly: true
+            - mountPath: /mnt/data
+              name: nextcloud-aio-apache
+      volumes:
+        - name: nextcloud-aio-nextcloud
+          persistentVolumeClaim:
+            claimName: nextcloud-aio-nextcloud
+            readOnly: true
+        - name: nextcloud-aio-apache
+          persistentVolumeClaim:
+            claimName: nextcloud-aio-apache
--- a/helm-chart/templates/nextcloud-aio-apache-persistentvolumeclaim.yaml
+++ b/helm-chart/templates/nextcloud-aio-apache-persistentvolumeclaim.yaml
@@ -1,0 +1,15 @@
+apiVersion: v1
+kind: PersistentVolumeClaim
+metadata:
+  labels:
+    io.kompose.service: nextcloud-aio-apache
+  name: nextcloud-aio-apache
+spec:
+  {{- if .Values.STORAGE_CLASS }}
+  storageClassName: {{ .Values.STORAGE_CLASS }}
+  {{- end }}
+  accessModes:
+    - ReadWriteMany
+  resources:
+    requests:
+      storage: {{ .Values.MAX_STORAGE_SIZE }}
--- a/helm-chart/templates/nextcloud-aio-apache-service.yaml
+++ b/helm-chart/templates/nextcloud-aio-apache-service.yaml
@@ -1,0 +1,18 @@
+apiVersion: v1
+kind: Service
+metadata:
+  annotations:
+    kompose.cmd: kompose convert -c -f latest.yml
+    kompose.version: 1.27.0 (b0ed6a2c9)
+  labels:
+    io.kompose.service: nextcloud-aio-apache
+  name: nextcloud-aio-apache
+spec:
+  ports:
+    - name: "{{ .Values.APACHE_PORT }}"
+      port: {{ .Values.APACHE_PORT }}
+      targetPort: {{ .Values.APACHE_PORT }}
+  selector:
+    io.kompose.service: nextcloud-aio-apache
+status:
+  loadBalancer: {}
--- a/helm-chart/templates/nextcloud-aio-clamav-deployment.yaml
+++ b/helm-chart/templates/nextcloud-aio-clamav-deployment.yaml
@@ -1,0 +1,40 @@
+{{- if eq .Values.CLAMAV_ENABLED "yes" }}
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  annotations:
+    kompose.cmd: kompose convert -c -f latest.yml
+    kompose.version: 1.27.0 (b0ed6a2c9)
+  labels:
+    io.kompose.service: nextcloud-aio-clamav
+  name: nextcloud-aio-clamav
+spec:
+  replicas: 1
+  selector:
+    matchLabels:
+      io.kompose.service: nextcloud-aio-clamav
+  template:
+    metadata:
+      annotations:
+        kompose.cmd: kompose convert -c -f latest.yml
+        kompose.version: 1.27.0 (b0ed6a2c9)
+      labels:
+        io.kompose.network/nextcloud-aio: "true"
+        io.kompose.service: nextcloud-aio-clamav
+    spec:
+      containers:
+        - env:
+            - name: CLAMD_STARTUP_TIMEOUT
+              value: "90"
+            - name: TZ
+              value: "{{ .Values.TIMEZONE }}"
+          image: nextcloud/aio-clamav:20221229_091124-{{ .Values.IMAGE_TAG }}
+          name: nextcloud-aio-clamav
+          volumeMounts:
+            - mountPath: /var/lib/clamav
+              name: nextcloud-aio-clamav
+      volumes:
+        - name: nextcloud-aio-clamav
+          persistentVolumeClaim:
+            claimName: nextcloud-aio-clamav
+{{- end }}
--- a/helm-chart/templates/nextcloud-aio-clamav-persistentvolumeclaim.yaml
+++ b/helm-chart/templates/nextcloud-aio-clamav-persistentvolumeclaim.yaml
@@ -1,0 +1,15 @@
+apiVersion: v1
+kind: PersistentVolumeClaim
+metadata:
+  labels:
+    io.kompose.service: nextcloud-aio-clamav
+  name: nextcloud-aio-clamav
+spec:
+  {{- if .Values.STORAGE_CLASS }}
+  storageClassName: {{ .Values.STORAGE_CLASS }}
+  {{- end }}
+  accessModes:
+    - ReadWriteMany
+  resources:
+    requests:
+      storage: {{ .Values.MAX_STORAGE_SIZE }}
--- a/helm-chart/templates/nextcloud-aio-collabora-deployment.yaml
+++ b/helm-chart/templates/nextcloud-aio-collabora-deployment.yaml
@@ -1,0 +1,44 @@
+{{- if eq .Values.COLLABORA_ENABLED "yes" }}
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  annotations:
+    kompose.cmd: kompose convert -c -f latest.yml
+    kompose.version: 1.27.0 (b0ed6a2c9)
+  labels:
+    io.kompose.service: nextcloud-aio-collabora
+  name: nextcloud-aio-collabora
+spec:
+  replicas: 1
+  selector:
+    matchLabels:
+      io.kompose.service: nextcloud-aio-collabora
+  template:
+    metadata:
+      annotations:
+        kompose.cmd: kompose convert -c -f latest.yml
+        kompose.version: 1.27.0 (b0ed6a2c9)
+      labels:
+        io.kompose.network/nextcloud-aio: "true"
+        io.kompose.service: nextcloud-aio-collabora
+    spec:
+      containers:
+        - env:
+            - name: TZ
+              value: "{{ .Values.TIMEZONE }}"
+            - name: aliasgroup1
+              value: https://{{ .Values.NC_DOMAIN }}:443
+            - name: dictionaries
+              value: "{{ .Values.COLLABORA_DICTIONARIES }}"
+            - name: extra_params
+              value: --o:ssl.enable=false --o:ssl.termination=true --o:logging.level=warning --o:home_mode.enable=true {{ .Values.COLLABORA_SECCOMP_POLICY }} --o:remote_font_config.url=https://{{ .Values.NC_DOMAIN }}/apps/richdocuments/settings/fonts.json
+          image: nextcloud/aio-collabora:20221229_091124-{{ .Values.IMAGE_TAG }}
+          name: nextcloud-aio-collabora
+          volumeMounts:
+            - mountPath: /opt/cool/systemplate/tmpfonts
+              name: nextcloud-aio-collabora-fonts
+      volumes:
+        - name: nextcloud-aio-collabora-fonts
+          persistentVolumeClaim:
+            claimName: nextcloud-aio-collabora-fonts
+{{- end }}
--- a/helm-chart/templates/nextcloud-aio-collabora-fonts-persistentvolumeclaim.yaml
+++ b/helm-chart/templates/nextcloud-aio-collabora-fonts-persistentvolumeclaim.yaml
@@ -1,0 +1,15 @@
+apiVersion: v1
+kind: PersistentVolumeClaim
+metadata:
+  labels:
+    io.kompose.service: nextcloud-aio-collabora-fonts
+  name: nextcloud-aio-collabora-fonts
+spec:
+  {{- if .Values.STORAGE_CLASS }}
+  storageClassName: {{ .Values.STORAGE_CLASS }}
+  {{- end }}
+  accessModes:
+    - ReadWriteMany
+  resources:
+    requests:
+      storage: {{ .Values.MAX_STORAGE_SIZE }}
--- a/helm-chart/templates/nextcloud-aio-database-deployment.yaml
+++ b/helm-chart/templates/nextcloud-aio-database-deployment.yaml
@@ -1,0 +1,50 @@
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  annotations:
+    kompose.cmd: kompose convert -c -f latest.yml
+    kompose.version: 1.27.0 (b0ed6a2c9)
+  labels:
+    io.kompose.service: nextcloud-aio-database
+  name: nextcloud-aio-database
+spec:
+  replicas: 1
+  selector:
+    matchLabels:
+      io.kompose.service: nextcloud-aio-database
+  template:
+    metadata:
+      annotations:
+        kompose.cmd: kompose convert -c -f latest.yml
+        kompose.version: 1.27.0 (b0ed6a2c9)
+      labels:
+        io.kompose.network/nextcloud-aio: "true"
+        io.kompose.service: nextcloud-aio-database
+    spec:
+      containers:
+        - env:
+            - name: PGTZ
+              value: "{{ .Values.TIMEZONE }}"
+            - name: POSTGRES_DB
+              value: nextcloud_database
+            - name: POSTGRES_PASSWORD
+              value: "{{ .Values.DATABASE_PASSWORD }}"
+            - name: POSTGRES_USER
+              value: nextcloud
+            - name: TZ
+              value: "{{ .Values.TIMEZONE }}"
+          image: nextcloud/aio-postgresql:20221229_091124-{{ .Values.IMAGE_TAG }}
+          name: nextcloud-aio-database
+          volumeMounts:
+            - mountPath: /var/lib/postgresql/data
+              name: nextcloud-aio-database
+            - mountPath: /mnt/data
+              name: nextcloud-aio-database-dump
+      terminationGracePeriodSeconds: 1800
+      volumes:
+        - name: nextcloud-aio-database
+          persistentVolumeClaim:
+            claimName: nextcloud-aio-database
+        - name: nextcloud-aio-database-dump
+          persistentVolumeClaim:
+            claimName: nextcloud-aio-database-dump
--- a/helm-chart/templates/nextcloud-aio-database-dump-persistentvolumeclaim.yaml
+++ b/helm-chart/templates/nextcloud-aio-database-dump-persistentvolumeclaim.yaml
@@ -1,0 +1,15 @@
+apiVersion: v1
+kind: PersistentVolumeClaim
+metadata:
+  labels:
+    io.kompose.service: nextcloud-aio-database-dump
+  name: nextcloud-aio-database-dump
+spec:
+  {{- if .Values.STORAGE_CLASS }}
+  storageClassName: {{ .Values.STORAGE_CLASS }}
+  {{- end }}
+  accessModes:
+    - ReadWriteMany
+  resources:
+    requests:
+      storage: {{ .Values.MAX_STORAGE_SIZE }}
--- a/helm-chart/templates/nextcloud-aio-fulltextsearch-deployment.yaml
+++ b/helm-chart/templates/nextcloud-aio-fulltextsearch-deployment.yaml
@@ -1,0 +1,42 @@
+{{- if eq .Values.FULLTEXTSEARCH_ENABLED "yes" }}
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  annotations:
+    kompose.cmd: kompose convert -c -f latest.yml
+    kompose.version: 1.27.0 (b0ed6a2c9)
+  labels:
+    io.kompose.service: nextcloud-aio-fulltextsearch
+  name: nextcloud-aio-fulltextsearch
+spec:
+  replicas: 1
+  selector:
+    matchLabels:
+      io.kompose.service: nextcloud-aio-fulltextsearch
+  template:
+    metadata:
+      annotations:
+        kompose.cmd: kompose convert -c -f latest.yml
+        kompose.version: 1.27.0 (b0ed6a2c9)
+      labels:
+        io.kompose.network/nextcloud-aio: "true"
+        io.kompose.service: nextcloud-aio-fulltextsearch
+    spec:
+      containers:
+        - env:
+            - name: ES_JAVA_OPTS
+              value: -Xms1024M -Xmx1024M
+            - name: TZ
+              value: "{{ .Values.TIMEZONE }}"
+            - name: discovery.type
+              value: single-node
+          image: nextcloud/aio-fulltextsearch:20221229_091124-{{ .Values.IMAGE_TAG }}
+          name: nextcloud-aio-fulltextsearch
+          volumeMounts:
+            - mountPath: /usr/share/elasticsearch/data
+              name: nextcloud-aio-elasticsearch
+      volumes:
+        - name: nextcloud-aio-elasticsearch
+          persistentVolumeClaim:
+            claimName: nextcloud-aio-elasticsearch
+{{- end }}
--- a/helm-chart/templates/nextcloud-aio-networkpolicy.yaml
+++ b/helm-chart/templates/nextcloud-aio-networkpolicy.yaml
@@ -1,0 +1,13 @@
+apiVersion: networking.k8s.io/v1
+kind: NetworkPolicy
+metadata:
+  name: nextcloud-aio
+spec:
+  ingress:
+    - from:
+        - podSelector:
+            matchLabels:
+              io.kompose.network/nextcloud-aio: "true"
+  podSelector:
+    matchLabels:
+      io.kompose.network/nextcloud-aio: "true"
--- a/helm-chart/templates/nextcloud-aio-nextcloud-data-persistentvolumeclaim.yaml
+++ b/helm-chart/templates/nextcloud-aio-nextcloud-data-persistentvolumeclaim.yaml
@@ -1,0 +1,15 @@
+apiVersion: v1
+kind: PersistentVolumeClaim
+metadata:
+  labels:
+    io.kompose.service: nextcloud-aio-nextcloud-data
+  name: nextcloud-aio-nextcloud-data
+spec:
+  {{- if .Values.STORAGE_CLASS }}
+  storageClassName: {{ .Values.STORAGE_CLASS }}
+  {{- end }}
+  accessModes:
+    - ReadWriteMany
+  resources:
+    requests:
+      storage: {{ .Values.MAX_STORAGE_SIZE }}
--- a/helm-chart/templates/nextcloud-aio-nextcloud-deployment.yaml
+++ b/helm-chart/templates/nextcloud-aio-nextcloud-deployment.yaml
@@ -1,0 +1,129 @@
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  annotations:
+    kompose.cmd: kompose convert -c -f latest.yml
+    kompose.version: 1.27.0 (b0ed6a2c9)
+  labels:
+    io.kompose.service: nextcloud-aio-nextcloud
+  name: nextcloud-aio-nextcloud
+spec:
+  replicas: 1
+  selector:
+    matchLabels:
+      io.kompose.service: nextcloud-aio-nextcloud
+  template:
+    metadata:
+      annotations:
+        kompose.cmd: kompose convert -c -f latest.yml
+        kompose.version: 1.27.0 (b0ed6a2c9)
+      labels:
+        io.kompose.network/nextcloud-aio: "true"
+        io.kompose.service: nextcloud-aio-nextcloud
+    spec:
+      containers:
+        - env:
+            - name: ADDITIONAL_APKS
+              value: "{{ .Values.NEXTCLOUD_ADDITIONAL_APKS }}"
+            - name: ADDITIONAL_PHP_EXTENSIONS
+              value: "{{ .Values.NEXTCLOUD_ADDITIONAL_PHP_EXTENSIONS }}"
+            - name: ADMIN_PASSWORD
+              value: "{{ .Values.NEXTCLOUD_PASSWORD }}"
+            - name: ADMIN_USER
+              value: admin
+            - name: AIO_TOKEN
+              value: "{{ .Values.AIO_TOKEN }}"
+            - name: AIO_URL
+              value: "{{ .Values.AIO_URL }}"
+            - name: CLAMAV_ENABLED
+              value: "{{ .Values.CLAMAV_ENABLED }}"
+            - name: CLAMAV_HOST
+              value: nextcloud-aio-clamav
+            - name: COLLABORA_ENABLED
+              value: "{{ .Values.COLLABORA_ENABLED }}"
+            - name: COLLABORA_HOST
+              value: nextcloud-aio-collabora
+            - name: FULLTEXTSEARCH_ENABLED
+              value: "{{ .Values.FULLTEXTSEARCH_ENABLED }}"
+            - name: FULLTEXTSEARCH_HOST
+              value: nextcloud-aio-fulltextsearch
+            - name: IMAGINARY_ENABLED
+              value: "{{ .Values.IMAGINARY_ENABLED }}"
+            - name: IMAGINARY_HOST
+              value: nextcloud-aio-imaginary
+            - name: NC_DOMAIN
+              value: "{{ .Values.NC_DOMAIN }}"
+            - name: NEXTCLOUD_DATA_DIR
+              value: /mnt/ncdata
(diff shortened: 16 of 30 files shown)


helm-chart/Chart.yaml
```diff
@@ -0,0 +1,8 @@
+name: latest
+description: A generated Helm Chart for latest from Skippbox Kompose
+version: 0.0.1
+apiVersion: v1
+keywords:
+  - latest
+sources:
+home:
```
szaimen, line 8: also automatically update versions, description and co?

helm-chart/readme.md
```diff
@@ -0,0 +1 @@
+# You can also install the AIO containers on Kubernetes using this Helm Chart
```
szaimen, line 1: add some more documentation

helm-chart/templates/nextcloud-aio-apache-deployment.yaml
```diff
@@ -0,0 +1,69 @@
...
+              value: "{{ .Values.NC_DOMAIN }}"
+            - name: NEXTCLOUD_HOST
+              value: nextcloud-aio-nextcloud
+            - name: ONLYOFFICE_HOST
+              value: nextcloud-aio-onlyoffice
+            - name: TALK_HOST
+              value: nextcloud-aio-talk
+            - name: TZ
+              value: "{{ .Values.TIMEZONE }}"
+          image: nextcloud/aio-apache:latest
```
szaimen, line 48: should use the date-latest tag from docker hub...

helm-chart/Chart.yaml
szaimen: maybe re-use the version numbers from AIO?

helm-chart/templates/nextcloud-aio-nextcloud-data-persistentvolumeclaim.yaml
```diff
@@ -0,0 +1,14 @@
...
+  creationTimestamp: null
+  labels:
+    io.kompose.service: nextcloud-aio-nextcloud-data
+  name: nextcloud-aio-nextcloud-data
+spec:
+  accessModes:
+    - ReadWriteOnce
+  resources:
+    requests:
+      storage: 100Mi
```
szaimen, line 13: maybe we need to add a variable for this?

szaimen: or rather change the default?

helm-chart/values.yaml
```diff
@@ -0,0 +1,32 @@
...
+NEXTCLOUD_ADDITIONAL_PHP_EXTENSIONS: imagick        # This allows to add additional php extensions to the Nextcloud container permanently. Default is imagick but can be overwritten by modifying this value.
+NEXTCLOUD_DATADIR: nextcloud_aio_nextcloud_data          # You can change this to e.g. /mnt/ncdata to map it to a location on your host. It needs to be adjusted before the first startup and never afterwards!
+NEXTCLOUD_MAX_TIME: 3600          # This allows to change the upload time limit of the Nextcloud container
+NEXTCLOUD_MEMORY_LIMIT: 512M          # This allows to change the PHP memory limit of the Nextcloud container
+NEXTCLOUD_MOUNT: /mnt/          # This allows the Nextcloud container to access directories on the host. It must never be equal to the value of NEXTCLOUD_DATADIR!
+NEXTCLOUD_PASSWORD: ""          # TODO! This is the password of the initially created Nextcloud admin with username admin.
+NEXTCLOUD_STARTUP_APPS: deck tasks calendar contacts files_lock        # Allows to modify the Nextcloud apps that are installed on starting AIO the first time
```
szaimen, line 21: do spaces work here?

st3iny: Yes, this is perfectly fine.

helm-chart/values.yaml
```diff
@@ -0,0 +1,32 @@
+AIO_TOKEN: 123456          # Has no function but needs to be set!
+AIO_URL: localhost          # Has no function but needs to be set!
+APACHE_MAX_SIZE: 10737418240          # This needs to be an integer and in sync with NEXTCLOUD_UPLOAD_LIMIT
+APACHE_PORT: 443          # Changing this to a different value than 443 will allow you to run it behind a reverse proxy.
+CLAMAV_ENABLED: no          # Setting this to yes enables the option in Nextcloud automatically.
+COLLABORA_DICTIONARIES: de_DE en_GB en_US es_ES fr_FR it nl pt_BR pt_PT ru        # You can change this in order to enable other dictionaries for collabora
```
szaimen, line 6: Do spaces work?

helm-chart/values.yaml
```diff
@@ -0,0 +1,32 @@
...
+DATABASE_PASSWORD: ""          # TODO! This needs to be a unique and good password!
+FULLTEXTSEARCH_ENABLED: no          # Setting this to yes enables the option in Nextcloud automatically.
+IMAGINARY_ENABLED: no          # Setting this to yes enables the option in Nextcloud automatically.
+JANUS_API_KEY: ""          # TODO! This needs to be a unique and good password!
+NC_DOMAIN: yourdomain.com          # TODO! Needs to be changed to the domain that you want to use for Nextcloud.
+NEXTCLOUD_ADDITIONAL_APKS: imagemagick        # This allows to add additional packages to the Nextcloud container permanently. Default is imagemagick but can be overwritten by modifying this value.
+NEXTCLOUD_ADDITIONAL_PHP_EXTENSIONS: imagick        # This allows to add additional php extensions to the Nextcloud container permanently. Default is imagick but can be overwritten by modifying this value.
+NEXTCLOUD_DATADIR: nextcloud_aio_nextcloud_data          # You can change this to e.g. /mnt/ncdata to map it to a location on your host. It needs to be adjusted before the first startup and never afterwards!
```
szaimen, line 16: This should not be here

helm-chart/values.yaml
```diff
@@ -0,0 +1,32 @@
...
+NEXTCLOUD_MAX_TIME: 3600          # This allows to change the upload time limit of the Nextcloud container
+NEXTCLOUD_MEMORY_LIMIT: 512M          # This allows to change the PHP memory limit of the Nextcloud container
+NEXTCLOUD_MOUNT: /mnt/          # This allows the Nextcloud container to access directories on the host. It must never be equal to the value of NEXTCLOUD_DATADIR!
```
szaimen, line 19: As well

helm-chart/values.yaml
st3iny: Will be parsed as a string: "deck tasks calendar contacts files_lock"

helm-chart/values.yaml
```diff
@@ -0,0 +1,32 @@
...
+NEXTCLOUD_PASSWORD: ""          # TODO! This is the password of the initially created Nextcloud admin with username admin.
+NEXTCLOUD_STARTUP_APPS: deck tasks calendar contacts files_lock        # Allows to modify the Nextcloud apps that are installed on starting AIO the first time
+NEXTCLOUD_TRUSTED_CACERTS_DIR: /usr/local/share/ca-certificates/my-custom-ca          # Nextcloud container will trust all the Certification Authorities, whose certificates are included in the given directory.
```
szaimen, line 22: As well

helm-chart/Chart.yaml
st3iny: From my experience, many projects version their charts independently from the software itself. This is more confusing for the user but sometimes changes have to be made to the chart alone. Those changes can be reflected in a version bump of the chart.

A pure AIO version bump could then be realized via a minor bump of the chart.

helm-chart/templates/nextcloud-aio-nextcloud-deployment.yaml
```diff
@@ -0,0 +1,137 @@
...
+            - name: IMAGINARY_ENABLED
+              value: "{{ .Values.IMAGINARY_ENABLED }}"
+            - name: IMAGINARY_HOST
+              value: nextcloud-aio-imaginary
+            - name: NC_DOMAIN
+              value: "{{ .Values.NC_DOMAIN }}"
+            - name: NEXTCLOUD_DATA_DIR
+              value: /mnt/ncdata
+            - name: NEXTCLOUD_MOUNT
+              value: "{{ .Values.NEXTCLOUD_MOUNT }}"
```
szaimen, line 63: Should not be configurable or maybw this onw actuall could...

helm-chart/templates/nextcloud-aio-nextcloud-deployment.yaml
```diff
@@ -0,0 +1,137 @@
...
+              value: "{{ .Values.UPDATE_NEXTCLOUD_APPS }}"
+          image: nextcloud/aio-nextcloud:latest
+          name: nextcloud-aio-nextcloud
+          resources: {}
+          volumeMounts:
+            - mountPath: /var/www/html
+              name: nextcloud-aio-nextcloud
+            - mountPath: /mnt/ncdata
+              name: nextcloud-aio-nextcloud-data
+            - mountPath: /mnt/
```
szaimen, line 116: allow to adjust this

helm-chart/values.yaml
szaimen: No, inside the container it can be adjustable or does thid really make sense?

helm-chart/templates/nextcloud-aio-apache-deployment.yaml
```diff
@@ -0,0 +1,69 @@
...
+      terminationGracePeriodSeconds: 10
+      volumes:
+        - name: nextcloud-aio-nextcloud
+          persistentVolumeClaim:
+            claimName: nextcloud-aio-nextcloud
+            readOnly: true
+        - name: nextcloud-aio-apache
+          persistentVolumeClaim:
+            claimName: nextcloud-aio-apache
+status: {}
```
st3iny, line 69:
```suggestion
```

Applies to all other manifests as well. The status is read-only.

helm-chart/templates/nextcloud-aio-nextcloud-data-persistentvolumeclaim.yaml
```diff
@@ -0,0 +1,14 @@
+apiVersion: v1
+kind: PersistentVolumeClaim
+metadata:
+  creationTimestamp: null
+  labels:
+    io.kompose.service: nextcloud-aio-nextcloud-data
+  name: nextcloud-aio-nextcloud-data
+spec:
+  accessModes:
```
st3iny, line 9:
```suggestion
spec:
  storageClassName: {{ .Values.STORAGE_CLASS }}
  accessModes:
```

Let admins configure their storage classes. This also applies to other manifests.

This should probably be guarded by an if condition. The default storage class will only be applied if the `storageClassName` is missing.

helm-chart/templates/nextcloud-aio-apache-deployment.yaml
```diff
@@ -0,0 +1,69 @@
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  annotations:
+    kompose.cmd: kompose convert -c -f latest.yml
+    kompose.version: 1.27.0 (b0ed6a2c9)
+  creationTimestamp: null
```
st3iny, line 7:
```suggestion
```

This is a leftover I presume. It's probably a good idea to remove `creationTimestamp` from all manifests.

st3iny (commented):
Good stuff! I left some comments.

st3iny:
I just realized that all manifests are generated through a script. That is a sensible approach for the initial version of the chart. However, I would not recommend keeping it automated as there are many moving parts that require manual fixing/intervention.

If you keep it automated, it would make sense to couple the version of the chart with the version of AIO.

helm-chart/templates/nextcloud-aio-apache-deployment.yaml
st3iny: Note: Keeping the status won't break anything as k8s will just ignore the field.

helm-chart/templates/nextcloud-aio-nextcloud-persistentvolumeclaim.yaml
```diff
@@ -0,0 +1,14 @@
+apiVersion: v1
+kind: PersistentVolumeClaim
+metadata:
+  creationTimestamp: null
+  labels:
+    io.kompose.service: nextcloud-aio-nextcloud
+  name: nextcloud-aio-nextcloud
+spec:
+  accessModes:
+    - ReadOnlyMany
```
szaimen, line 10: Should be writeabls?

helm-chart/templates/nextcloud-aio-nextcloud-trusted-cacerts-persistentvolumeclaim.yaml
```diff
@@ -0,0 +1,14 @@
+apiVersion: v1
+kind: PersistentVolumeClaim
+metadata:
+  creationTimestamp: null
+  labels:
+    io.kompose.service: nextcloud-aio-nextcloud-trusted-cacerts
+  name: nextcloud-aio-nextcloud-trusted-cacerts
+spec:
+  accessModes:
+    - ReadOnlyMany
```
szaimen, line 10: Should be writeabwe?

.github/workflows/update-helm.yml
```diff
@@ -10,6 +10,7 @@ jobs:
     name: update helm chart
     runs-on: ubuntu-latest
     steps:
+      # TODO: check for new docker images via docker hub repo.
```
szaimen, line 13: this also needs to check if a new container update is available and only update the hel chart then.

helm-chart/values.yaml
szaimen: perfect, thanks

helm-chart/values.yaml
szaimen: works

helm-chart/templates/nextcloud-aio-nextcloud-data-persistentvolumeclaim.yaml
szaimen: done

helm-chart/values.yaml
szaimen: done

helm-chart/values.yaml
szaimen: done

helm-chart/values.yaml
szaimen: done

helm-chart/templates/nextcloud-aio-nextcloud-deployment.yaml
szaimen: done

helm-chart/templates/nextcloud-aio-nextcloud-deployment.yaml
szaimen: done

helm-chart/templates/nextcloud-aio-apache-deployment.yaml
szaimen: done

helm-chart/templates/nextcloud-aio-apache-deployment.yaml
szaimen: done

helm-chart/templates/nextcloud-aio-nextcloud-persistentvolumeclaim.yaml
szaimen: done

helm-chart/templates/nextcloud-aio-nextcloud-trusted-cacerts-persistentvolumeclaim.yaml
szaimen: done

helm-chart/readme.md
szaimen: done

helm-chart/templates/nextcloud-aio-nextcloud-data-persistentvolumeclaim.yaml
szaimen: done

helm-chart/values.yaml
```diff
@@ -0,0 +1,31 @@
+IMAGE_TAG: latest          # Version of docker images, should be latest or latest-arm64. Note: latest-arm64 has no clamav support
```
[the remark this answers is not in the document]
szaimen: done

.github/workflows/update-helm.yml
szaimen: done

helm-chart/values.yaml
```diff
@@ -0,0 +1,34 @@
...
+ONLYOFFICE_ENABLED: no          # Setting this to yes enables the option in Nextcloud automatically.
+ONLYOFFICE_SECRET:           # TODO! This needs to be a unique and good password!
+REDIS_PASSWORD:           # TODO! This needs to be a unique and good password!
+SIGNALING_SECRET:           # TODO! This needs to be a unique and good password!
+TALK_ENABLED: yes          # Setting this to yes enables the option in Nextcloud automatically.
+TALK_PORT: 3478          # This allows to adjust the port that the talk container is using.
+TIMEZONE: Europe/Berlin          # TODO! This is the timezone that your containers will use.
+TURN_SECRET:           # TODO! This needs to be a unique and good password!
+UPDATE_NEXTCLOUD_APPS: no          # When setting to yes, it will automatically update all installed Nextcloud apps upon container startup on saturdays.
+MAX_STORAGE_SIZE: 10Ti        # You can adjust the max storage that each volume can use with this value
```
szaimen, line 33: Wdyt, is this a good default?

helm-chart/templates/nextcloud-aio-onlyoffice-deployment.yaml
```diff
@@ -0,0 +1,48 @@
...
+  labels:
+    io.kompose.service: nextcloud-aio-onlyoffice
+  name: nextcloud-aio-onlyoffice
+spec:
+  replicas: 1
+  selector:
+    matchLabels:
+      io.kompose.service: nextcloud-aio-onlyoffice
+  strategy:
+    type: Recreate
```
szaimen, line 17: Is this a good strategy?

helm-chart/templates/nextcloud-aio-talk-deployment.yaml
```diff
@@ -0,0 +1,48 @@
...
+    kompose.version: 1.27.0 (b0ed6a2c9)
+  labels:
+    io.kompose.service: nextcloud-aio-talk
+  name: nextcloud-aio-talk
+spec:
+  replicas: 1
+  selector:
+    matchLabels:
+      io.kompose.service: nextcloud-aio-talk
+  strategy: {}
```
szaimen, line 16: Do we always need a strategy?

helm-chart/templates/nextcloud-aio-talk-deployment.yaml
```diff
@@ -0,0 +1,48 @@
...
+              value: "{{ .Values.TURN_SECRET }}"
+            - name: TZ
+              value: "{{ .Values.TIMEZONE }}"
+          image: nextcloud/aio-talk:20221229_091124-{{ .Values.IMAGE_TAG }}
+          name: nextcloud-aio-talk
+          ports:
+            - containerPort: {{ .Values.TALK_PORT }}
+            - containerPort: {{ .Values.TALK_PORT }}
+              protocol: UDP
+          resources: {}
```
szaimen, line 46: Are ressources unlimitsd if not provided?

helm-chart/templates/nextcloud-aio-apache-deployment.yaml
```diff
@@ -0,0 +1,65 @@
...
+          ports:
+            - containerPort: {{ .Values.APACHE_PORT }}
+          resources: {}
+          volumeMounts:
+            - mountPath: /var/www/html
+              name: nextcloud-aio-nextcloud
+              readOnly: true
+            - mountPath: /mnt/data
+              name: nextcloud-aio-apache
+      restartPolicy: Always
```
szaimen, line 57: Is this good or is there also something like unless-stopped?

helm-chart/values.yaml
st3iny: IMO, 10Ti is a bit high. Volumes are provisioned statically with the requested size right away. If required later, a volume can be resized at runtime by increasing the storage request of its claim. Kubernetes and the underlying driver will handle the rest.

Also, it might be a good idea to differentiate the requested size based on the volume/application.

helm-chart/templates/nextcloud-aio-talk-deployment.yaml
st3iny: Exactly. You could omit the whole `resources` key.

helm-chart/templates/nextcloud-aio-onlyoffice-deployment.yaml
st3iny: Yes, this is a reasonable default (especially for non-distributed applications). 

However, given that there is only one replica, it doesn't make much of a difference.

helm-chart/templates/nextcloud-aio-apache-deployment.yaml
st3iny: No, there is no such thing in Kubernetes. Stopping means deleting the deployment (or scaling it to 0 replicas).

You could omit the `restartPolicy` entirely, as it will default to `Always` anyway.

helm-chart/templates/nextcloud-aio-talk-deployment.yaml
st3iny: No. I recommend omitting it, as I'm not sure if the current version is valid (for Kubernetes). The strategy defaults to `RollingUpdate` which means replacing pods one by one. 

Again, as there is only a single replica, there is not much difference to the alternative strategy `Recreate`.

helm-chart/templates/nextcloud-aio-nextcloud-deployment.yaml
```diff
@@ -0,0 +1,133 @@
...
+              value: "{{ .Values.NEXTCLOUD_TRUSTED_CACERTS_DIR }}"
+            - name: TURN_SECRET
+              value: "{{ .Values.TURN_SECRET }}"
+            - name: TZ
+              value: "{{ .Values.TIMEZONE }}"
+            - name: UPDATE_NEXTCLOUD_APPS
+              value: "{{ .Values.UPDATE_NEXTCLOUD_APPS }}"
+          image: nextcloud/aio-nextcloud:20221229_091124-{{ .Values.IMAGE_TAG }}
+          name: nextcloud-aio-nextcloud
+          resources: {}
```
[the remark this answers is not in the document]
st3iny: Exactly. You could omit the whole resources key.

helm-chart/values.yaml
szaimen: changed the default to 10Gi. Having different defaults makes indeed sense...

helm-chart/templates/nextcloud-aio-talk-deployment.yaml
szaimen: removed strategies

helm-chart/templates/nextcloud-aio-talk-deployment.yaml
szaimen: removed ressources

helm-chart/templates/nextcloud-aio-nextcloud-deployment.yaml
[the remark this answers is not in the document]
szaimen: removed ressources

helm-chart/templates/nextcloud-aio-apache-deployment.yaml
szaimen: removed resstarpolicy

helm-chart/values.yaml
szaimen: will be done in a follow-up

helm-chart/templates/nextcloud-aio-talk-deployment.yaml
```diff
@@ -0,0 +1,45 @@
...
+            - name: TALK_PORT
+              value: "{{ .Values.TALK_PORT }}"
+            - name: TURN_SECRET
+              value: "{{ .Values.TURN_SECRET }}"
+            - name: TZ
+              value: "{{ .Values.TIMEZONE }}"
+          image: nextcloud/aio-talk:20221229_091124-{{ .Values.IMAGE_TAG }}
+          name: nextcloud-aio-talk
+          ports:
+            - containerPort: {{ .Values.TALK_PORT }}
```
provokateurin, line 42: Maybe explicitly set it to TCP to avoid confusion.

helm-chart/templates/nextcloud-aio-talk-service.yaml
```diff
@@ -0,0 +1,24 @@
...
+    kompose.cmd: kompose convert -c -f latest.yml
+    kompose.version: 1.27.0 (b0ed6a2c9)
+  labels:
+    io.kompose.service: nextcloud-aio-talk
+  name: nextcloud-aio-talk
+spec:
+  ports:
+    - name: "{{ .Values.TALK_PORT }}"
+      port: {{ .Values.TALK_PORT }}
+      targetPort: {{ .Values.TALK_PORT }}
```
provokateurin, line 15: Also here better set it to TCP explicitly

helm-chart/templates/nextcloud-aio-talk-service.yaml
```diff
@@ -0,0 +1,24 @@
...
+    - name: {{ .Values.TALK_PORT }}-udp
+      port: {{ .Values.TALK_PORT }}
+      protocol: UDP
+      targetPort: {{ .Values.TALK_PORT }}
+  selector:
+    io.kompose.service: nextcloud-aio-talk
+status:
+  loadBalancer: {}
```
provokateurin, line 23: Useless?

helm-chart/values.yaml
```diff
@@ -0,0 +1,34 @@
...
+ONLYOFFICE_SECRET:           # TODO! This needs to be a unique and good password!
+REDIS_PASSWORD:           # TODO! This needs to be a unique and good password!
+SIGNALING_SECRET:           # TODO! This needs to be a unique and good password!
+TALK_ENABLED: yes          # Setting this to yes enables the option in Nextcloud automatically.
+TALK_PORT: 3478          # This allows to adjust the port that the talk container is using.
+TIMEZONE: Europe/Berlin          # TODO! This is the timezone that your containers will use.
+TURN_SECRET:           # TODO! This needs to be a unique and good password!
+UPDATE_NEXTCLOUD_APPS: no          # When setting to yes, it will automatically update all installed Nextcloud apps upon container startup on saturdays.
+MAX_STORAGE_SIZE: 10Gi        # You can adjust the max storage that each volume can use with this value
+STORAGE_CLASS:        # By setting this, you can adjust the storage class for your volumes
```
provokateurin, line 34: This format obviously works, but is quite unconventional. It's usually nested objects, which you should be able to do by splitting each underscore into a new level (could be hard with bash script though).

helm-chart/templates/nextcloud-aio-apache-persistentvolumeclaim.yaml
```diff
@@ -0,0 +1,15 @@
...
+metadata:
+  labels:
+    io.kompose.service: nextcloud-aio-apache
+  name: nextcloud-aio-apache
+spec:
+  {{- if .Values.STORAGE_CLASS }}
+  storageClassName: {{ .Values.STORAGE_CLASS }}
+  {{- end }}
+  accessModes:
+    - ReadWriteMany
```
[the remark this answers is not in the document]
st3iny: Good point. Common block storage providers only allow RWO anyway.

helm-chart/templates/nextcloud-aio-fulltextsearch-deployment.yaml
```diff
@@ -0,0 +1,42 @@
...
+        kompose.cmd: kompose convert -c -f latest.yml
+        kompose.version: 1.27.0 (b0ed6a2c9)
+      labels:
+        io.kompose.network/nextcloud-aio: "true"
+        io.kompose.service: nextcloud-aio-fulltextsearch
+    spec:
+      containers:
+        - env:
+            - name: ES_JAVA_OPTS
+              value: -Xms1024M -Xmx1024M
```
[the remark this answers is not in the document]
st3iny: It might also be a good idea to add resource request to this specific deployment in order to prevent failure caused by being scheduled on an insufficient pod.

helm-chart/templates/nextcloud-aio-networkpolicy.yaml
```diff
@@ -0,0 +1,13 @@
+apiVersion: networking.k8s.io/v1
+kind: NetworkPolicy
```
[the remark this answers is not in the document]
st3iny: Policies are a pretty advanced feature and not easy to test. How about implementing them later and require users/admins to not enforce policies for now?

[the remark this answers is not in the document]
provokateurin: Sounds good.

helm-chart/templates/nextcloud-aio-apache-deployment.yaml
```diff
@@ -0,0 +1,61 @@
...
+              value: "{{ .Values.NC_DOMAIN }}"
+            - name: NEXTCLOUD_HOST
+              value: nextcloud-aio-nextcloud
+            - name: ONLYOFFICE_HOST
+              value: nextcloud-aio-onlyoffice
+            - name: TALK_HOST
+              value: nextcloud-aio-talk
+            - name: TZ
+              value: "{{ .Values.TIMEZONE }}"
+          image: nextcloud/aio-apache:20221229_091124-{{ .Values.IMAGE_TAG }}
```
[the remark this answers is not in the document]
szaimen: It is not. See e.g. https://hub.docker.com/layers/nextcloud/all-in-one/20221229_091124-latest/images/sha256-069b4d7d59fb1538356124163ef4ff2811c7ed5791c100ff1aebe798969bf17c?context=explore

helm-chart/templates/nextcloud-aio-apache-deployment.yaml
```diff
@@ -0,0 +1,61 @@
...
+          name: nextcloud-aio-apache
+          ports:
+            - containerPort: {{ .Values.APACHE_PORT }}
+          volumeMounts:
+            - mountPath: /var/www/html
+              name: nextcloud-aio-nextcloud
+              readOnly: true
```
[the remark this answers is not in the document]
szaimen: No, as Nextcloud is running in a different container.

helm-chart/templates/nextcloud-aio-apache-persistentvolumeclaim.yaml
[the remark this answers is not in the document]
szaimen: Will be changed in https://github.com/nextcloud/all-in-one/pull/1848

helm-chart/templates/nextcloud-aio-apache-persistentvolumeclaim.yaml
```diff
@@ -0,0 +1,15 @@
...
+  resources:
+    requests:
+      storage: {{ .Values.MAX_STORAGE_SIZE }}
```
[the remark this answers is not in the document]
szaimen: Will be changed in https://github.com/nextcloud/all-in-one/pull/1848

helm-chart/templates/nextcloud-aio-apache-service.yaml
```diff
@@ -0,0 +1,18 @@
+apiVersion: v1
+kind: Service
+metadata:
+  annotations:
+    kompose.cmd: kompose convert -c -f latest.yml
+    kompose.version: 1.27.0 (b0ed6a2c9)
+  labels:
+    io.kompose.service: nextcloud-aio-apache
+  name: nextcloud-aio-apache
+spec:
```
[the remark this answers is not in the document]
szaimen: sorry, what does this do?

helm-chart/templates/nextcloud-aio-apache-service.yaml
```diff
@@ -0,0 +1,18 @@
...
+  ports:
+    - name: "{{ .Values.APACHE_PORT }}"
+      port: {{ .Values.APACHE_PORT }}
```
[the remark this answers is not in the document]
szaimen: the port is configurable by providing apache_port?

helm-chart/templates/nextcloud-aio-apache-service.yaml
```diff
@@ -0,0 +1,18 @@
...
+      targetPort: {{ .Values.APACHE_PORT }}
+  selector:
+    io.kompose.service: nextcloud-aio-apache
+status:
+  loadBalancer: {}
```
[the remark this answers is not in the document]
szaimen: will be fixed in https://github.com/nextcloud/all-in-one/pull/1848

helm-chart/templates/nextcloud-aio-clamav-deployment.yaml
```diff
@@ -0,0 +1,40 @@
...
+        io.kompose.network/nextcloud-aio: "true"
+        io.kompose.service: nextcloud-aio-clamav
+    spec:
+      containers:
+        - env:
+            - name: CLAMD_STARTUP_TIMEOUT
+              value: "90"
+            - name: TZ
+              value: "{{ .Values.TIMEZONE }}"
+          image: nextcloud/aio-clamav:20221229_091124-{{ .Values.IMAGE_TAG }}
```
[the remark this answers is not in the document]
szaimen: see above

helm-chart/templates/nextcloud-aio-clamav-persistentvolumeclaim.yaml
```diff
@@ -0,0 +1,15 @@
...
+metadata:
+  labels:
+    io.kompose.service: nextcloud-aio-clamav
+  name: nextcloud-aio-clamav
+spec:
+  {{- if .Values.STORAGE_CLASS }}
+  storageClassName: {{ .Values.STORAGE_CLASS }}
+  {{- end }}
+  accessModes:
+    - ReadWriteMany
```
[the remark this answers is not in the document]
szaimen: see above

helm-chart/templates/nextcloud-aio-clamav-persistentvolumeclaim.yaml
```diff
@@ -0,0 +1,15 @@
...
+  resources:
+    requests:
+      storage: {{ .Values.MAX_STORAGE_SIZE }}
```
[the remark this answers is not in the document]
szaimen: will be fixed in https://github.com/nextcloud/all-in-one/pull/1848

helm-chart/templates/nextcloud-aio-collabora-deployment.yaml
```diff
@@ -0,0 +1,44 @@
...
+        - env:
+            - name: TZ
+              value: "{{ .Values.TIMEZONE }}"
+            - name: aliasgroup1
+              value: https://{{ .Values.NC_DOMAIN }}:443
+            - name: dictionaries
+              value: "{{ .Values.COLLABORA_DICTIONARIES }}"
+            - name: extra_params
+              value: --o:ssl.enable=false --o:ssl.termination=true --o:logging.level=warning --o:home_mode.enable=true {{ .Values.COLLABORA_SECCOMP_POLICY }} --o:remote_font_config.url=https://{{ .Values.NC_DOMAIN }}/apps/richdocuments/settings/fonts.json
+          image: nextcloud/aio-collabora:20221229_091124-{{ .Values.IMAGE_TAG }}
```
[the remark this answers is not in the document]
szaimen: see above

helm-chart/templates/nextcloud-aio-collabora-fonts-persistentvolumeclaim.yaml
```diff
@@ -0,0 +1,15 @@
...
+metadata:
+  labels:
+    io.kompose.service: nextcloud-aio-collabora-fonts
+  name: nextcloud-aio-collabora-fonts
+spec:
+  {{- if .Values.STORAGE_CLASS }}
+  storageClassName: {{ .Values.STORAGE_CLASS }}
+  {{- end }}
+  accessModes:
+    - ReadWriteMany
```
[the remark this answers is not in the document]
szaimen: see above

helm-chart/templates/nextcloud-aio-collabora-fonts-persistentvolumeclaim.yaml
```diff
@@ -0,0 +1,15 @@
...
+  resources:
+    requests:
+      storage: {{ .Values.MAX_STORAGE_SIZE }}
```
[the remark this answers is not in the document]
szaimen: see above

helm-chart/templates/nextcloud-aio-database-deployment.yaml
```diff
@@ -0,0 +1,50 @@
...
+              value: "{{ .Values.TIMEZONE }}"
+            - name: POSTGRES_DB
+              value: nextcloud_database
+            - name: POSTGRES_PASSWORD
+              value: "{{ .Values.DATABASE_PASSWORD }}"
+            - name: POSTGRES_USER
+              value: nextcloud
+            - name: TZ
+              value: "{{ .Values.TIMEZONE }}"
+          image: nextcloud/aio-postgresql:20221229_091124-{{ .Values.IMAGE_TAG }}
```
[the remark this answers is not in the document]
szaimen: see above

helm-chart/templates/nextcloud-aio-database-dump-persistentvolumeclaim.yaml
```diff
@@ -0,0 +1,15 @@
...
+metadata:
+  labels:
+    io.kompose.service: nextcloud-aio-database-dump
+  name: nextcloud-aio-database-dump
+spec:
+  {{- if .Values.STORAGE_CLASS }}
+  storageClassName: {{ .Values.STORAGE_CLASS }}
+  {{- end }}
+  accessModes:
+    - ReadWriteMany
```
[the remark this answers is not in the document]
szaimen: see above

helm-chart/templates/nextcloud-aio-database-dump-persistentvolumeclaim.yaml
```diff
@@ -0,0 +1,15 @@
...
+  resources:
+    requests:
+      storage: {{ .Values.MAX_STORAGE_SIZE }}
```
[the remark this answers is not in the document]
szaimen: see abovee

helm-chart/templates/nextcloud-aio-database-deployment.yaml
```diff
@@ -0,0 +1,50 @@
...
+          name: nextcloud-aio-database
+          volumeMounts:
+            - mountPath: /var/lib/postgresql/data
+              name: nextcloud-aio-database
+            - mountPath: /mnt/data
```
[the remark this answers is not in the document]
szaimen: It allows stores the database dump and other required stuff for the database container

helm-chart/templates/nextcloud-aio-fulltextsearch-deployment.yaml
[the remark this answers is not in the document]
szaimen: hm... this is also not configurable for AIO currently so would need to be addressed for AIO in general first...

helm-chart/templates/nextcloud-aio-networkpolicy.yaml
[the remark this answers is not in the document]
szaimen: I thought the network policy makes sure that the containers are in the same network and able to talk to each other? Also this was configured by kompose automatically so I thought we need it?

helm-chart/templates/nextcloud-aio-talk-deployment.yaml
szaimen: thanks for the suggestion but I think this is fine

helm-chart/templates/nextcloud-aio-talk-service.yaml
szaimen: will be addressed in https://github.com/nextcloud/all-in-one/pull/1848

helm-chart/templates/nextcloud-aio-talk-service.yaml
szaimen: thanks for the suggestion but I think this is fine

helm-chart/values.yaml
szaimen: So a Storage_class is not a string?

helm-chart/templates/nextcloud-aio-apache-deployment.yaml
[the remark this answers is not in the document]
provokateurin: I see, wouldn't it be better to have proper semver tags?

helm-chart/templates/nextcloud-aio-apache-service.yaml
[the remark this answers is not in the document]
provokateurin: NodePort exposes the port on the host. ClusterIP for example only exposes inside the cluster. It could be useful for people running without it without an ingress.

helm-chart/templates/nextcloud-aio-apache-service.yaml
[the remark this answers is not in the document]
provokateurin: Yes the targetPort is correct, but you might not want to change the apache port while changing the port of the service

helm-chart/templates/nextcloud-aio-networkpolicy.yaml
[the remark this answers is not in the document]
provokateurin: No the networking should work without it if everything is configured correctly. I have never used a NetworkPolicy, but this one also seems to not do much?

helm-chart/values.yaml
provokateurin: It was meant for the whole file. storage_class is actually fine, but I'd like to see something like this (which is how it's done normally):
```yaml
image:
  tag: latest
aio:
  token: 123456
  url: localhost
apache:
  max_size: 10737418240
  port: 443
...
```

helm-chart/templates/nextcloud-aio-apache-deployment.yaml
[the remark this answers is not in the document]
provokateurin: Also since the tag is configured in the values file, it should be `nextcloud/aio-apache:{{ .Values.IMAGE_TAG }}`!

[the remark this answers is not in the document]
szaimen: > nextcloud/aio-apache:{{ .Values.IMAGE_TAG }}

No, as all the other config depends on the exact version. So the helm chart comes bundled with the versioned containers that match their config in the chart itself.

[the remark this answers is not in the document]
provokateurin: Meh, this isn't how a Helm Chart should be done

helm-chart/templates/nextcloud-aio-apache-service.yaml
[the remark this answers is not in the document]
szaimen: So a service is not enough to make this run on all different cluster types?
See https://github.com/nextcloud/all-in-one/blob/main/helm-chart/templates/nextcloud-aio-apache-service.yaml

helm-chart/templates/nextcloud-aio-networkpolicy.yaml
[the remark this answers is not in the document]
szaimen: Interesting! So that is thdn something where docker and kubernetes differ. Is it okay if I keep this?

helm-chart/values.yaml
szaimen: Ah I see. Yes this is a bit hard to do with a bsh script so I wiuld like to keep the current syntax.

helm-chart/templates/nextcloud-aio-networkpolicy.yaml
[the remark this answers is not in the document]
provokateurin: If it works, yes I guess

helm-chart/templates/nextcloud-aio-apache-deployment.yaml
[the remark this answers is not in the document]
szaimen: But cannot you pin a specific helm chart version?

helm-chart/templates/nextcloud-aio-networkpolicy.yaml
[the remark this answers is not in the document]
szaimen: Okay, thanks

PR #1577, opened 2022-12-22
migrate mastercontainer to alpine

Zoey2936:
- [x] TODO: check if the www-data user has the same id
- [x] Needs: #1587

Build works, but I didn't test it at this pint, so needs to be good tested

### DIFF
--- a/Containers/mastercontainer/Dockerfile
+++ b/Containers/mastercontainer/Dockerfile
@@ -9,7 +9,6 @@ FROM php:8.1.14-fpm-alpine3.17
 
 RUN set -ex; \
     apk add --no-cache shadow; \
-    deluser apache2; \
     groupmod -g 333 xfs; \
     usermod -u 333 -g 333 xfs; \
     addgroup -g 33 -S www-data; \
--- a/Containers/mastercontainer/Dockerfile
+++ b/Containers/mastercontainer/Dockerfile
@@ -87,21 +87,21 @@ RUN mkdir -p /etc/apache2/certs && \
 COPY mastercontainer.conf /etc/apache2/sites-available/
 
 RUN sed -i \
-                -e '/^Listen /d' \
-                -e 's/User apache/User www-data/g' \
-                -e 's/Group apache/Group www-data/g' \
-                -e 's/^#\(LoadModule .*mod_rewrite.so\)/\1/' \
-                -e 's/^#\(LoadModule .*mod_headers.so\)/\1/' \
-                -e 's/^#\(LoadModule .*mod_env.so\)/\1/' \
-                -e 's/^#\(LoadModule .*mod_mime.so\)/\1/' \
-                -e 's/^#\(LoadModule .*mod_dir.so\)/\1/' \
-                -e 's/^#\(LoadModule .*mod_authz_core.so\)/\1/' \
-                -e 's/^#\(LoadModule .*mod_proxy.so\)/\1/' \
-                -e 's/^#\(LoadModule .*mod_proxy_http.so\)/\1/' \
-                -e 's/^#\(LoadModule .*mod_ssl.so\)/\1/' \
-                -e 's/^#\(LoadModule .*mod_mpm_event.so\)/\1/' \
-                -e 's/\(LoadModule .*mod_mpm_worker.so\)/#\1/' \
-                -e 's/\(LoadModule .*mod_mpm_prefork.so\)/#\1/' \
+            -e '/^Listen /d' \
+            -e 's/User apache/User www-data/g' \
+            -e 's/Group apache/Group www-data/g' \
+            -e 's/^#\(LoadModule .*mod_rewrite.so\)/\1/' \
+            -e 's/^#\(LoadModule .*mod_headers.so\)/\1/' \
+            -e 's/^#\(LoadModule .*mod_env.so\)/\1/' \
+            -e 's/^#\(LoadModule .*mod_mime.so\)/\1/' \
+            -e 's/^#\(LoadModule .*mod_dir.so\)/\1/' \
+            -e 's/^#\(LoadModule .*mod_authz_core.so\)/\1/' \
+            -e 's/^#\(LoadModule .*mod_proxy.so\)/\1/' \
+            -e 's/^#\(LoadModule .*mod_proxy_http.so\)/\1/' \
+            -e 's/^#\(LoadModule .*mod_ssl.so\)/\1/' \
+            -e 's/^#\(LoadModule .*mod_mpm_event.so\)/\1/' \
+            -e 's/\(LoadModule .*mod_mpm_worker.so\)/#\1/' \
+            -e 's/\(LoadModule .*mod_mpm_prefork.so\)/#\1/' \
         /etc/apache2/httpd.conf; \
         echo "ServerName localhost" | tee -a /etc/apache2/httpd.conf
 
--- a/Containers/mastercontainer/Dockerfile
+++ b/Containers/mastercontainer/Dockerfile
@@ -103,6 +103,7 @@ RUN sed -i \
             -e 's/\(LoadModule .*mod_mpm_prefork.so\)/#\1/' \
         /etc/apache2/httpd.conf; \
         mkdir -p /etc/apache2/logs; \
+        rm /etc/apache2/conf.d/ssl.conf; \
         echo "ServerName localhost" | tee -a /etc/apache2/httpd.conf; \
         echo "Include /etc/apache2/sites-available/mastercontainer.conf" | tee -a /etc/apache2/httpd.conf
 
--- a/Containers/mastercontainer/Dockerfile
+++ b/Containers/mastercontainer/Dockerfile
@@ -36,7 +36,8 @@ RUN set -ex; \
         supervisor \
         openssl \
         sudo \
-        netcat-openbsd
+        netcat-openbsd \
+        grep
 
 RUN set -ex; \
     apk add --no-cache --virtual .build-deps \
--- a/Containers/mastercontainer/Dockerfile
+++ b/Containers/mastercontainer/Dockerfile
@@ -79,7 +79,6 @@ RUN sed -i \
 		-e '/^Listen /d' \
 		-e 's/User apache/User www-data/g' \
 		-e 's/Group apache/Group www-data/g' \
-		-e 's/\(LoadModule .*mod_mpm_prefork.so\)/#\1/' \
 		-e 's/^#\(LoadModule .*mod_rewrite.so\)/\1/' \
 		-e 's/^#\(LoadModule .*mod_headers.so\)/\1/' \
 		-e 's/^#\(LoadModule .*mod_env.so\)/\1/' \
--- a/Containers/mastercontainer/Dockerfile
+++ b/Containers/mastercontainer/Dockerfile
@@ -87,21 +87,21 @@ RUN mkdir -p /etc/apache2/certs && \
 COPY mastercontainer.conf /etc/apache2/sites-available/
 
 RUN sed -i \
-		-e '/^Listen /d' \
-		-e 's/User apache/User www-data/g' \
-		-e 's/Group apache/Group www-data/g' \
-		-e 's/^#\(LoadModule .*mod_rewrite.so\)/\1/' \
-		-e 's/^#\(LoadModule .*mod_headers.so\)/\1/' \
-		-e 's/^#\(LoadModule .*mod_env.so\)/\1/' \
-		-e 's/^#\(LoadModule .*mod_mime.so\)/\1/' \
-		-e 's/^#\(LoadModule .*mod_dir.so\)/\1/' \
-		-e 's/^#\(LoadModule .*mod_authz_core.so\)/\1/' \
-		-e 's/^#\(LoadModule .*mod_proxy.so\)/\1/' \
-		-e 's/^#\(LoadModule .*mod_proxy_http.so\)/\1/' \
-		-e 's/^#\(LoadModule .*mod_ssl.so\)/\1/' \
-		-e 's/^#\(LoadModule .*mod_mpm_event.so\)/\1/' \
-		-e 's/\(LoadModule .*mod_mpm_worker.so\)/#\1/' \
-		-e 's/\(LoadModule .*mod_mpm_prefork.so\)/#\1/' \
+                -e '/^Listen /d' \
+                -e 's/User apache/User www-data/g' \
+                -e 's/Group apache/Group www-data/g' \
+                -e 's/^#\(LoadModule .*mod_rewrite.so\)/\1/' \
+                -e 's/^#\(LoadModule .*mod_headers.so\)/\1/' \
+                -e 's/^#\(LoadModule .*mod_env.so\)/\1/' \
+                -e 's/^#\(LoadModule .*mod_mime.so\)/\1/' \
+                -e 's/^#\(LoadModule .*mod_dir.so\)/\1/' \
+                -e 's/^#\(LoadModule .*mod_authz_core.so\)/\1/' \
+                -e 's/^#\(LoadModule .*mod_proxy.so\)/\1/' \
+                -e 's/^#\(LoadModule .*mod_proxy_http.so\)/\1/' \
+                -e 's/^#\(LoadModule .*mod_ssl.so\)/\1/' \
+                -e 's/^#\(LoadModule .*mod_mpm_event.so\)/\1/' \
+                -e 's/\(LoadModule .*mod_mpm_worker.so\)/#\1/' \
+                -e 's/\(LoadModule .*mod_mpm_prefork.so\)/#\1/' \
         /etc/apache2/httpd.conf; \
         echo "ServerName localhost" | tee -a /etc/apache2/httpd.conf
 
--- a/Containers/mastercontainer/Dockerfile
+++ b/Containers/mastercontainer/Dockerfile
@@ -36,7 +36,7 @@ RUN set -ex; \
 
 RUN set -ex; \
     apk add --no-cache shadow; \
-    deluser www-data; \
+    deluser apache2; \
     groupmod -g 333 xfs; \
     usermod -u 333 -g 333 xfs; \
     addgroup -g 33 -S www-data; \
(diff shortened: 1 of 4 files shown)


Containers/mastercontainer/Dockerfile
```diff
@@ -5,7 +5,7 @@ FROM docker:20.10.21-dind-alpine3.16 as dind
 FROM caddy:2.6.2-alpine as caddy
 
 # From https://github.com/docker-library/php/blob/master/8.0/bullseye/apache/Dockerfile
-FROM php:8.0.26-apache-bullseye
+FROM alpine:3.16.3
```
szaimen, line 8: Thanks! But can we maybe use the php-alpine image for this or is this not possible?

Zoey2936: I think it isn't possible or very hard to integrate, this is the easiest way I see
Maybe we could use the php:fpm-alpine image and run fpm using supervisor

szaimen: > I think it isn't possible or very hard to integrate, this is the easiest way I see

Ah I see.

> Maybe we could use the php:fpm-alpine image and run fpm using supervisor

Yes, I would prefer this over building from alpine directly

Zoey2936: @szaimen do you think this could work?

szaimen: Yes, looks pretty good to me already! Thanks a lot! :)
I will also include this in the release after 4.0.1 :)

Zoey2936:
The www-data user is called on apache2 on alpine

Containers/mastercontainer/Dockerfile
```diff
@@ -4,8 +4,8 @@ FROM docker:20.10.21-dind-alpine3.16 as dind
 # Caddy is a requirement
 FROM caddy:2.6.2-alpine as caddy
 
-# From https://github.com/docker-library/php/blob/master/8.0/bullseye/apache/Dockerfile
-FROM php:8.1.13-apache-bullseye
+# From https://github.com/docker-library/php/blob/master/8.1/alpine3.17/fpm/Dockerfile
+FROM php:8.1.13-fpm-alpine3.17
```
szaimen, line 8: lets wait for alpine 3.17.1 then as well

kri164: Current Alpine Version 3.17.1 (Released Jan 09, 2023)

Zoey2936: Thanks for reminding for this, but it seems to be not available yeet on docker hub: https://hub.docker.com/_/alpine/tags?name=3.17.1
And even then it could take a while until the PHP repo updates to alpine v3.17.1

Zoey2936: the php image was now pushed to 3.17.1

szaimen: It looks like I will not get to this due to time problems before the next beta release which is planned for coming tuesday. Sorry for that! I'll try to include it in the version afterwards.

szaimen:
I'd like to keep the userid and maybe also the user and group-name. Can we add something like this here as well? https://github.com/nextcloud/all-in-one/blob/b249e02d53f063e1b75df96f50007a329d94717e/Containers/nextcloud/Dockerfile#L4-L11

Containers/mastercontainer/Dockerfile
```diff
@@ -19,31 +19,26 @@ RUN mkdir -p /var/www/docker-aio;
...
-        openssl \
-        sudo \
-        dpkg-dev \
-        netcat \
-    ; \
-    rm -rf /var/lib/apt/lists/*
-
+RUN set -ex; \
+    apk add --no-cache ca-certificates wget tzdata \ 
+		       bash apache2 supervisor openssl netcat-openbsd \
```
szaimen, line 24: can we keep the former formatting?

Zoey2936: done

Zoey2936:
> I'd like to keep the userid and maybe also the user and group-name. Can we add something like this here as well?
> 
> https://github.com/nextcloud/all-in-one/blob/b249e02d53f063e1b75df96f50007a329d94717e/Containers/nextcloud/Dockerfile#L4-L11

we just need the www-data user/group with id 33?

Containers/mastercontainer/supervisord.conf
```diff
@@ -8,12 +8,20 @@ logfile_backups=10
...
+user=root
+
 [program:apache]
 # stdout_logfile=/dev/stdout
 # stdout_logfile_maxbytes=0
 stderr_logfile=/dev/stderr
 stderr_logfile_maxbytes=0
-command=apache2-foreground
+command=httpd -DFOREGROUND
 user=root
```
szaimen, line 25:
```suggestion
user=www-data
```

Containers/mastercontainer/supervisord.conf
```diff
@@ -8,14 +8,22 @@ logfile_backups=10
...
 
-[program:apache]
+[program:php-fpm]
 # stdout_logfile=/dev/stdout
 # stdout_logfile_maxbytes=0
 stderr_logfile=/dev/stderr
 stderr_logfile_maxbytes=0
-command=apache2-foreground
+command=php-fpm
 user=root
```
Zoey2936, line 17: maybe also use www-data user?
```suggestion
user=www-data
```

Containers/mastercontainer/Dockerfile
```diff
@@ -4,8 +4,17 @@ FROM docker:20.10.21-dind-alpine3.16 as dind
...
 FROM caddy:2.6.2-alpine as caddy
 
-# From https://github.com/docker-library/php/blob/master/8.0/bullseye/apache/Dockerfile
-FROM php:8.1.14-apache-bullseye
+# From https://github.com/docker-library/php/blob/master/8.1/alpine3.17/fpm/Dockerfile
+FROM php:8.1.14-fpm-alpine3.17
+
+RUN set -ex; \
+    apk add --no-cache shadow; \
+    deluser apache2; \
```
szaimen, line 12:
```suggestion
```

Containers/mastercontainer/Dockerfile
```diff
@@ -52,38 +75,39 @@ RUN set -e && \
     cd ..; \
     rm -f /usr/local/bin/composer; \
     chmod 770 -R ./; \
+    chmod 777 -R /root; \
```
szaimen, line 78:
```suggestion
```

Containers/mastercontainer/Dockerfile
```diff
@@ -52,38 +75,39 @@ RUN set -e && \
...
+		-e 's/Group apache/Group www-data/g' \
+		-e 's/^#\(LoadModule .*mod_rewrite.so\)/\1/' \
+		-e 's/^#\(LoadModule .*mod_headers.so\)/\1/' \
+		-e 's/^#\(LoadModule .*mod_env.so\)/\1/' \
+		-e 's/^#\(LoadModule .*mod_mime.so\)/\1/' \
+		-e 's/^#\(LoadModule .*mod_dir.so\)/\1/' \
+		-e 's/^#\(LoadModule .*mod_authz_core.so\)/\1/' \
+		-e 's/^#\(LoadModule .*mod_mpm_event.so\)/\1/' \
+		-e 's/\(LoadModule .*mod_mpm_worker.so\)/#\1/' \
+		-e 's/\(LoadModule .*mod_mpm_prefork.so\)/#\1/' \
```
szaimen, line 102: I think proxy, proxy_http and ssl are missing here?

Containers/mastercontainer/Dockerfile
```diff
@@ -52,38 +75,39 @@ RUN set -e && \
...
+		/etc/apache2/httpd.conf
+
+RUN echo "ServerName localhost" | tee -a /etc/apache2/httpd.conf
```
szaimen, line 105:
```suggestion
        /etc/apache2/httpd.conf; \
        echo "ServerName localhost" | tee -a /etc/apache2/httpd.conf
```

Containers/mastercontainer/supervisord.conf
szaimen: yeah, we could try that 👍

Containers/mastercontainer/Dockerfile
Zoey2936: I will have look later

Containers/mastercontainer/Dockerfile
```diff
@@ -19,31 +27,46 @@ RUN mkdir -p /var/www/docker-aio;
...
-RUN apt-get update; \
-    apt-get install -y --no-install-recommends \
-        git \
+RUN set -ex; \
+    apk add --no-cache 
+        ca-certificates \
+        wget \
+        tzdata \
+        bash \
+        apache2 \
```
szaimen, line 36:
```suggestion
```

szaimen, line 36: is it really called apache and not http?

Zoey2936: yes: https://pkgs.alpinelinux.org/package/v3.17/main/x86_64/apache2

Containers/mastercontainer/start.sh
```diff
@@ -277,4 +277,7 @@ https://your-domain-that-points-to-this-server.tld:8443"
 # Set the timezone to UTC
 export TZ=UTC
```
Zoey2936, line 278: Why is the timezone forced to be utc?

szaimen: Because the logic in the mastercontainer assumes that the timezone is UTC.

szaimen: However of course all other containers except the borgbackup container use the timezone that is chosen in the AIO interface.

Zoey2936: > Because the logic in the mastercontainer assumes that the timezone is UTC.

so it breaks when using another timezone?

szaimen: Yes, basically

Containers/mastercontainer/Dockerfile
Zoey2936: done

Zoey2936:
@szaimen should now alos be done

Containers/mastercontainer/Dockerfile
```diff
@@ -54,36 +77,38 @@ RUN set -e && \
...
+		-e 's/^#\(LoadModule .*mod_env.so\)/\1/' \
+		-e 's/^#\(LoadModule .*mod_mime.so\)/\1/' \
+		-e 's/^#\(LoadModule .*mod_dir.so\)/\1/' \
+		-e 's/^#\(LoadModule .*mod_authz_core.so\)/\1/' \
+		-e 's/^#\(LoadModule .*mod_proxy.so\)/\1/' \
+		-e 's/^#\(LoadModule .*mod_proxy_http.so\)/\1/' \
+		-e 's/^#\(LoadModule .*mod_ssl.so\)/\1/' \
+		-e 's/^#\(LoadModule .*mod_mpm_event.so\)/\1/' \
+		-e 's/\(LoadModule .*mod_mpm_worker.so\)/#\1/' \
+		-e 's/\(LoadModule .*mod_mpm_prefork.so\)/#\1/' \
```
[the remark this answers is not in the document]
Zoey2936:
```suggestion
                -e '/^Listen /d' \
                -e 's/User apache/User www-data/g' \
                -e 's/Group apache/Group www-data/g' \
                -e 's/^#\(LoadModule .*mod_rewrite.so\)/\1/' \
                -e 's/^#\(LoadModule .*mod_headers.so\)/\1/' \
                -e 's/^#\(LoadModule .*mod_env.so\)/\1/' \
                -e 's/^#\(LoadModule .*mod_mime.so\)/\1/' \
                -e 's/^#\(LoadModule .*mod_dir.so\)/\1/' \
                -e 's/^#\(LoadModule .*mod_authz_core.so\)/\1/' \
                -e 's/^#\(LoadModule .*mod_proxy.so\)/\1/' \
                -e 's/^#\(LoadModule .*mod_proxy_http.so\)/\1/' \
                -e 's/^#\(LoadModule .*mod_ssl.so\)/\1/' \
                -e 's/^#\(LoadModule .*mod_mpm_event.so\)/\1/' \
                -e 's/\(LoadModule .*mod_mpm_worker.so\)/#\1/' \
                -e 's/\(LoadModule .*mod_mpm_prefork.so\)/#\1/' \
```

Containers/mastercontainer/Dockerfile
```diff
@@ -54,36 +77,38 @@ RUN set -e && \
...
+                -e 's/^#\(LoadModule .*mod_env.so\)/\1/' \
+                -e 's/^#\(LoadModule .*mod_mime.so\)/\1/' \
+                -e 's/^#\(LoadModule .*mod_dir.so\)/\1/' \
+                -e 's/^#\(LoadModule .*mod_authz_core.so\)/\1/' \
+                -e 's/^#\(LoadModule .*mod_proxy.so\)/\1/' \
+                -e 's/^#\(LoadModule .*mod_proxy_http.so\)/\1/' \
+                -e 's/^#\(LoadModule .*mod_ssl.so\)/\1/' \
+                -e 's/^#\(LoadModule .*mod_mpm_event.so\)/\1/' \
+                -e 's/\(LoadModule .*mod_mpm_worker.so\)/#\1/' \
+                -e 's/\(LoadModule .*mod_mpm_prefork.so\)/#\1/' \
```
szaimen, line 104:
```suggestion
            -e '/^Listen /d' \
            -e 's/User apache/User www-data/g' \
            -e 's/Group apache/Group www-data/g' \
            -e 's/^#\(LoadModule .*mod_rewrite.so\)/\1/' \
            -e 's/^#\(LoadModule .*mod_headers.so\)/\1/' \
            -e 's/^#\(LoadModule .*mod_env.so\)/\1/' \
            -e 's/^#\(LoadModule .*mod_mime.so\)/\1/' \
            -e 's/^#\(LoadModule .*mod_dir.so\)/\1/' \
            -e 's/^#\(LoadModule .*mod_authz_core.so\)/\1/' \
            -e 's/^#\(LoadModule .*mod_proxy.so\)/\1/' \
            -e 's/^#\(LoadModule .*mod_proxy_http.so\)/\1/' \
            -e 's/^#\(LoadModule .*mod_ssl.so\)/\1/' \
            -e 's/^#\(LoadModule .*mod_mpm_event.so\)/\1/' \
            -e 's/\(LoadModule .*mod_mpm_worker.so\)/#\1/' \
            -e 's/\(LoadModule .*mod_mpm_prefork.so\)/#\1/' \
```

szaimen:
LGTM now but will try to test this locally first 👍

Containers/mastercontainer/Dockerfile
```diff
@@ -54,36 +78,41 @@ RUN set -e && \
...
+            -e 's/Group apache/Group www-data/g' \
+            -e 's/^#\(LoadModule .*mod_rewrite.so\)/\1/' \
+            -e 's/^#\(LoadModule .*mod_headers.so\)/\1/' \
+            -e 's/^#\(LoadModule .*mod_env.so\)/\1/' \
+            -e 's/^#\(LoadModule .*mod_mime.so\)/\1/' \
+            -e 's/^#\(LoadModule .*mod_dir.so\)/\1/' \
+            -e 's/^#\(LoadModule .*mod_authz_core.so\)/\1/' \
+            -e 's/^#\(LoadModule .*mod_proxy.so\)/\1/' \
+            -e 's/^#\(LoadModule .*mod_proxy_http.so\)/\1/' \
+            -e 's/^#\(LoadModule .*mod_proxy_fcgi.so\)/\1/' \
```
Zoey2936, line 102:
```suggestion
```
When installing the package `apache2-proxy` they are automatically activated

Containers/mastercontainer/Dockerfile
```diff
@@ -34,6 +34,7 @@ RUN set -ex; \
         bash \
         apache2 \
         apache2-proxy \
+        apache2-ssl \
```
Zoey2936, line 37: This will a file called `/etc/apache2/conf.d/ssl.conf` or something like this, which contains "Listen 443" and a virtual host on port 443, we should remove the virtualhost and the listen line

szaimen: removed the ssl.conf

Containers/mastercontainer/Dockerfile
szaimen: done

Containers/mastercontainer/mastercontainer.conf
```diff
@@ -1,6 +1,10 @@
 Listen 8000
 Listen 8080
 
+# Include ssl modules
+LoadModule ssl_module modules/mod_ssl.so
+LoadModule socache_shmcb_module modules/mod_socache_shmcb.so
```
Zoey2936, line 6: For what is this module (socache_shmcb) needed? And why do we include this here and not in the main httpd.conf?

szaimen: I could add it to the main httpd.conf, it was former in the ssl.conf that I removed. Will add it to the main httpd.conf

Zoey2936: Would be better that all modules get loaded in one conf

szaimen: jup,, done

szaimen:
@Zoey2936 does this look good to you?

Containers/mastercontainer/mastercontainer.conf
Zoey2936: > I could add it to the main httpd.conf, it was former in the ssl.conf that I removed. Will add it to the main httpd.conf

It seems that the socache_shmcb module is required for the ssl module

szaimen: let me add this then back

szaimen: done. pleae have another look

Zoey2936:
> @Zoey2936 does this look good to you now? I'll merge then :)

looks good

szaimen:
Thanks! :)

szaimen (approved):
LGTM

szaimen:
The change is crazy:

<details>
<summary>Click to show</summary>

image

</details>

szaimen:
For the apache container the same regarding security improvements. Also the size there is reduced from 88 MB to 55 MB

Zoey2936:
Wow, I didn't think the change would be so big

szaimen:
Yes, me neither 🚀

PR #1578, opened 2022-12-22
migrate borgbackup container to alpine

szaimen:
- [x] needs https://github.com/nextcloud/all-in-one/pull/1767

### DIFF
--- a/Containers/borgbackup/Dockerfile
+++ b/Containers/borgbackup/Dockerfile
@@ -1,17 +1,14 @@
-FROM debian:bullseye-20230109-slim
+FROM alpine:3.17.1
 
 RUN set -ex; \
     \
-    echo "deb http://deb.debian.org/debian bullseye-backports main" >> /etc/apt/sources.list; \
-    apt-get update; \
-    apt-get install -y --no-install-recommends borgbackup -t bullseye-backports; \
-    apt-get install -y --no-install-recommends \
+    apk add --no-cache \
+        bash \
+        borgbackup \
         rsync \
         fuse \
-        python3-llfuse \
-        jq \
-    ; \
-    rm -rf /var/lib/apt/lists/*
+        py3-llfuse \
+        jq
 
 VOLUME /root
 


Containers/borgbackup/Dockerfile
```diff
@@ -1,14 +1,13 @@
-FROM debian:bullseye-20221219-slim
+FROM alpine:3.17.0
 
 RUN set -ex; \
     \
-    echo "deb http://deb.debian.org/debian bullseye-backports main" >> /etc/apt/sources.list; \
-    apt-get update; \
-    apt-get install -y --no-install-recommends borgbackup -t bullseye-backports; \
-    apt-get install -y --no-install-recommends \
+    apk add --update --no-cache \
```
Zoey2936, line 5:
```suggestion
    apk add --no-cache \
```

Because of:
```
 apk add --help | grep update
  -U, --update-cache    Alias for '--cache-max-age 1'
```
Or does `--update` something, but the help doesn't show it?

Zoey2936 (changes requested):

Containers/borgbackup/Dockerfile
szaimen: > Or does `--update` something, but the help doesn't show it?

yes, it does :)

szaimen: https://stackoverflow.com/a/59052608

Containers/borgbackup/Dockerfile
```diff
@@ -1,14 +1,13 @@
...
+        bash \
+        borgbackup \
         rsync \
         fuse \
-        python3-llfuse \
+        py3-llfuse \
         jq \
     ; \
     rm -rf /var/lib/apt/lists/*
```
Zoey2936, line 13: needs to be removed

Containers/borgbackup/Dockerfile
```diff
@@ -1,14 +1,13 @@
-FROM debian:bullseye-20221219-slim
+FROM alpine:3.17.0
```
szaimen, line 1: I guess I would also like to wait for 3.17.1. Hope it will not take too long...

Containers/borgbackup/Dockerfile
Zoey2936: But why do we say?:
 - Don't create or use a cache (--no-cache)
and:
- Create a cache (--update-cache)

szaimen: ah, that is a good point. I guess will have to investigate...

Containers/borgbackup/Dockerfile
Zoey2936, line 1:
```suggestion
FROM alpine:3.17.1
```
3.17.1 is now available on docker hub

Containers/borgbackup/Dockerfile
szaimen: https://github.com/nextcloud/all-in-one/pull/1767

Containers/borgbackup/Dockerfile
szaimen: done

szaimen: done

Containers/borgbackup/Dockerfile
szaimen: done

Containers/borgbackup/Dockerfile
szaimen: done

Zoey2936 (approved):

szaimen:
Thanks Zoey! :)

szaimen:
It works 🎉🎉🎉🎉🎉

PR #1580, opened 2022-12-22
improve the apache reverse proxy documentation

### LINKED ISSUE #1314: Improve apache reverse proxy doc
Capture the whole request uri using a dedicated block and add that to the reverse proxy. Should be better than allowing specific characters and is also recommended in the B doxumentation

Zoey2936:
Close https://github.com/nextcloud/all-in-one/issues/1314

Got it from: https://stackoverflow.com/a/41388585 & https://httpd.apache.org/docs/2.4/mod/mod_rewrite.html & https://httpd.apache.org/docs/2.4/mod/mod_proxy.html#proxypass
In my test, it works

### DIFF
--- a/reverse-proxy.md
+++ b/reverse-proxy.md
@@ -46,10 +46,11 @@ Add this as a new Apache site config:
     RewriteEngine On
     ProxyPreserveHost On
     AllowEncodedSlashes NoDecode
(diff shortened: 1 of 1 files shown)


szaimen (approved):
Seems to work in my test as well. Thanks! :)

PR #1584, opened 2022-12-23
upgrade postgres to 15

szaimen:
Close part of https://github.com/nextcloud/all-in-one/discussions/1396

- [x] TODO: apply security changes to upgrade and restore: 
image
from https://docs.nextcloud.com/server/stable/admin_manual/configuration_database/linux_database_configuration.html#postgresql-database


Containers/postgresql/Dockerfile
```diff
@@ -1,5 +1,5 @@
 # From https://github.com/docker-library/postgres/blob/master/13/alpine/Dockerfile
-FROM postgres:14.6-alpine
+FROM postgres:15.1-alpine3.16
```
szaimen, line 2: and wait for alpine 3.17.1?

```suggestion
FROM postgres:15.1-alpine
```

Containers/postgresql/Dockerfile
```diff
@@ -1,5 +1,5 @@
 # From https://github.com/docker-library/postgres/blob/master/13/alpine/Dockerfile
```
Zoey2936, line 1:
```suggestion
# From https://github.com/docker-library/postgres/blob/master/15/alpine/Dockerfile
```

Containers/postgresql/Dockerfile
Zoey2936:
```suggestion
FROM 15.1-alpine3.17
```
maybe pin the alpine version?

szaimen: thought about this as well, however there was a time when minor posgres updates where not pushed with older alpine version anymore. That is why I removed that. Also it is pretty unlikely that alpine updates break this specific container.

Containers/postgresql/Dockerfile
szaimen: done

Containers/postgresql/Dockerfile
szaimen: removed that again.

Containers/postgresql/start.sh
```diff
@@ -110,6 +110,8 @@ if ( [ -f "$DATADIR/PG_VERSION" ] && [ "$PG_MAJOR" != "$(cat "$DATADIR/PG_VERSIO
         psql -v ON_ERROR_STOP=1 --username "$POSTGRES_USER" --dbname "$POSTGRES_DB" <<-EOSQL
             CREATE USER "$DB_OWNER" WITH PASSWORD '$POSTGRES_PASSWORD' CREATEDB;
             ALTER DATABASE "$POSTGRES_DB" OWNER TO "$DB_OWNER";
+            GRANT ALL PRIVILEGES ON DATABASE "$POSTGRES_DB" TO "$DB_OWNER";
+            GRANT ALL PRIVILEGES ON SCHEMA public TO "$DB_OWNER";
```
szaimen, line 114: same here

Zoey2936:
A Question, when Postgres starts it loads the database dump, right?
Since I think Postgres support upgrading major version only through exporting a dump and then importing it again. (But I could be wrong)

szaimen:
> A Question, when Postgres starts it loads the database dump, right?

Yes, ahen the major version between files ans image dont match it starts the import from the database dump on startup 👍

Containers/postgresql/init-user-db.sh
```diff
@@ -6,6 +6,8 @@ touch "$DUMP_DIR/initialization.failed"
 psql -v ON_ERROR_STOP=1 --username "$POSTGRES_USER" --dbname "$POSTGRES_DB" <<-EOSQL
 	CREATE USER "oc_$POSTGRES_USER" WITH PASSWORD '$POSTGRES_PASSWORD' CREATEDB;
 	ALTER DATABASE "$POSTGRES_DB" OWNER TO "oc_$POSTGRES_USER";
+	GRANT ALL PRIVILEGES ON DATABASE "$POSTGRES_DB" TO "oc_$POSTGRES_USER";
+	GRANT ALL PRIVILEGES ON SCHEMA public TO "oc_$POSTGRES_USER";
```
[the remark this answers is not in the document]
Zoey2936: I think it should work, but testing will show it

Zoey2936 (approved):

Containers/postgresql/init-user-db.sh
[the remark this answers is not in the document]
szaimen: True. Thanks for the approval! I'll merge it when 4.2.0 stable is released.

szaimen:
It works 🎉🎉🎉🎉

PR #1585, opened 2022-12-23
update redis to v7

szaimen:

### DIFF
--- a/Containers/redis/Dockerfile
+++ b/Containers/redis/Dockerfile
@@ -1,5 +1,5 @@
-# From https://github.com/docker-library/redis/blob/master/6.2/alpine/Dockerfile
-FROM redis:6.2.8-alpine
+# From https://github.com/docker-library/redis/blob/master/7.0/alpine/Dockerfile
+FROM redis:7.0.7-alpine
 
 RUN apk add --update --no-cache openssl bash
 


Containers/redis/Dockerfile
```diff
@@ -1,5 +1,5 @@
 # From https://github.com/docker-library/redis/blob/master/6.2/alpine/Dockerfile
-FROM redis:6.2.8-alpine
+FROM redis:7.0.7-alpine
```
szaimen, line 2: wait for alpine 3.17?

Containers/redis/Dockerfile
```diff
@@ -1,5 +1,5 @@
 # From https://github.com/docker-library/redis/blob/master/6.2/alpine/Dockerfile
```
Zoey2936, line 1:
```suggestion
# From https://github.com/docker-library/redis/blob/master/7.0/alpine/Dockerfile
```

Containers/redis/Dockerfile
Zoey2936: Maybe also pin the alpine version?
```suggestion
FROM 7.0.7-alpine3.17
```

szaimen: thought about this as well, however there was a time when minor posgres updates where not pushed with older alpine version anymore. That is why I removed that. Also it is pretty unlikely that alpine updates break this specific container.

Zoey2936 (approved):

PR #1590, opened 2022-12-23
manual-install: update readme to use default docker environment file

ManOki:
By using file `.env` directly, the parameter `--env-file my.conf` can be omitted.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,3 +6,4 @@
 /manual-install/*.conf
 !/manual-install/sample.conf
 /manual-install/docker-compose.yml
+/manual-install/.env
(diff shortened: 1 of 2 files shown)


szaimen (changes requested):
Fine by me but you should probably add the file to gitignore: https://github.com/nextcloud/all-in-one/blob/main/.gitignore

szaimen (commented):
Thanks! Maxbe it would also be good to add a note for people that are already using it to continue using their my.conf or rename their file to .env?

ManOki:
> Thanks! Maxbe it would also be good to add a note for people that are already using it to continue using their my.conf or rename their file to .env?

What would be the appropriate place for this?

szaimen:
> What would be the appropriate place for this?

I think the readme of manual-install would be a good place :)

ManOki:
I added an instruction to the _How to update?_ section.

szaimen (approved):
LGTM, thanks! :)

PR #1586, opened 2022-12-23
manual-install: add docker profiles for optional services

ManOki:
see #1576

PR #1591, opened 2022-12-23
manual-install: add IMAGE_TAG to omit separate latest-arm64.yml

ManOki:
Add an additional parameter IMAGE_TAG to configure the docker-compose.yml image attributes, i.e. the architecture amd64 or arm64.

This PR is especially useful if PR #1586 is also accepted and optional services can be omitted if they are not available on certain architectures like clamav on arm64.


szaimen (commented):
Hm... Not sure about this as clamav is not available on arm64...

szaimen:
> Not sure about this as clamav is not available on arm64...

Due to this I would honestly rather keep it sepeate. WDYT @ManOki ? :)

ManOki:
@szaimen I think this is a good PR :) otherwise I would have not created it.

But of course without #1586, users will get an error for clamav on arm64 and are not able to start their containers so the other PR is a hard dependency. 
Another advantage is that you may use other image tags as well like `develop` or `beta`
https://hub.docker.com/r/nextcloud/aio-nextcloud/tags
https://hub.docker.com/r/nextcloud/aio-collabora/tags

szaimen:
All right! Then please add a note to the readme stating that clamav is not supported on arm64. Also we will need to adjust the sctipt for this as well since it will recreate the latest-arm64 file otherwise.

ManOki:
There is already a [hint for clamav](https://github.com/ManOki/nextcloud-all-in-one/blob/docker_image_tag/manual-install/readme.md?plain=1#L24) and an updated version of [update-yaml.sh](https://github.com/ManOki/nextcloud-all-in-one/blob/docker_image_tag/manual-install/update-yaml.sh#L126). I did not test the complete script as I don't know/have the runtime environment.

szaimen:
This needs a rebase now due to conflicts...

szaimen:
Do you need help with the rebase? :)

manual-install/update-yaml.sh
```diff
@@ -128,12 +130,6 @@ networks:
...
 
 cat containers.yml > latest.yml
-sed -i '/image:/s/$/:latest/' latest.yml
-
-cat containers.yml > latest-arm64.yml
-sed -i '/image:/s/$/:latest-arm64/' latest-arm64.yml
-sed -i '/  nextcloud-aio-clamav:/,/^$/d' latest-arm64.yml
-sed -i '/nextcloud[-_]aio[-_]clamav/d' latest-arm64.yml
-sed -i '/CLAMAV_ENABLED/d' latest-arm64.yml
+sed -i '/image:/s/$/:${IMAGE_TAG}/' latest.yml
```
szaimen, line 133:
```suggestion
sed -i '/image:/s/$/:\${IMAGE_TAG}/' latest.yml
```
You can check if this fixes SC

manual-install/sample.conf
```diff
@@ -1,8 +1,9 @@
+IMAGE_TAG=latest          # Version of docker images, should be latest or latest-arm. Note: latest-arm64 has no clamav support
```
szaimen, line 1:
```suggestion
IMAGE_TAG=latest          # Version of docker images, should be latest or latest-arm64. Note: latest-arm64 has no clamav support
```

manual-install/update-yaml.sh
```diff
@@ -128,12 +130,6 @@ networks:
...
 
 cat containers.yml > latest.yml
-sed -i '/image:/s/$/:latest/' latest.yml
-
-cat containers.yml > latest-arm64.yml
-sed -i '/image:/s/$/:latest-arm64/' latest-arm64.yml
-sed -i '/  nextcloud-aio-clamav:/,/^$/d' latest-arm64.yml
-sed -i '/nextcloud[-_]aio[-_]clamav/d' latest-arm64.yml
-sed -i '/CLAMAV_ENABLED/d' latest-arm64.yml
+sed -i '/image:/s/$/:\${IMAGE_TAG}/' latest.yml
```
szaimen, line 133: Another attempt:
```suggestion
sed -i "/image:/s/$/:\${IMAGE_TAG}/" latest.yml
```

manual-install/update-yaml.sh
```diff
@@ -57,6 +58,7 @@ do
 done
 
 sed -i 's|_ENABLED=|_ENABLED=no          # Setting this to "yes" enables the option in Nextcloud automatically.|' sample.conf
+sed -i 's|CLAMAV_ENABLED=no|_ENABLED=no          # Setting this to "yes" enables the option in Nextcloud automatically. Note: latest-arm64 has no clamav support|' sample.conf
```
szaimen, line 61:
```suggestion
sed -i 's|CLAMAV_ENABLED=no|CLAMAV_ENABLED=no          # Setting this to "yes" enables the option in Nextcloud automatically. Note: latest-arm64 has no clamav support|' sample.conf
```

szaimen (approved):
Thanks, LGTM! :)

ManOki:
Thank you for your detailed review and fixes.

szaimen:
No problem! Thank you for your contributions! :)

PR #1597, opened 2022-12-25
allow to add container to host network

szaimen:
For https://github.com/nextcloud/all-in-one/discussions/1581

- [x] Needs https://github.com/nextcloud/all-in-one/pull/1595

This is best reviewed like this: https://github.com/nextcloud/all-in-one/pull/1597/files?diff=unified&w=1

### DIFF
--- a/php/src/Docker/DockerActionManager.php
+++ b/php/src/Docker/DockerActionManager.php
@@ -124,8 +124,9 @@ class DockerActionManager
         }
 
         $containerName = $container->GetIdentifier();
-        if ($container->GetInternalPort() !== "") {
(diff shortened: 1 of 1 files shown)


php/src/Docker/DockerActionManager.php
```diff
@@ -655,12 +665,14 @@ private function ConnectContainerIdToNetwork(string $id) : void
 
     public function ConnectMasterContainerToNetwork() : void
     {
-        $this->ConnectContainerIdToNetwork('nextcloud-aio-mastercontainer');
+        $container = $this->containerDefinitionFetcher->GetContainerById('nextcloud-aio-mastercontainer');
```
szaimen, line 668: this does not workk

Issue #1611, opened 2022-12-28
Temporary unhealthy containers during auto backups


frederikb96:
### How to use GitHub

* Please use the 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to show that you are affected by the same issue.
* Please don't comment if you have no relevant information to add. It's just extra noise for everyone subscribed to this issue.
* Subscribe to receive notifications on status change and new comments. 

### Steps to reproduce
1. Make some big changes to your files (such that the backup runs for at least some minutes)
1. Via the Nextcloud AIO interface, schedule an auto backup for in 2 minutes
2. Open portainer or the command line, and observer the containers
3. When the backup starts, all containers are stopped, the backup container starts and all other containers start

### Expected behavior 
All containers directly go into the healthy state

### Actual behavior 
The Nextcloud container and the Apache container goes into the unhealthy state and afterwards go into the healthy state

### My guess
If I manually do a backup, I manually have to start the containers again, and every container goes directly into the healthy state.
I guess, since the containers are already somehow started in the auto backup process with the backup container and the real start is just postponed, they run into some timeout and go into the unhealthy state. This is annoying, since I get notifications from netdata, that there are unhealthy containers, every night. I would rather not adjust the netdata settings, since I really want to be warned if there are unhealthy containers.

I guess if the backup is really small, the containers start fast enough, such that no unhealthy state is triggered.

### Host OS 
Ubuntu 22.04 server

#### Nextcloud AIO version 
v4.0.1

#### Current channel 
Beta

Zoey2936:
The Container reports unhealthy if the services inside the container don't bind their ports, sometimes the containers need to wait for other containers to start and then bind their ports and because of this theoretically the containers are unhealthy. I don't see an option to differentiate between a container that takes long to start and a really unhealthy container.
The only Thing I see we could do is to increase the start period (https://docs.docker.com/engine/reference/builder/#healthcheck)

szaimen:
> I would rather not adjust the netdata settings, since I really want to be warned if there are unhealthy containers.

I would rather suggest checking if one of the containers is restarting which is also what we do in AIO and the better approach, imho.

> I don't see an option to differentiate between a container that takes long to start and a really unhealthy container.
> The only Thing I see we could do is to increase the start period (https://docs.docker.com/engine/reference/builder/#healthcheck)

[...]

PR #1615, opened 2022-12-28
improve healthcheck to only exit 1 if it is started

szaimen:
Close https://github.com/nextcloud/all-in-one/issues/1611

Disclaimer: this is not tested yet


szaimen:
@Zoey2936 WDYT, would this work?

Zoey2936:
I think it could work

Zoey2936 (approved):

PR #1645, opened 2023-01-02
add types for missing ones and bring psalm-baseline down to zero

szaimen:


php/public/index.php
```diff
@@ -12,6 +12,8 @@
 use Slim\Factory\AppFactory;
 use Slim\Views\Twig;
 use Slim\Views\TwigMiddleware;
+use Psr\Http\Message\ResponseInterface as Response;
+use Psr\Http\Message\RequestInterface as Request;
```
szaimen, line 16: not completely sure if `ServerRequestInterface` would not be the better one and/or only for authentication handling... Would need to be tested.

szaimen: psalm added the solution

php/public/index.php
```diff
@@ -140,7 +142,7 @@
 });
 
 // Auth Redirector
-$app->get('/', function (\Psr\Http\Message\RequestInterface $request, \Psr\Http\Message\ResponseInterface $response, $args) use ($container) {
+$app->get('/', function (\Psr\Http\Message\RequestInterface $request, Response $response, mixed $args) use ($container) {
```
come-nc, line 145:
```suggestion
$app->get('/', function (Request $request, Response $response, mixed $args) use ($container) {
```

consistency

come-nc (changes requested):
I think `$args` should be `array`: https://www.slimframework.com/docs/v4/objects/routing.html#route-callbacks

php/public/index.php
szaimen: No, RequestInterface should be correct IIRC... (It is not a ServerRequest but a Client-Side request IIRC)...

szaimen:
> I think `$args` should be `array`: https://www.slimframework.com/docs/v4/objects/routing.html#route-callbacks

Thanks for figuring this out! Fixed :)

come-nc (approved):

php/public/index.php
come-nc: Name is misleading then, maybe import both as Request and ServerRequest to avoid the confusion.

szaimen:
Thanks a lot guys! :)

php/public/index.php
szaimen: sorry, saw this too late

PR #1657, opened 2023-01-02
fix healthcheck for nextcloud and database

szaimen:
Follow-up of https://github.com/nextcloud/all-in-one/pull/1615

### DIFF
--- a/Containers/nextcloud/Dockerfile
+++ b/Containers/nextcloud/Dockerfile
@@ -250,7 +250,8 @@ RUN set -ex; \
     chmod +x /cron.sh && \
     chmod +x /notify.sh && \
     chmod +x /notify-all.sh && \
-    chmod +x /activate-collabora.sh
+    chmod +x /activate-collabora.sh && \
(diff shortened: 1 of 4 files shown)


Containers/nextcloud/Dockerfile
```diff
@@ -250,7 +250,8 @@ RUN set -ex; \
     chmod +x /cron.sh && \
     chmod +x /notify.sh && \
     chmod +x /notify-all.sh && \
-    chmod +x /activate-collabora.sh
+    chmod +x /activate-collabora.sh && \
+    chmod +x healthcheck.sh
```
Zoey2936, line 254: why not also?
```suggestion
    chmod +x /healthcheck.sh
```
And does the file even get copied?

szaimen: > why not also?

Typo

> And does the file even get copied?

Yes

Zoey2936 (approved):
If the file gets copied, it should work

PR #1658, opened 2023-01-02
add twofactor_totp back as default enabled app

szaimen:
Since https://github.com/nextcloud/server/pull/35944 was not accepted, we should add it back


Zoey2936 (approved):

szaimen:
Thanks Zoey! :)

PR #1672, opened 2023-01-04
Add Parameter NC_WEBROOT to configure subdirectory of nextcloud

ManOki:
`NC_WEBROOT` values can be empty string, '/' (default), or any path like '/nextcloud'.

Required docker entrypoint scripts build+export an additional variable `NC_WEBROOT_P`, which is added directly after particular `host:port` combinations ('/' is converted to '' so `http://$host:$post$NC_WEBROOT_P/push` does not become `http://host:port//push`)

`NC_WEBROOT` is added to all possible paths of nextcloud, push, talk and onlyoffice but not collabora as there is no reasonable way to update their paths.

For mastercontainer/php configuration there is currently only the default value '/', but for manual-install this PR can be used directly.

### DIFF
--- a/Containers/apache/Caddyfile
+++ b/Containers/apache/Caddyfile
@@ -8,17 +8,25 @@
 
 {$PROTOCOL}://{$NC_DOMAIN}:{$APACHE_PORT} {
 
+    # well-known
+    redir /.well-known/carddav {$NC_WEBROOT_P}/remote.php/dav
+    redir /.well-known/caldav {$NC_WEBROOT_P}/remote.php/dav
+    redir /.well-known/webdav {$NC_WEBROOT_P}/remote.php/dav
+    redir /.well-known/webfinger {$NC_WEBROOT_P}/index.php/.well-known/webfinger
+    redir /.well-known/nodeinfo {$NC_WEBROOT_P}/index.php/.well-known/nodeinfo
+
+
     # Notify Push
-    route /push/* {
-        uri strip_prefix /push
+    route {$NC_WEBROOT_P}/push/* {
+        uri strip_prefix {$NC_WEBROOT_P}/push
         reverse_proxy {$NEXTCLOUD_HOST}:7867 {
             # trusted_proxies placeholder
         }
     }
 
     # Talk
-    route /standalone-signaling/* {
-        uri strip_prefix /standalone-signaling
+    route {$NC_WEBROOT_P}/standalone-signaling/* {
+        uri strip_prefix {$NC_WEBROOT_P}/standalone-signaling
         reverse_proxy {$TALK_HOST}:8081 {
             # trusted_proxies placeholder
         }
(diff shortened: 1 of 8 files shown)


Zoey2936:
Even if I think that this function would cause many more problems and using a subdomain is a much cleaner and better way,
I want to tell you that currently for example the domain check wouldn't work, see [this line](https://github.com/nextcloud/all-in-one/blob/16054318272f241517baf0f79e12b59537a87c9b/php/src/Data/ConfigurationManager.php#L292) and I think there are more Things to adjust, which wouldn't work currently with this implementation. Also, I think there should be adjusted versions of all reverse proxy configurations.

szaimen:
Hi @ManOki, thanks a lot for your contribution!

However as stated in the readme of this project, I don't want to add support for this due to the following arguments:

- As stated by Zoey is using a subdomain or an additional domain the much cleaner approach. It is also very easy to add one (imo) and more secure since assets usually cannot be shared between (sub-)domains in contrast to subdirectories.
- I am not able to maintain this feature since I will not test it because it would add yet another thing to test for me and would thus make testing for me more complicated. 
- It would also take away time for maintaining the feature (by having to answer questions and/or issues or having to improve details or documentation) that I would like to invest into other parts of the project
- The feature is not complete in its current implementation as stated by Zoey above

I hope you understand the reasons and hope you have a nice evening!

PS: I don't want to close the PR as I feel like this would be rude. So feel free to do it youself.

ManOki:
I agree, a separate domain is a cleaner way. This is for users, 

1. who are not able to get/have a subdomain or 
2. who already have a nextcloud setup with a subdirectory and want to switch to nextcloud-aio.

> the domain check wouldn't work, see [this line](https://github.com/nextcloud/all-in-one/blob/16054318272f241517baf0f79e12b59537a87c9b/php/src/Data/ConfigurationManager.php#L292)

The current state of the PR only _supports_ manual-install (change NC_WEBROOT in docker environment file), the only code for mastercontainer/php set the default value '/' and therefore do not change the other functionality. Is the acceptance for this PR higher if the rest of the code also updated? From my POV, _just_ manual-install is a good trade-off because those users are much more into docker, configuration, proxies and so on.

> Also, I think there should be adjusted versions of all reverse proxy configurations.

@Zoey2936 Do you mean the readme?

Zoey2936:
> 1. who are not able to get/have a subdomain or

If they own a domain, they could easily get an subdomain, even when using a DDNS service, it is possible to get multiple domains

> 2. who already have a nextcloud setup with a subdirectory and want to switch to nextcloud-aio.

They could use an subdomain and create an redirect

ManOki:
> It would also take away time for maintaining the feature (by having to answer questions and/or issues or having to improve details or documentation) that I would like to invest into other parts of the project

@szaimen So adding (more) documentation/readme would increase the acceptance of this PR?

> The feature is not complete in its current implementation as stated by Zoey above

The feature is complete in terms of manual-install. There are also other features that are only available in manual-install, so I assumed that this would not be a problem for PR.

ManOki:
Fun fact: There is already support to [OVERWRITEWEBROOT](https://github.com/nextcloud/all-in-one/blob/main/Containers/nextcloud/config/reverse-proxy.config.php#L14) ;)

PR #1150, opened 2022-09-14
allow to disable seccomp for collabora

szaimen:
Close https://github.com/nextcloud/all-in-one/issues/1149

Signed-off-by: szaimen <szaimen@e.mail.de>

PR #1692, opened 2023-01-07
Add Parameter FULLTEXTSEARCH_INDEX_ONCE to prevent index every startup

ManOki:
In case of large setups with a lot of data, there will (imho) always be some errors while indexing fulltextsearch. This results into a complete laborious re-index every time the container is start up (without nextcloud available for the users).

This PR adds the parameter FULLTEXTSEARCH_INDEX_ONCE that create the marker file `fts-index.done` no matter if `occ fulltextsearch:index` returns an error or not.

Note: There is currently no code for mastercontainer/php except setting the default/old behaviour value 'no'.


Zoey2936:
> Note: There is currently no code for mastercontainer/php except setting the default/old behaviour value 'no'.

Without reviewing the changed yet or thinking about the sense of this feature, from my opinion there shouldn't be features that are only available inside manual install

ManOki:
> Without reviewing the changed yet or thinking about the sense of this feature, from my opinion there shouldn't be features that are only available inside manual install

From my point of view, this is not just a manual-install feature but I'm not able to add the needed code for mastercontainer/php. Shall I delete the PR and instead open an issue?

Zoey2936:
> > Without reviewing the changed yet or thinking about the sense of this feature, from my opinion there shouldn't be features that are only available inside manual install
> 
> From my point of view, this is not just a manual-install feature but I'm not able to add the needed code for mastercontainer/php. Shall I delete the PR and instead open an issue?

You could have a look here: https://github.com/nextcloud/all-in-one/pull/1150/files how you can integrate this into the master container

szaimen:
Hi, before going deeper into this, please first open a discussion about this. I have a few words to say...

ManOki:
See #1694

PR #1696, opened 2023-01-07
Add hint about Cloudflare Rocket Loader and Rename Cloudflare Argo Tu…

szaimen:
…nnel to Cloudflare Tunnel

Close https://github.com/nextcloud/all-in-one/discussions/1677


szaimen:
@itsMaxio does this look good to you? :)

Zoey2936:
I think the rocket loader part should be also generally mentioned somewhere, since AIO can also be running behind Cloudflare without using tunnels

itsMaxio (approved):
Yeah looks good thanks for the quick reply :)

szaimen:
> I think the rocket loader part should be also generally mentioned somewhere, since AIO can also be running behind Cloudflare without using tunnels

Interesting! @Zoey2936 , can you create a new issue for that or maybe create a PR that mentions this directly? Thanks! :)

szaimen:
And thankks for the approval @itsMaxio ! :)

PR #1716, opened 2023-01-10
add note to resolver in nginx reverse proxy docs

Zoey2936:

### DIFF
--- a/reverse-proxy.md
+++ b/reverse-proxy.md
@@ -257,8 +257,8 @@ server {
     server_name <your-nc-domain>;
 
     location / {
-        resolver localhost;
-        proxy_pass http://localhost:11000$request_uri;
+        resolver localhost; # Note: you need to set a valid dns resolver here or use 127.0.0.1 / [::1] instead of localhost in the line below. See https://stackoverflow.com/a/49642310 for a better explanation
+        proxy_pass http://localhost:11000$request_uri; # Note: you need to change localhost to 127.0.0.1 / [::1], if you don't use a valid dns resolver in the line above
 
         proxy_set_header Host $host;
         proxy_set_header X-Real-IP $remote_addr;
--- a/reverse-proxy.md
+++ b/reverse-proxy.md
@@ -258,7 +258,7 @@ server {
 
     location / {
         resolver localhost; # Note: you need to set a valid dns resolver here or use 127.0.0.1 / [::1] instead of localhost in the line below. See https://stackoverflow.com/a/49642310 for a better explanation
-        proxy_pass http://localhost:11000$request_uri; # Note: you need to change localhost to 127.0.0.1 / [::1], if you don't use a valid dns resolver in the line above
+        proxy_pass http://localhost:11000$request_uri; # Note: you need to change localhost to 127.0.0.1 or [::1], if you don't use a valid dns resolver in the line above
 
         proxy_set_header Host $host;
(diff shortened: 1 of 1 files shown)


reverse-proxy.md
```diff
@@ -257,7 +257,7 @@ server {
     server_name <your-nc-domain>;
 
     location / {
-        resolver localhost;
+        resolver localhost; # Note: if you don't run a dns resolver on localhost:53, you should use another resolver
```
szaimen, line 260: Ah thar explains things!

szaimen:
```suggestion
        resolver localhost; # Note: if you don't run a dns resolver on localhost:53, you should use another resolver. Otherwise you need to use 127.0.0.1 in the line below this one instead of localhost
```

szaimen: Or maybe add the note into the line directly? That if not resolver is defined you need to use 127.0.0.1?

Zoey2936:
```suggestion
        resolver localhost; # Note: you need to set a valid dns resolver here or use 127.0.0.1 / [::1] instead of localhost in the line below
```

szaimen: Yeah looks good, or maybe add the advice in the line below?

reverse-proxy.md
```diff
@@ -257,7 +257,7 @@ server {
...
         proxy_pass http://localhost:11000$request_uri;
```
Zoey2936, line 261:
```suggestion
        proxy_pass http://localhost:11000$request_uri; # Node: you need to change localhost to 127.0.0.1 / [::1], if you don't use a valid dns resolver in the line above
```

szaimen: Yeah looks good but should be `Note` instead of `Node`

reverse-proxy.md
szaimen:
```suggestion
        resolver localhost; # Note: if you don't run a dns resolver on localhost:53, you should use another resolver. See https://stackoverflow.com/a/49642310 for a better explanation
```

reverse-proxy.md
szaimen: Also would you need to add two lines if you use both ipv4 and ipv6?

Zoey2936: > Also would you need to add two lines if you use both ipv4 and ipv6?

No, but if a system runs ipv6 only it needs [::1] and ipv4 only, it needs 127.0.0.1, if both are enabled it doesn't matter what is used

reverse-proxy.md
```diff
@@ -257,8 +257,8 @@ server {
     server_name <your-nc-domain>;
 
     location / {
-        resolver localhost;
-        proxy_pass http://localhost:11000$request_uri;
+        resolver localhost; # Note: you need to set a valid dns resolver here or use 127.0.0.1 / [::1] instead of localhost in the line below. See https://stackoverflow.com/a/49642310 for a better explanation
+        proxy_pass http://localhost:11000$request_uri; # Note: you need to change localhost to 127.0.0.1 / [::1], if you don't use a valid dns resolver in the line above
```
szaimen, line 261:
```suggestion
        proxy_pass http://localhost:11000$request_uri; # Note: you need to change localhost to 127.0.0.1 or [::1], if you don't use a valid dns resolver in the line above
```

szaimen (approved):
LGTM

PR #1719, opened 2023-01-11
Fix the middleware name in the Traefik doc

hunhejj:
The naming of the chain middleware in the Traefik example is inconsistent.


szaimen (approved):
LGTM, thanks! :)

PR #1732, opened 2023-01-12
redis: add hint to wiki if overcommit disabled

ManOki:
You may see a warning by redis like `WARNING Memory overcommit must be enabled! Without it, a background save or replication may fail under low memory condition.`

In this case you have to enable vm.overcommit on your host machine (cannot be done by docker/redis), so this PR links to a discussion with commands and background information if vm.overcommit is disabled.

### DIFF
--- a/Containers/redis/start.sh
+++ b/Containers/redis/start.sh
@@ -1,7 +1,7 @@
 #!/bin/bash
 
 # Show wiki if vm.overcommit is disabled
-if [[ $(sysctl -n vm.overcommit_memory)  != "1" ]]; then
+if [ "$(sysctl -n vm.overcommit_memory)" != "1" ]; then
     echo Memory overcommit is disabled but necessary for safe operation
     echo See https://github.com/nextcloud/all-in-one/discussions/1731 how to enable overcommit
 fi
--- a/Containers/redis/start.sh
+++ b/Containers/redis/start.sh
@@ -3,7 +3,7 @@
 # Show wiki if vm.overcommit is disabled
 if [ "$(sysctl -n vm.overcommit_memory)" != "1" ]; then
     echo "Memory overcommit is disabled but necessary for safe operation"
-    echo See https://github.com/nextcloud/all-in-one/discussions/1731 how to enable overcommit
+    echo "See https://github.com/nextcloud/all-in-one/discussions/1731 how to enable overcommit"
 fi
 
 # Run redis with a password if provided
--- a/Containers/redis/start.sh
+++ b/Containers/redis/start.sh
@@ -2,7 +2,7 @@
 
 # Show wiki if vm.overcommit is disabled
 if [ "$(sysctl -n vm.overcommit_memory)" != "1" ]; then
-    echo Memory overcommit is disabled but necessary for safe operation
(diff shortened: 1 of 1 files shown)


Containers/redis/start.sh
```diff
@@ -1,5 +1,11 @@
 #!/bin/bash
 
+# Show wiki if vm.overcommit is disabled
+if [[ $(sysctl -n vm.overcommit_memory)  != "1" ]]; then
```
Zoey2936, line 4:
```suggestion
if [ "$(sysctl -n vm.overcommit_memory)" != "1" ]; then
```

Containers/redis/start.sh
```diff
@@ -1,5 +1,11 @@
...
+    echo Memory overcommit is disabled but necessary for safe operation
```
Zoey2936, line 5:
```suggestion
    echo "Memory overcommit is disabled but necessary for safe operation"
```

Containers/redis/start.sh
```diff
@@ -1,5 +1,11 @@
...
+    echo See https://github.com/nextcloud/all-in-one/discussions/1731 how to enable overcommit
```
Zoey2936, line 6:
```suggestion
    echo "See https://github.com/nextcloud/all-in-one/discussions/1731 how to enable overcommit"
```

Zoey2936 (changes requested):

Zoey2936:
I don't know, but maybe this should also be mentioned also somewhere outside the Redis container logs
And I hope this is no problem for you, but I've edited your discussion comment a bit

ManOki:
> maybe this should also be mentioned also somewhere outside the Redis container logs

Yes, I agree. But I do not know exactly where is the best place. Maybe you can add a hint in mastercontainer (while setup). As you have [seen ](https://github.com/nextcloud/all-in-one/pull/1732/files#diff-431d00898671d8860c780d0f07baab8600a4ea0cd89d9448b218658b475b520eR4) docker containers are able to read the setting of the host system via `sysctl -n vm.overcommit_memory`.

> I've edited your discussion comment a bit

Feel free, you are much more advanced in AIO than me ;)

szaimen:
>> maybe this should also be mentioned also somewhere outside the Redis container logs
>
> Yes, I agree. But I do not know exactly where is the best place. Maybe you can add a hint in mastercontainer (while setup). 

Since redis is only used as cache, I don't think that it is important enough to mention it in the mastercontainer. Only showing it in the redis container is fine, imho.

szaimen (approved):
LGTM. Thank you Zoey and @ManOki! :)

szaimen:
Btw, @ManOki, I just added you as contributor to this repo since you've created quite a few PRs already and I think it would make things easier when you have write permission on the repo :)

ManOki:
> I just invited you as contributor to this repo since you've created quite a few PRs already and I think it would make things easier when you have write permission on the repo :)

Thank you! And thanks for approval.

szaimen:
This is now released with v4.2.0 Beta. Testing and feedback is welcome! See https://github.com/nextcloud/all-in-one#how-to-switch-the-channel

PR #1733, opened 2023-01-12
Fix the Traefik example

hunhejj:
The Traefik example partly used deprecated properties, partly was incorrect. This MR aims to provide a working configuration.


szaimen (approved):
LGTM, thanks! :)

PR #1741, opened 2023-01-13
Rewrite Traefik examples using YAML

hunhejj:
Since [all examples](https://doc.traefik.io/traefik/providers/file/) are shown as YAML by default on the Traefik documentation page and I personally find YAML way more readable than TOML, I propose to rewrite our examples using YAML. 

Note that I used the oppurtunity to use more descriptive and sometimes compacter names.

### DIFF
--- a/reverse-proxy.md
+++ b/reverse-proxy.md
@@ -356,51 +356,61 @@ Of course you need to modify `<your-nc-domain>` to the domain on which you want 
 
 **Disclaimer:** It might be possible that the config below is not working 100% correctly, yet. Improvements to it are very welcome!
 
-1. Add a `nextcloud.toml` to the Treafik rules folder with the following content:
+1. Add a `nextcloud.yml` to the Treafik rules folder with the following content
 
-    ```toml
-    [http.routers]
-        [http.routers.nc-rtr]
-            entryPoints = ["https"]
-            rule = "Host(<your-nc-domain>)"
-            service = "nc-svc"
-            middlewares = ["chain-nc"]
-            [http.routers.nc-rtr.tls]
-                certresolver = "le"
-
-    [http.services]
-        [http.services.nc-svc]
-            [http.services.nc-svc.loadBalancer]
-                passHostHeader = true
-                [[http.services.nc-svc.loadBalancer.servers]]
-                    url = "http://localhost:11000"
+    ```yml
+    http:
+        routers:
+            nextcloud:
+                rule: "Host(<your-nextcloud-domain>)"
+                entrypoints:
+                    - "https"
+                service: nextcloud
+                middlewares:
+                    - nextcloud-chain
+                tls:
+                    certresolver: "le"
+        services:
+            nextcloud:
+                loadBalancer:
+                    servers:
+                        - url: "http://localhost:11000"
     ```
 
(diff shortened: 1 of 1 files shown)


szaimen:
@hunhejj thanks a lot for the PR! However I think offering both would be something good. WDYT?

Maybe we could simply add two sections: `Traefik 2 (YAML)` and `Traefik 2 (TOML)`?

hunhejj:
That would mean you need to maintain both of them and chances are higher that they get out of sync @szaimen.

I think anyone can "convert" YAML into TOML by checking the examples on the Traefik site, so I think one or the other should be enough.

Btw. note that AFAIK soon Traefik 3 will be released too.

szaimen:
In my opinion it would be perfect to offer both versions. As soon as Treafik v3 gets released, we can re-think this. Would that be okay for you?

hunhejj:
What is the benefit of offering both versions? Even if someone defined his other rules in TOML, Traefik would allow him to define these rules in YAML (or the other way round).

I guess your goal is to make the situation of starters as easy as possible. But IMO if you are overwhelmed with choices you just get confused.

When you start using Traefik, most probably you will start defining everything with docker labels as that is the suggested and easiest way. You only need configuration files in more advanced situations. So as a beginner most probably when you arrive at this readme, you did not write any dynamic configuration at all for Traefik and you already feel lost anyways. How should you decide which of the two examples you should pick? You don't know anything about YAML vs TOML yet. You are then facing a choice instead of KISS which is against the principles of this awesome project IMO :wink:

szaimen:
Okay, you have a point here! Then maybe lets at least add a note how to translate the files into toml if you should be using that?

szaimen:
And btw, do you know if there is a guide out there where is described where to put the files on a usual traefik installation running in a docker container? I think that would probably be helpful for many...

hunhejj:
> And btw, do you know if there is a guide out there where is described where to put the files on a usual traefik installation running in a docker container? I think that would probably be helpful for many...

Sure, I will enrich the setup process in a dedicated MR.

szaimen:
> Sure, I will enrich the setup description in a dedicated MR.

Thanks a lot!

As for the toml topic, ciuld you add a hint how to convert it to toml if a person shoild be using that? Apart from does the PR look good to me 👍

hunhejj:
WDYM @szaimen? Should I link a random online YAML to TOML generator that I find?

szaimen:
> WDYM @szaimen? Should I link a random online YAML to TOML generator that I find?

Yeah fhat would work or simply add a hint that the person can use a yaml to toml converter if they need the toml version of it...

hunhejj:
Sounds good, I added the hint. Feel free to reformulate it if you prefer @szaimen .

szaimen (approved):
LGTM, thanks a lot! :)

szaimen:
Btw, @hunhejj , I just invited you as contributor to this repo since you've created quite a few PRs already and I think it would make things easier when you have write permission on the repo :)

hunhejj:
I appreciate that @szaimen! :)

hunhejj:
I will try to rework the Traefik part to make it more dummy-proof.

szaimen:
> I will try to rework the Traefik part to make it more dummy-proof.

Thank you very much! :)

PR #1750, opened 2023-01-13
Add Notes on Cloudflare to readme

Zoey2936:

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -128,6 +128,10 @@ Also, you may be interested in adjusting Nextcloud's Datadir to store the files 
 ### How to run AIO with Portainer?
 The easiest way to run it with Portainer on Linux is to use Portainer's stacks feature and use [this docker-compose file](./docker-compose.yml) in order to start AIO correctly. 
 
+### Notes on Cloudflare (proxy/tunnel)
+- Make sure to [disable Cloudflares Rocket Loader feature](https://help.nextcloud.com/t/login-page-not-working-solved/149417/8) as otherwise Nextcloud's login prompt will not be shown.
+- Cloudflare only supports uploading files up to 100 MB in the free plan, if you try to upload bigger files you will get an error (413 - Payload Too Large). If you need to upload bigger files, you need to disable the proxy option in your DNS settings, or you must use another proxy than Cloudflare tunnels. Both options will disable Cloudflare DDoS protection.
+
 ### How to run Nextcloud behind a Cloudflare Tunnel?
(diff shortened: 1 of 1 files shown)


readme.md
```diff
@@ -128,6 +128,10 @@ Also, you may be interested in adjusting Nextcloud's Datadir to store the files
 ### How to run AIO with Portainer?
 The easiest way to run it with Portainer on Linux is to use Portainer's stacks feature and use [this docker-compose file](./docker-compose.yml) in order to start AIO correctly. 
 
+### Notes on Cloudflare (proxy/tunnel)
+- Make sure to [disable Cloudflares Rocket Loader feature](https://help.nextcloud.com/t/login-page-not-working-solved/149417/8) as otherwise Nextcloud's login prompt will not be shown.
+- Cloudflare only supports uploading files up to 100 MB in the free plan, if you try to upload bigger files you will get an error (413 - Payload Too Large). If you need to upload bigger files, you need to disable the proxy option in your DNS settings, or you must use another proxy than Cloudflare tunnels. Both options will disable Cloudflare DDoS protection.
+
```
szaimen, line 134: Maybe we could add an advice regarding the wopi-allowlist?
https://github.com/nextcloud/all-in-one/discussions/1358

szaimen (commented):
Thanks, looks pretty good to me already, just a small question :)

readme.md
Zoey2936: So that these IPs should be added?
https://www.cloudflare.com/ips-v4
https://www.cloudflare.com/ips-v6

Zoey2936: But we could also directly add them by default to the list, or not?

szaimen: Hm, no I dont think this is a good idea as it introduces potential security problems when adding all cloudflare ip-addresses by default. Better would be adding a hint here that you need to follow the guide if collabora should not work as expected which is likely, imho.

Zoey2936:
```suggestion
- you should add all ips from [here](https://www.cloudflare.com/ips-v4) and [here](https://www.cloudflare.com/ips-v6) to this option: `Allow list for WOPI requests` on this site: `https://yourdomain.com/settings/admin/richdocuments`

```

szaimen:
```suggestion
- It is very likely that the in AIO included collabora (Nextcloud Office) does not work out of the box behind Cloudflare. You need to follow https://github.com/nextcloud/all-in-one/discussions/1358 in order to resolve this yourself. There is unfortunately no way to automate this for you.
```

szaimen (approved):
Thanks, LGTM! :)

PR #1756, opened 2023-01-14
add a hint that NEXTCLOUD_DATADIR needs to be specified before the fi…

szaimen:
…rst starutp


Verhoeckx (commented):
Yes, that looks good :blush: !

Instead of showing the explanations of the commands in the drop-down paragraph, I would add it below the instructions. This way it's not hidden for user/reader.

PR #1769, opened 2023-01-15
add further note about cloudflare and domain validation

szaimen:

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -126,6 +126,7 @@ Also, you may be interested in adjusting Nextcloud's Datadir to store the files 
 The easiest way to run it with Portainer on Linux is to use Portainer's stacks feature and use [this docker-compose file](./docker-compose.yml) in order to start AIO correctly. 
 
 ### Notes on Cloudflare (proxy/tunnel)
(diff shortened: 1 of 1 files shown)


szaimen:
@Zoey2936 btw, there are now already so many problems with Cloudflare that I slowly get tired of this...

Zoey2936:
> @Zoey2936 btw, there are now already so many problems with Cloudflare that I slowly get tired of this...

Yes... we could also mention this: https://github.com/nextcloud/all-in-one/discussions/1697

szaimen:
> Yes... we could also mention this: #1697

indeed: https://github.com/nextcloud/all-in-one/pull/1770

Zoey2936:
And stun/talk also wont work since cloudflare dont proxy (port 3478) them, we should mention that in the talk settings, the domain needs to be replaced with the external hostname / external ip of the server

szaimen:
> And stun/talk also wont work since cloudflare dont proxy (port 3478) them, we should mention that in the talk settings, the domain needs to be replaced with the external hostname / external ip of the server

right. Let me create an additional PR.

Zoey2936 (approved):

PR #1803, opened 2023-01-18
adjust macOS docs about DOCKER_SOCKET_PATH

szaimen:
Fix https://github.com/nextcloud/all-in-one/discussions/1801


szaimen:
@RetroYogi, does this look good to you? :)

RetroYogi:
> @RetroYogi, does this look good to you? :)

It's kind of an exception to an exception, but it's understandable :)

szaimen:
> It's kind of an exception to an exception, but it's understandable :)

yeah, I am also really confused... 😅

But good that it seems to work which was confirmed by you 👍

szaimen:
> can it not be the default value and keep unset?

yes, it could. However for current instances that may have the wrong value it would need to be set back to the default value as otherwise it will use the old value when simply removing it...

szaimen:
Adjusted, please have another look

Zoey2936 (approved):

szaimen:
Thanks Zoey! :)

PR #1809, opened 2023-01-19
note on proxy_read_timeout

Zoey2936:

### DIFF
--- a/reverse-proxy.md
+++ b/reverse-proxy.md
@@ -271,7 +271,7 @@ server {
         proxy_set_header Host $host;
         proxy_set_header X-Real-IP $remote_addr;
         proxy_set_header X-Forwarded-For $proxy_add_x_forwarded_for;
-        proxy_read_timeout 3600; # set it to the same value as this env: https://github.com/nextcloud/all-in-one#how-to-adjust-the-max-execution-time-for-nextcloud
+        proxy_read_timeout 86400s;
         client_max_body_size 0;
 
         # Websocket


reverse-proxy.md
```diff
@@ -271,6 +271,7 @@ server {
         proxy_set_header Host $host;
         proxy_set_header X-Real-IP $remote_addr;
         proxy_set_header X-Forwarded-For $proxy_add_x_forwarded_for;
+        proxy_read_timeout 3600; # set it to the same value as this env: https://github.com/nextcloud/all-in-one#how-to-adjust-the-max-execution-time-for-nextcloud
         client_max_body_size 0;
```
Zoey2936, line 275: I personally would leave on zero by default (so no change on this)
```suggestion
        client_max_body_size 10G; # set it to the same value as this env: https://github.com/nextcloud/all-in-one#how-to-adjust-the-upload-limit-for-nextcloud (or set it to 0 to allow any size)
```

szaimen: So then maybe just also set proxy_read_timeout to 0 since we already have both limits in apache/php

Zoey2936: > So then maybe just also set proxy_read_timeout to 0 since we already have both limits in apache/php

we could set it to 1d (one Day) by default in this config

reverse-proxy.md
```diff
@@ -271,6 +271,7 @@ server {
         proxy_set_header Host $host;
         proxy_set_header X-Real-IP $remote_addr;
         proxy_set_header X-Forwarded-For $proxy_add_x_forwarded_for;
+        proxy_read_timeout 3600; # set it to the same value as this env: https://github.com/nextcloud/all-in-one#how-to-adjust-the-max-execution-time-for-nextcloud
```
Zoey2936, line 274: One day:
```suggestion
        proxy_read_timeout 86400s;
```

reverse-proxy.md
szaimen: hm yeah, then maybe 2d which is likely close to unlimited or rather more?

Zoey2936: I would say one day is more than enough, I would say no one would upload a file which is so big that the upload takes over one day

reverse-proxy.md
szaimen:
```suggestion
        proxy_read_timeout 86400s; # This is equal to 24h. If you have so big files that they need more than 1 day to upload, it will fail tu upload them and lead to a 504 error
```

Zoey2936: The upload won't fail as already said in #979 it will just show an error, but the file will be uploaded without error

szaimen: ah I see. So then just use your suggestion. Still it is an error that gets triggered when exeeding this limit?

Zoey2936: > ah I see. So then just use your suggestion. Still it is an error that gets triggered when exeeding this limit?

The error will be shown when exceeding one day, but even if the file takes 3 days to upload, it will show the error after 3 days, but the file be saved on the cloud without errors

szaimen: okay, then just use your suggested value 👍

Sid127: change seems good to me, fwiw

reverse-proxy.md
```diff
@@ -271,6 +271,7 @@ server {
         proxy_set_header Host $host;
         proxy_set_header X-Real-IP $remote_addr;
         proxy_set_header X-Forwarded-For $proxy_add_x_forwarded_for;
+        proxy_read_timeout 86400s;
         client_max_body_size 0;
```
szaimen, line 275: btw, do we need both also for the other nginx based reverse proxies and what about the http2_body_preread_size?

Sid127 (approved):

reverse-proxy.md
Zoey2936: client_max_body_size:
- Original NPM has set it to [2GB](https://github.com/NginxProxyManager/nginx-proxy-manager/blob/fd30cfe98bba365e9630a791bdc86f01c08c6fa1/docker/rootfs/etc/nginx/nginx.conf#L36), it needs to be increased inside the advanced conf tab
- My NPM fork: https://github.com/ZoeyVid/nginx-proxy-manager/blob/21e765caa29cbf8987c5caf37d52096495488294/rootfs/usr/local/nginx/conf/nginx.conf#L29
- Synology no idea (my disk station seem to not support this proxy function)

proxy_read_timeout:
- Original NPM, needs to be set inside advanced tab
- My npm fork: https://github.com/ZoeyVid/nginx-proxy-manager/blob/21e765caa29cbf8987c5caf37d52096495488294/rootfs/usr/local/nginx/conf/nginx.conf#L46
- Synology no idea (my disk station seem to not support this proxy function)

And to http2_body_preread_size:
After reading [this](https://blog.cloudflare.com/delivering-http-2-upload-speed-improvements), I think we should instead set `client_body_buffer_size` to 256kb / 512kb

szaimen: Thanks Zoey! Lets handle this in a follow-up.

PR #1812, opened 2023-01-19
nginx proxy docs changes

Zoey2936:
- add more proxy headers
- change client_body_buffer_size
- enable ssl_early_data

Signed-off-by: Zoey <zoey@z0ey.de>

PR #1822, opened 2023-01-22
increase H2WindowSize limit (Reverse Proxy documentation)

DenuxPlays:
After recent tests, I found out that 512 KB is a more performant limit for the standard 10 GB upload limit.
With the 128 KB limit, the upload speed dropped from 6 MB/s to 4 MB/s after 1 GB was uploaded.
Now it stays between 5,5 MB/s and 6 MB/s for the entire upload.

Signed-off-by: -Denux <83671398+DenuxPlays@users.noreply.github.com>

### DIFF
--- a/reverse-proxy.md
+++ b/reverse-proxy.md
@@ -56,7 +56,7 @@ Add this as a new Apache site config:
     Protocols h2 h2c http/1.1
     
     # Solves slow upload speeds caused by http2
-    H2WindowSize 1048576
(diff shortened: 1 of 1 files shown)


Zoey2936 (changes requested):
The nginx part will be already handled by #1812, so the Apache change would be enough

Zoey2936 (approved):

szaimen (approved):
LGTM, thanks!

PR #1848, opened 2023-01-28
helm update script

szaimen:


helm-chart/templates/nextcloud-aio-apache-persistentvolumeclaim.yaml
```diff
@@ -9,7 +9,7 @@ spec:
   storageClassName: {{ .Values.STORAGE_CLASS }}
   {{- end }}
   accessModes:
-    - ReadWriteMany
+    - ReadWriteOnce
   resources:
     requests:
-      storage: {{ .Values.MAX_STORAGE_SIZE }}
+      storage: 1Gi
```
szaimen, line 15: So I googled about auto-resizing and it seems like there is no auto-resizing based on used storage. This is so dump because it means the admin needs to monitor this for each pvm as otherwise he/she risks that something runs out-of-space! Why is kubernetes does kubernetes not have this basic feature???

szaimen: Also now we need to make this adjustable at least foe the most-important vplumes that might exceed 1Gi...

szaimen: Or do you have an idea how to solve this problem systematically maybe?

helm-chart/templates/nextcloud-aio-database-persistentvolumeclaim.yaml
```diff
@@ -9,7 +9,7 @@ spec:
   storageClassName: {{ .Values.STORAGE_CLASS }}
   {{- end }}
   accessModes:
-    - ReadWriteMany
+    - ReadWriteOnce
   resources:
     requests:
-      storage: {{ .Values.MAX_STORAGE_SIZE }}
+      storage: 1Gi
```
szaimen, line 15: another question: does kubernetes handle the volume permissions automatically or do I need to adjust them manually? E.g. can I set them to 777? I suppose this works different here in contrast to docker volumes where correct permissions are automatically taken from the image...

szaimen: I looked into https://kubernetes.io/docs/tasks/configure-pod-container/security-context/#configure-volume-permission-and-ownership-change-policy-for-pods and can you tell me if https://github.com/nextcloud/all-in-one/pull/1852/files resolves the issue, WDYT?

szaimen: image

helm-chart/templates/nextcloud-aio-apache-persistentvolumeclaim.yaml
szaimen: Hopefully fixed with https://github.com/nextcloud/all-in-one/pull/1848/commits/8562eb09d43734438c0da65a0a9162e0b11256aa

helm-chart/update-helm.sh
```diff
@@ -28,21 +28,36 @@ sed -i "s|\${APACHE_PORT}:\${APACHE_PORT}/|$APACHE_PORT:$APACHE_PORT/|" latest.y
...
+# shellcheck disable=SC1083
+find ./ \( -not -name '*apache*' -not -name '*collabora-fonts*' -not -name '*trusted-cacerts*' -name '*persistentvolumeclaim.yaml' \) -exec sed -i "s|storage: 100Mi|storage: 1Gi|" \{} \;  
+# shellcheck disable=SC1083
+find ./ -name '*nextcloud-persistentvolumeclaim.yaml' -exec sed -i "s|storage: 1Gi|storage: 2Gi|" \{} \;  
+# shellcheck disable=SC1083
+find ./ -name '*nextcloud-data-persistentvolumeclaim.yaml' -exec sed -i "s|storage: 1Gi|storage: {{ .Values.NEXTCLOUD_DATA_STORAGE_SIZE }}|" \{} \;  
+# shellcheck disable=SC1083
+find ./ -name '*database*persistentvolumeclaim.yaml' -exec sed -i "s|storage: 1Gi|storage: {{ .Values.DATABASE_STORAGE_SIZE }}|" \{} \;  
+# shellcheck disable=SC1083
+find ./ -name '*elasticsearch-persistentvolumeclaim.yaml' -exec sed -i "s|storage: 1Gi|storage: {{ .Values.ELASTICSEARCH_STORAGE_SIZE }}|" \{} \;
```
szaimen, line 58: Rename to FULLTEXTSEARCH_STORAGE_SIZE? Also below?

szaimen: done

helm-chart/templates/nextcloud-aio-onlyoffice-persistentvolumeclaim.yaml
```diff
@@ -9,7 +9,7 @@ spec:
   storageClassName: {{ .Values.STORAGE_CLASS }}
   {{- end }}
   accessModes:
-    - ReadWriteMany
+    - ReadWriteOnce
   resources:
     requests:
-      storage: {{ .Values.MAX_STORAGE_SIZE }}
```
mrueg, line 15: It looks like the replacement doesn't work here anymore?

mrueg:
A couple of suggestions to make it look more like a helm chart:

* Lowercase all variables
* Drop the nextcloud-aio prefix in front of the file names (as it's already included in the chart name, this is uncommon)

helm-chart/Chart.yaml
```diff
@@ -1,6 +1,6 @@
 name: Nextcloud AIO Helm Chart
 description: A generated Helm Chart for Nextcloud AIO from Skippbox Kompose
-version: 4.0.1
+version: 4.2.0
 apiVersion: v1
```
mrueg, line 4: This could be v2 https://helm.sh/docs/topics/charts/#the-apiversion-field

helm-chart/templates/nextcloud-aio-apache-persistentvolumeclaim.yaml
```diff
@@ -9,7 +9,7 @@ spec:
   storageClassName: {{ .Values.STORAGE_CLASS }}
   {{- end }}
   accessModes:
-    - ReadWriteMany
+    - ReadWriteOnce
   resources:
     requests:
-      storage: {{ .Values.MAX_STORAGE_SIZE }}
+      storage: 100Mi
```
mrueg, line 15: Looks like this wasn't replaced anymore

mrueg:
fwiw, this might be also worth a try: https://github.com/arttor/helmify

helm-chart/templates/nextcloud-aio-onlyoffice-persistentvolumeclaim.yaml
szaimen: i initially didnt intend to make all volumes configurable but found a solution with https://github.com/nextcloud/all-in-one/pull/1848/commits/b6c3a777b0628cf3116b045bbade24e440d04eb9

helm-chart/templates/nextcloud-aio-apache-persistentvolumeclaim.yaml
szaimen: i initially didnt intend to make all volumes configurable but found a solution with https://github.com/nextcloud/all-in-one/pull/1848/commits/b6c3a777b0628cf3116b045bbade24e440d04eb9

helm-chart/Chart.yaml
szaimen: Are you completely sure? Then I can add this. I just used the files that [kompose](https://github.com/kubernetes/kompose) generated automatically.

szaimen:
Thanks for the feedback @mrueg!

> * Lowercase all variables

Thanks but since we automatically generate this with a bash script, I don't think it is worth the hassle.

> * Drop the nextcloud-aio prefix in front of the file names (as it's already included in the chart name, this is uncommon)

Thanks but since we generate this via a bash script using [kompose](https://github.com/kubernetes/kompose) and I don't think it is worth it.

> fwiw, this might be also worth a try: https://github.com/arttor/helmify

Thanks but what is the advantage over [kompose](https://github.com/kubernetes/kompose)? What would it change? Do you have an example with one of the files in this PR?

helm-chart/templates/nextcloud-aio-apache-deployment.yaml
```diff
@@ -16,11 +16,14 @@ spec:
...
         kompose.cmd: kompose convert -c -f latest.yml
-        kompose.version: 1.27.0 (b0ed6a2c9)
+        kompose.version: 1.28.0 (c4137012e)
       labels:
         io.kompose.network/nextcloud-aio: "true"
         io.kompose.service: nextcloud-aio-apache
     spec:
+      securityContext:
+        fsGroup: 65534
+        fsGroupChangePolicy: "OnRootMismatch"
```
szaimen, line 26: @mrueg do you know if this is the right way to solve possible permission issues with the volumes? I suppose kubernetes is not able to just use the correct permissions automatically that get applied inside a container image like docker volumes do?

helm-chart/Chart.yaml
mrueg: Correct me if I'm wrong: I think this file is not generated by kompose as kompose is not aware of helm charts.

mrueg:
> Thanks for the feedback @mrueg!
> 
> > * Lowercase all variables
> 
> Thanks but since we automatically generate this with a bash script from a docker-compose file, I don't think it is worth the hassle which also might lead to unexpected problems.
> 
> > * Drop the nextcloud-aio prefix in front of the file names (as it's already included in the chart name, this is uncommon)
> 
> Thanks but since we generate this via a bash script using [kompose](https://github.com/kubernetes/kompose) and I don't think it is worth it.
> 
> > fwiw, this might be also worth a try: https://github.com/arttor/helmify
> 
> Thanks but what is the advantage over [kompose](https://github.com/kubernetes/kompose)? What would it change? Do you have an example with one of the files in this PR?

What I'm suggesting is: `docker compose config  -> kompose -> kubernetes yaml files -> helmify -> helm chart` 

See here for an example:
https://github.com/arttor/helmify/blob/main/test_data/sample-app.yaml
gets you: https://github.com/arttor/helmify/tree/main/examples/app

helm-chart/Chart.yaml
szaimen: It actually is able to do so :)
https://github.com/nextcloud/all-in-one/blob/6ece0126809374350239c9811b96dc7dbf6cd1c5/helm-chart/update-helm.sh#L37

szaimen:
> What I'm suggesting is: `docker compose config -> kompose -> kubernetes yaml files -> helmify -> helm chart`
> 
> See here for an example: https://github.com/arttor/helmify/blob/main/test_data/sample-app.yaml gets you: https://github.com/arttor/helmify/tree/main/examples/app

Ah I see. However this looks a bit over-engineered to me as I am also not familiar with helmify. The current process is this:

`docker compose` -> `kompose` -> `helm chart`

Afterwards I just edit some details with a bash script and sed which I find much easier than modifying things with helmify. :)

helm-chart/Chart.yaml
szaimen: image

helm-chart/templates/nextcloud-aio-apache-persistentvolumeclaim.yaml
provokateurin: Ugh that solution is not good. Please just make every volume configurable. This is how helm charts work, they are complex and configurable :/

szaimen: Okay, I will add one for each pvc

szaimen: fixed with https://github.com/nextcloud/all-in-one/pull/1848/commits/133b149ee3272d6ac7815b3230edc5fd7215617a

helm-chart/templates/nextcloud-aio-apache-deployment.yaml
szaimen: Maybe @st3iny or @provokateurin know?

helm-chart/Chart.yaml
szaimen: Is this fine @mrueg ?
https://github.com/nextcloud/all-in-one/pull/1848/commits/c8576044b23dd2bd5b5fb80f02b96e4c90fa49a4

helm-chart/templates/nextcloud-aio-apache-persistentvolumeclaim.yaml
szaimen: improved with https://github.com/nextcloud/all-in-one/pull/1848/commits/133b149ee3272d6ac7815b3230edc5fd7215617a

helm-chart/templates/nextcloud-aio-onlyoffice-persistentvolumeclaim.yaml
szaimen: Improved with https://github.com/nextcloud/all-in-one/pull/1848/commits/133b149ee3272d6ac7815b3230edc5fd7215617a

helm-chart/templates/nextcloud-aio-apache-deployment.yaml
szaimen: Seems like this should work. Testing will show.

helm-chart/templates/nextcloud-aio-apache-persistentvolumeclaim.yaml
st3iny: There is auto resizing for some vendored cloud distributions (e.g. GKE). But yeah, for vanilla k8s the volume has to be manually recreated and migrated.

st3iny: Or something like https://longhorn.io/ can be used to easen volume handling.

szaimen: I see but I suppose this is up to the admin to choose?

st3iny: Exactly. Let admins handle their comes and just provide distinct variables.

szaimen: That is what I've done now :)

helm-chart/templates/nextcloud-aio-apache-deployment.yaml
```diff
@@ -16,11 +16,25 @@ spec:
...
         io.kompose.network/nextcloud-aio: "true"
         io.kompose.service: nextcloud-aio-apache
     spec:
+      initContainers:
+        - name: init-volumes
+          image: alpine
+          command:
+            - chmod
+            - -R
+            - 777
```
st3iny, line 30:
```suggestion
            - "777"
```

Should probably be a string.

helm-chart/templates/nextcloud-aio-clamav-deployment.yaml
```diff
@@ -17,18 +17,29 @@ spec:
...
         io.kompose.network/nextcloud-aio: "true"
         io.kompose.service: nextcloud-aio-clamav
     spec:
+      initContainers:
+        - name: init-volumes
+          image: alpine
+          command:
+            - chmod
+            - -R
+            - 777
```
st3iny, line 31: See above.

helm-chart/templates/nextcloud-aio-collabora-deployment.yaml
```diff
@@ -17,11 +17,22 @@ spec:
...
         io.kompose.network/nextcloud-aio: "true"
         io.kompose.service: nextcloud-aio-collabora
     spec:
+      initContainers:
+        - name: init-volumes
+          image: alpine
+          command:
+            - chmod
+            - -R
+            - 777
```
st3iny, line 31: See above.

helm-chart/templates/nextcloud-aio-database-deployment.yaml
```diff
@@ -16,11 +16,25 @@ spec:
...
         io.kompose.network/nextcloud-aio: "true"
         io.kompose.service: nextcloud-aio-database
     spec:
+      initContainers:
+        - name: init-volumes
+          image: alpine
+          command:
+            - chmod
+            - -R
+            - 777
```
st3iny, line 30: See above.

helm-chart/templates/nextcloud-aio-fulltextsearch-deployment.yaml
```diff
@@ -17,11 +17,22 @@ spec:
...
         io.kompose.network/nextcloud-aio: "true"
         io.kompose.service: nextcloud-aio-fulltextsearch
     spec:
+      initContainers:
+        - name: init-volumes
+          image: alpine
+          command:
+            - chmod
+            - -R
+            - 777
```
st3iny, line 31: See above.

helm-chart/templates/nextcloud-aio-nextcloud-deployment.yaml
```diff
@@ -16,11 +16,28 @@ spec:
...
         io.kompose.network/nextcloud-aio: "true"
         io.kompose.service: nextcloud-aio-nextcloud
     spec:
+      initContainers:
+        - name: init-volumes
+          image: alpine
+          command:
+            - chmod
+            - -R
+            - 777
```
st3iny, line 30: See above.

helm-chart/templates/nextcloud-aio-onlyoffice-deployment.yaml
```diff
@@ -17,11 +17,22 @@ spec:
...
         io.kompose.network/nextcloud-aio: "true"
         io.kompose.service: nextcloud-aio-onlyoffice
     spec:
+      initContainers:
+        - name: init-volumes
+          image: alpine
+          command:
+            - chmod
+            - -R
+            - 777
```
st3iny, line 31: See above.

helm-chart/templates/nextcloud-aio-redis-deployment.yaml
```diff
@@ -16,18 +16,29 @@ spec:
...
         io.kompose.network/nextcloud-aio: "true"
         io.kompose.service: nextcloud-aio-redis
     spec:
+      initContainers:
+        - name: init-volumes
+          image: alpine
+          command:
+            - chmod
+            - -R
+            - 777
```
st3iny, line 30: See above.

helm-chart/update-helm.sh
```diff
@@ -28,21 +28,49 @@ sed -i "s|\${APACHE_PORT}:\${APACHE_PORT}/|$APACHE_PORT:$APACHE_PORT/|" latest.y
...
+# shellcheck disable=SC1083
+find ./ -name '*networkpolicy.yaml' -exec sed -i "s|manual-install-nextcloud-aio|nextcloud-aio|" \{} \; 
+cat << EOL > /tmp/initcontainers
+      initContainers:
+        - name: init-volumes
+          image: alpine
+          command:
+            - chmod
+            - -R
+            - 777
```
st3iny, line 50:
```suggestion
            - "777"
```

See above.

helm-chart/Chart.yaml
```diff
@@ -1,7 +1,7 @@
 name: Nextcloud AIO Helm Chart
 description: A generated Helm Chart for Nextcloud AIO from Skippbox Kompose
-version: 4.0.1
-apiVersion: v1
+version: 4.2.0
+apiVersion: v2
```
mrueg, line 4: @szaimen looks good. I've sent https://github.com/kubernetes/kompose/pull/1567 upstream and it got merged, so next kompose release should be v2 by default.

helm-chart/templates/nextcloud-aio-apache-deployment.yaml
szaimen: fixed with the latest commit

helm-chart/update-helm.sh
szaimen: fixed with the latest commit

szaimen: fixed with the latest commitfixed with the latest commit

helm-chart/Chart.yaml
szaimen: all right! Thanks a lot! :)

PR #1887, opened 2023-02-01
get multiarch repodigest

szaimen:

### DIFF
--- a/php/src/Docker/DockerHubManager.php
+++ b/php/src/Docker/DockerHubManager.php
@@ -26,29 +26,31 @@ class DockerHubManager
         // If one of the links below should ever become outdated, we can still upgrade the mastercontainer via the webinterface manually by opening '/api/docker/getwatchtower'
 
         try {
-            $authTokenRequest = $this->guzzleClient->request(
+            $request = $this->guzzleClient->request(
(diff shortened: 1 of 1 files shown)


php/src/Docker/DockerHubManager.php
```diff
@@ -26,29 +26,30 @@ public function GetLatestDigestOfTag(string $name, string $tag) : ?string {
         // If one of the links below should ever become outdated, we can still upgrade the mastercontainer via the webinterface manually by opening '/api/docker/getwatchtower'
 
         try {
-            $authTokenRequest = $this->guzzleClient->request(
+            $request = $this->guzzleClient->request(
                 'GET',
-                'https://auth.docker.io/token?service=registry.docker.io&scope=repository:' . $name . ':pull'
+                'https://hub.docker.com/v2/repositories/' . $name . '/tags?page_size=128'
```
szaimen, line 31: @Zoey2936 , do you think requesting the most recent 128 tags is enough?

Zoey2936: I would say yes

szaimen: okay, thanks

PR #1890, opened 2023-02-01
fix daily backup running into conflict with elasticsearch

szaimen:
Fix https://github.com/nextcloud/all-in-one/discussions/1875

### DIFF
--- a/Containers/fulltextsearch/Dockerfile
+++ b/Containers/fulltextsearch/Dockerfile
@@ -11,5 +11,10 @@ RUN set -ex; \
     ; \
     rm -rf /var/lib/apt/lists/*
 
+COPY start.sh /
+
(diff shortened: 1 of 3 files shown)


szaimen:
@Zoey2936 do you maybe have a good idea how to fix https://github.com/nextcloud/all-in-one/discussions/1875 in a more general way?

Zoey2936:
At the moment not

szaimen:
> At the moment not

yeah, same for me unfortunately...

szaimen:
WDYT about adding a tcp connection to the backup container that other container listen to?

E.g. 
```
while nc -z "$BACKUP_HOST" "$BACKUP_HOST" &>/dev/null; then
    echo "waiting for backup container to stop..."
fi
```

PR #1904, opened 2023-02-02
helm - add network connections

szaimen:

### DIFF
--- a/helm-chart/network.yaml
+++ b/helm-chart/network.yaml
@@ -1,0 +1,136 @@
+---
+kind: Service
+apiVersion: v1
+metadata:
+  name: nextcloud-aio-apache
+spec:
+  selector:
+    app: nextcloud-aio-apache
+  ports:
+    - port: {{ Values.APACHE_PORT }}
+      protocol: TCP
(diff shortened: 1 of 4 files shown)


helm-chart/network.yaml
```diff
@@ -0,0 +1,136 @@
...
+metadata:
+  name: nextcloud-aio-apache
+spec:
+  selector:
+    app: nextcloud-aio-apache
+  ports:
+    - port: {{ Values.APACHE_PORT }}
+      protocol: TCP
+      name: nextcloud-aio-apache
+  type: ClusterIP
```
szaimen, line 13: @st3iny can you tell if this will collide with https://github.com/nextcloud/all-in-one/blob/main/helm-chart/templates/nextcloud-aio-apache-service.yaml? So do we need this service even? 
Additionally, what is this doing? https://github.com/nextcloud/all-in-one/blob/b428f3d1d89ea8a4f873e37539aa16bea906f894/helm-chart/templates/nextcloud-aio-apache-deployment.yaml#L59-L61

szaimen:
Will do it differently by exposing the ports in the docker-compose.

PR #1962, opened 2023-02-10
Add instructions for retrieving initial password

fakoor:
Signed-off-by: Ali <fakoor@gmail.com>


szaimen:
Hi, if the password was not shown to you, someone else may have seen it so if that was the case for you, I would rather recommend to reset aio properly: https://github.com/nextcloud/all-in-one#how-to-properly-reset-the-instance and start with a fresh install.

PR #1969, opened 2023-02-11
fix and hide a few caddy warnings

szaimen:
Close https://github.com/nextcloud/all-in-one/discussions/1965

### DIFF
--- a/Containers/mastercontainer/Caddyfile
+++ b/Containers/mastercontainer/Caddyfile
@@ -16,7 +16,12 @@ http://:80 {
   redir https://{host}{uri}
 }
 
-https://:8443 {
+# Match only host names and not ip-addresses:
+https://*.*:8443,
+https://*.*.*:8443,
+https://*.*.*.*:8443,
+https://*.*.*.*.*:8443,
+https://*.*.*.*.*.*:8443 {
 
     reverse_proxy localhost:8000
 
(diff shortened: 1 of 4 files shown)


Containers/mastercontainer/start.sh
```diff
@@ -280,4 +280,10 @@ export TZ=UTC
 # Fix apache startup
 rm -f /var/run/apache2/httpd.pid
 
+# Fix the Caddyfile format
+caddy fmt --overwrite /Caddyfile
+
+# Fix caddy log 
+chmod 777 /root
```
Zoey2936, line 287: Does it work without `chmod -R`?

szaimen: yes

Containers/mastercontainer/Caddyfile
```diff
@@ -6,17 +6,29 @@
...
   redir https://{host}{uri}
 }
 
-https://:8443 {
+# Match only host names and not ip-addresses:
+https://*.*:8443,
+https://*.*.*:8443,
+https://*.*.*.*:8443,
+https://*.*.*.*.*:8443,
+https://*.*.*.*.*.*:8443 {
```
szaimen, line 24: @Zoey2936 this would match at max 1.2.3.4.yourdomain.com:8443. Do you think this is fine or should we add another level?

Zoey2936: This should be enough, the least people will use more than a subdomain of a subdomain

szaimen: Okay good. Thanks!

PR #2000, opened 2023-02-16
document multiarch containers

szaimen:


Zoey2936:
Maybe we should also somewhere mention that AIO only supports arm64/aarch64/armv8 and amd64/x86/x86_64?
32Bit (armv7) Raspberry Pi OS is still often used and still preselected by the Raspberry Pi imager

szaimen:
> Maybe we should also somewhere mention that AIO only supports arm64/aarch64/armv8 and amd64/x86/x86_64?

Sounds good. Would you mind creating a PR? :)

Zoey2936:
I'm not sure where a good place is, to mention this inside the readme, I thought about the first line
And also how to call arm64/aarch64/armv8 and amd64/x86/x86_64 exactly because of this many names

szaimen:
Maybe just add it somewhere in the FAQ section? I would not like to put it into the instructions...

szaimen:
Thanks!

PR #2017, opened 2023-02-18
add gcompat in case of arm64 for recognize

szaimen:


Containers/nextcloud/start.sh
```diff
@@ -50,6 +50,17 @@ if ! [ -f "$NEXTCLOUD_DATA_DIR/this-is-a-test-file" ]; then
...
 
+# Correctly set CPU_ARCH for notify_push
+CPU_ARCH="$(uname -m)"
+export CPU_ARCH
+if [ -z "$CPU_ARCH" ]; then
+    echo "Could not get processor architecture. Exiting."
+    exit 1
+elif [ "$CPU_ARCH" != "x86_64" ]; then
+    export CPU_ARCH="aarch64"
+    ADDITIONAL_APKS+=" gcompat"
```
szaimen, line 61: @Zoey2936 do you have [recognize](https://github.com/nextcloud/recognize) running on arm64? If yes, was gcompat for it required to work?

Zoey2936: I run AIO only on amd64 and there it works without gcompat, so no Idea if it is required on aarch64

szaimen: Ah I see, thanks nonetheless!

szaimen:
Not needed after https://github.com/nextcloud/recognize/commit/30602dcfa972257f4fae1534fb2419ef319acdb9

Issue #2041, opened 2023-02-21
nextcloud logs remote ip-address as the gateways ip-address


szaimen:
We need to find a way to work around this. I am sure that it has worked at some point but seems like something changed and now it doesnt anymore. Can be easily verified by trying to log in with wrong credentials.

szaimen:
@Zoey2936 could it be possible that this was cause by the apache migration to alpine?

szaimen:
Looks like adding `"userland-proxy": false` to /etc/docker/daemon.json made it work...

szaimen:
> is this only in non reverse proxy mode? since in /settings/admin/logging is see the correct ips

No, it is in default mode... Maybe it is due to the docker daemon update to 23 that I recently installed?

Zoey2936:
I also use docker v23.0.1...

szaimen:
That is very strange. but since it fixes the issue for me, I guess it is worth documenting.

szaimen:
yeah, don't know why it works for you even without that setting. Anyhow, it is useful to document this imho.

Zoey2936:
I know why it works for me...
We trust private IPs only in reverse proxy mode: https://github.com/nextcloud/all-in-one/blob/dfc2d9798ee69b0183b4ad0299ab2e187b76d436/Containers/apache/start.sh#L42

szaimen:
> We trust private IPs only in reverse proxy mode:

I thought the same and tried this but it didnt made it work for me...

szaimen:
The actual only thing that made it work was disabling the userland-proxy.

szaimen:
I just tried that again and it didnt work without disabling the userland-proxy. Anyhow, it might come with a few downside so the wiki article must be enough for now.

szaimen:
@Zoey2936 I just had the thought why it works for you: this issue does not appear if the container uses the network_mode host. Only if the container runs in a bridged network.

szaimen:
So basically I still think that this is the way to go for a setup without reverse proxy. Reopening https://github.com/nextcloud/all-in-one/pull/2044

Zoey2936:
I've set up a new AIO instance without reverse proxy, and it is working:
![grafik](https://user-images.githubusercontent.com/75573284/220406749-599f4054-fe65-4927-ad6c-c6b8443fc203.png)

Zoey2936:
root@ip:~# docker version
Client: Docker Engine - Community
 Version:           23.0.1
[...]

PR #2044, opened 2023-02-21
document to disable userland-proxy

szaimen:
Fix https://github.com/nextcloud/all-in-one/issues/2041


szaimen:
Not doing this for now since it might come with a few downsides.

PR #2051, opened 2023-02-22
Cron: remove nextcloud-aio-domaincheck if stopped

Zoey2936:


Containers/mastercontainer/cron.sh
```diff
@@ -47,8 +47,14 @@ while true; do
     # Remove sessions older than 24h
     find "/mnt/docker-aio-config/session/" -mindepth 1 -mmin +1440 -delete
 
+    # Remove nextcloud-aio-domaincheck container
+    sudo -u www-data docker container remove nextcloud-aio-domaincheck --force
```
szaimen, line 51: we should only remove the container if it is stopped. You should be able to get it with e.g.  docker ps --filter "status=exited"

Containers/mastercontainer/cron.sh
```diff
@@ -47,8 +47,14 @@ while true; do
...
+
     # Remove dangling images
-    sudo -u www-data docker image prune -f
+    sudo -u www-data docker image prune --force
+    
+    # Remove dangling volumes
+    for volume in $(sudo -u www-data docker volume ls --quiet --filter dangling=true | grep -E "^[a-z0-9]{64}$"); do sudo -u www-data docker volume remove "$volume"; done
```
szaimen, line 57: I am honestly against removing any volumes automatically.

The better way is to remove anonymous volumes when the container gets removed which we already do in https://github.com/nextcloud/all-in-one/blob/3f22ec8b4d26cb3b8275cd7ae8a71ba2c957f067/php/src/Docker/DockerActionManager.php#L148

Containers/mastercontainer/cron.sh
Zoey2936:
```suggestion
    if [ "$(docker ps --format "{{.Names}}" --filter "status=exited" | grep -w "nextcloud-aio-domaincheck" | head -n 1)" == "nextcloud-aio-domaincheck" ]; then
        sudo -u www-data docker container remove nextcloud-aio-domaincheck
    fi
```

szaimen:
```suggestion
    if [ "$(docker ps --format "{{.Names}}" --filter "status=exited" | grep -q "^nextcloud-aio-domaincheck$")" == "nextcloud-aio-domaincheck" ]; then
        sudo -u www-data docker container remove nextcloud-aio-domaincheck
    fi
```

Zoey2936:
```suggestion
    if docker ps --format "{{.Names}}" --filter "status=exited" | grep -qw "^ttt$"; then
        sudo -u www-data docker container remove nextcloud-aio-domaincheck
    fi
```

Containers/mastercontainer/cron.sh
Zoey2936, line 57:
```suggestion
```

Containers/mastercontainer/cron.sh
```diff
@@ -47,8 +47,13 @@ while true; do
     # Remove sessions older than 24h
     find "/mnt/docker-aio-config/session/" -mindepth 1 -mmin +1440 -delete
 
+    # Remove nextcloud-aio-domaincheck container
+    if docker ps --format "{{.Names}}" --filter "status=exited" | grep -qw "^nextcloud-aio-domaincheck$"; then
+        sudo -u www-data docker container remove nextcloud-aio-domaincheck
```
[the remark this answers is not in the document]
Zoey2936: why not `container remove`?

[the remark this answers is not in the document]
szaimen: was just a suggestion. container remove is also fine for me

Containers/mastercontainer/cron.sh
```diff
@@ -47,8 +47,13 @@ while true; do
     # Remove sessions older than 24h
     find "/mnt/docker-aio-config/session/" -mindepth 1 -mmin +1440 -delete
 
+    # Remove nextcloud-aio-domaincheck container
+    if sudo -u www-data docker ps --format "{{.Names}}" --filter "status=exited" | grep -qw "^nextcloud-aio-domaincheck$"; then
```
szaimen, line 51:
```suggestion
    if sudo -u www-data docker ps --format "{{.Names}}" --filter "status=exited" | grep -q "^nextcloud-aio-domaincheck$"; then
```

Zoey2936: it works with -w and without -w also in my test

szaimen (approved):
LGTM, thanks!

szaimen:
This is now released with v4.5.0 Beta. Testing and feedback is welcome! See https://github.com/nextcloud/all-in-one#how-to-switch-the-channel

Zoey2936:
maybe it would be better to instead enable the autoremove option from docker for this container, while creating it

szaimen:
> maybe it would be better to instead enable the autoremove option from docker for this container, while creating it

Hm... not sure about this. This would make debuggin much harder if it for some reason cannot start correctly, exited and get removed directly without any log or so what went wrong...

Zoey2936:
thats true... but it would be a native docker implementation

Zoey2936:
but will the container not restart if it crashes? since it uses restart: unless-stopped?

szaimen:
> but will the container not restart if it crashes? since it uses restart: unless-stopped?

No it will not but you can still access its logs because it is not removed directly ;)

Zoey2936:
I thought unless-stoped means that container will first be stopped/stopped from restarting until it is manually? or does the container not use unless-stopped?

szaimen:
> or does the container not use unless-stopped?

it does not use unless-stopped

Zoey2936:
Ah, then auto remove makes no sense

PR #2091, opened 2023-03-02
improve the volume and container name check

szaimen:


Zoey2936:
Do I understand this correct that it will now check for the exact volume name, instead of just looking if it contains parts of the name?

szaimen:
Yes

Zoey2936 (approved):

PR #2092, opened 2023-03-02
don't wait for clamav until forever

szaimen:
Close https://github.com/nextcloud/all-in-one/discussions/2089
Close https://www.reddit.com/r/NextCloud/comments/11ftptu/comment/jalhipy/
This is best reviewed like this: https://github.com/nextcloud/all-in-one/pull/2092/files?diff=unified&w=1

### DIFF
--- a/Containers/nextcloud/entrypoint.sh
+++ b/Containers/nextcloud/entrypoint.sh
@@ -534,23 +534,30 @@ fi
 
 # Clamav
 if [ "$CLAMAV_ENABLED" = 'yes' ]; then
-    while ! nc -z "$CLAMAV_HOST" 3310; do
+    count=0
+    while ! nc -z "$CLAMAV_HOST" 3310 && [ "$count" -lt 90 ]; do
         echo "waiting for clamav to become available..."
+        count=$((count+5))
         sleep 5
     done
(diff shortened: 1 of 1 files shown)


Containers/nextcloud/entrypoint.sh
```diff
@@ -534,23 +534,29 @@ fi
...
         sleep 5
     done
-    if ! [ -d "/var/www/html/custom_apps/files_antivirus" ]; then
-        php /var/www/html/occ app:install files_antivirus
-    elif [ "$(php /var/www/html/occ config:app:get files_antivirus enabled)" != "yes" ]; then
-        php /var/www/html/occ app:enable files_antivirus
-    elif [ "$SKIP_UPDATE" != 1 ]; then
-        php /var/www/html/occ app:update files_antivirus
+    if [ "$count" -ge 90 ]; then
+        echo "Clamav did not start in time. Skipping initialization."
```
szaimen, line 544: should we disable the clamav app in that case? As it might break uploads if the clamav host is not reachable...

Zoey2936: would say yes

Zoey2936 (approved):

szaimen:
Thanks Zoey! :)

PR #2102, opened 2023-03-03
Fix directory path example

shaan7:
`/run/desktop/mnt/host/c/ncdata` does not really contain the word `backup`, so I'm guessing `C:\backup` is a typo.


szaimen (approved):
Thanks!

PR #2109, opened 2023-03-05
change how permission check setting works

szaimen:
Close https://github.com/nextcloud/all-in-one/discussions/2108

### DIFF
--- a/Containers/nextcloud/entrypoint.sh
+++ b/Containers/nextcloud/entrypoint.sh
@@ -212,9 +212,8 @@ if ! [ -f "$NEXTCLOUD_DATA_DIR/skip.update" ]; then
                 exit 1
             fi
 
-            if [ "$SKIP_DATA_DIRECTORY_PERMISSION_CHECK" = yes ]; then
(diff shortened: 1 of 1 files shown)


Zoey2936:
so it is disabled, unless manually enabled?

szaimen:
> so it is disabled, unless manually enabled?

the permission check gets now always skipped upon initial install but after initial install the permission check gets enabled again if not skipped via environmental variable.

Zoey2936:
> > so it is disabled, unless manually enabled?
> 
> the permission check gets now always skipped upon initial install but after initial install the permission check gets enabled again if not skipped via environmental variable.

but why?

szaimen:
> but why?

Because if we don't do this the initial installation will fail if you e.g. have added 777 permisions. The check in the container is checking if it can write but nextcloud installation will fail because permissions are too open. So I'd rather skip the permission check of Nextcloud always during the initial installationi since we have our own check in place. Afterwards I add the permission check again if it should not get skipped so that it is shown in the admin overview and you can do needed changes afterwarrds.

Zoey2936 (approved):

szaimen:
Thanks! :)

PR #2185, opened 2023-03-23
enable JIT

szaimen:
based on https://github.com/nextcloud/documentation/pull/9230


DenuxPlays:
I am not that familiar with php, but could this mean that we feel some serious performance improvements?

szaimen:
> I am not that familiar with php, but could this mean that we experience some serious performance improvements?

No, probably rather minor improvements.

PR #2207, opened 2023-03-25
add dot to allowed characters in additional backup volumes

szaimen:
Close https://github.com/nextcloud/all-in-one/discussions/2203


szaimen (commented):
Might also need an update of the description in the ui

PR #2273, opened 2023-04-04
add default path for borg compact

MariusQuabeck:
all examples up until the compact section work under the assumption that you are using the default backup path "/mnt/backup/borg" so path for compact should be the same 💙


szaimen (approved):
Thanks! :)

PR #2289, opened 2023-04-06
add nodejs in order to make recognize easier to install on arm64

szaimen:
See https://github.com/nextcloud/recognize/commit/af8f727e028874dd3d70ed50ce438efc975ea3e9

### DIFF
--- a/Containers/nextcloud/Dockerfile
+++ b/Containers/nextcloud/Dockerfile
@@ -205,6 +205,7 @@ RUN set -ex; \
         mawk \
         sudo \
         grep \
+        nodejs \
         coreutils;
 
 RUN set -ex; \


Containers/nextcloud/Dockerfile
```diff
@@ -205,6 +205,7 @@ RUN set -ex; \
         mawk \
         sudo \
         grep \
+        nodejs \
```
Zoey2936, line 208: Does it require `nodejs ` (lts) or `nodejs-current` (current)?

szaimen: cc @marcelklehr?

szaimen: I think nodejs should be sufficient

PR #2301, opened 2023-04-08
Talk use netcat to check stun/turn port using domain

Zoey2936:


Containers/talk/Dockerfile
```diff
@@ -67,5 +67,5 @@ USER talk
 ENTRYPOINT ["start.sh"]
 CMD ["/usr/bin/supervisord", "-c", "/supervisord.conf"]
 
-HEALTHCHECK CMD (nc -z localhost 8081 && nc -z localhost 8188 && nc -z localhost 4222 && nc -z localhost $TALK_PORT) || exit 1
+HEALTHCHECK CMD (nc -z localhost 8081 && nc -z localhost 8188 && nc -z localhost 4222 && nc -z localhost "$TALK_PORT" && nc -z "$NC_DOMAIN" "$TALK_PORT") || exit 1
```
szaimen, line 70: I fear this will not work behin cloudflare so this would mean it would not be possible to fix the healthcheck behind that?

Zoey2936: It would mark the talk container as unhealthy, if the container runs behind cloudflare or the firewall is closed

szaimen: > It would mark the talk container as unhealthy, if the container runs behind cloudflare

yes, but this is not what we want, no?

> the firewall is closed

That would actually be an improvement, yes

Zoey2936: > > It would mark the talk container as unhealthy, if the container runs behind cloudflare
> 
> yes, but this is not what we want, no?
why? If talk runs behind cloudflare, the talk container would never work.. so it would be unhealthy

szaimen: True but in that case they should then spin-up their own turn server and hpb and keep the talk container enabled (because otherwise the spreed app gets disabled) which would then always report it as unhealthy

Zoey2936: AIO disabled spread, if talk is disabled?

szaimen: yes, if you disable the talk option in the aio interface, the spreed app also gets disabled...

Zoey2936: but why this? would it not be better to let the app enabled? Because I think this change would be a good improvement to the healthcheck of the talk container

szaimen: It is the same logic for the other apps as well since we want to make sure that removing the talk container app does not break the app. If the talk container would be disabled and the spreed app kept enabled it would basically break it because the turnserver and hpb would of course not work...

Zoey2936: and if the app only get disabled if one of hpb/stun/turn settings is set to the aio defaults?

szaimen: > and if the app only get disabled if one of hpb/stun/turn settings is set to the aio defaults?

I fear this would be much more complicated and also more error-prone...

Zoey2936: > True but in that case they should then spin-up their own turn server and hpb and keep the talk container enabled (because otherwise the spreed app gets disabled) which would then always report it as unhealthy

but where is there the problem, if the container is  not reachable, it is unhealthy, even if the people don't use it...

szaimen: okay, lets do this then

szaimen:
This is now released with v4.9.0 Beta. Testing and feedback is welcome! See https://github.com/nextcloud/all-in-one#how-to-switch-the-channel

PR #2302, opened 2023-04-08
apache healthcheck, check localhost apacheport with every apache port

Zoey2936:


szaimen:
Thanks Zoey! Is there a reason for this change?

Zoey2936:
> Thanks Zoey! Is there a reason for this change?

No... But I thought that localhost can be checked every time

Zoey2936:
Other question, AIO only supports offially port 443, what about also alwas checking port 443 from external?

szaimen:
> Other question, AIO only supports offially port 443, what about also alwas checking port 443 from external?

That would basically mean that the healthcheck would break behind cloudflare, IIRC?

szaimen:
> > Thanks Zoey! Is there a reason for this change?
> 
> No... But I thought that localhost can be checked every time

Does this have any benefit? (I mean if the domain and 443 works, the container should be healthy, no?

Zoey2936:
> 

why should it break? cloudflare has port 443 open

Zoey2936:
> > > Thanks Zoey! Is there a reason for this change?
> > 
> > 
> > No... But I thought that localhost can be checked every time
> 
> Does this have any benefit? (I mean if the domain and 443 works, the container should be healthy, no?

if it runs behind cloudflare the port is open, but the apache container could be stopped

szaimen:
> > 
> 
> why should it break? cloudflare has port 443 open

In the past were connection attempts blocked by cloudflare... but you are probably right because this only checks if a tcp connection is possible which is likely not blocked by cloudflare.

szaimen:
However what about scenarios where like when using dns challenge where the port will not be opened?

szaimen (approved):
Fine by me in the current form

Zoey2936:
> However what about scenarios where like when using dns challenge where the port will not be opened?

you mean in local instances? even they need to have port 443 open and the nextcloud domain should point to the domain

szaimen:
> you mean in local instances? even they need to have port 443 open and the nextcloud domain should point to the domain

True. Okay then lets always check it

Containers/apache/healthcheck.sh
```diff
@@ -2,8 +2,5 @@
 
 nc -z "$NEXTCLOUD_HOST" 9000 || exit 0
 nc -z localhost 8000 || exit 1
-if [ "$APACHE_PORT" != '443' ]; then
-    nc -z localhost "$APACHE_PORT" || exit 1
-else
-    nc -z "$NC_DOMAIN" "$APACHE_PORT" || exit 1
-fi
+nc -z localhost "$APACHE_PORT" || exit 1
+nc -z "$NC_DOMAIN" "$APACHE_PORT" || exit 1
```
szaimen, line 6:
```suggestion
nc -z "$NC_DOMAIN" 443 || exit 1
```

szaimen (changes requested):

szaimen (approved):

szaimen:
This is now released with v4.9.0 Beta. Testing and feedback is welcome! See https://github.com/nextcloud/all-in-one#how-to-switch-the-channel

PR #2327, opened 2023-04-13
Missing brace in readme.md

ItsQuadrus:
`sudo docker volume ls --format {{.Name}` command is invalid and gives the error `template parsing error: template: :1: unexpected "}" in operand`

This change provides a fix by adding the missing brace.


szaimen (approved):
Thanks! :)

PR #2334, opened 2023-04-14
disabling FTS on NC26 as the migration to ES8 is still pending

szaimen:


Containers/nextcloud/entrypoint.sh
```diff
@@ -616,6 +616,11 @@ if version_greater "$installed_version" "24.0.0.0"; then
     fi
 fi
 
+# Migration to ES8 is pending, thus disabling FTS for now.
+if [ "$INSTALL_LATEST_MAJOR" = yes ] || version_greater "$installed_version" "26.0.0.0"; then
+    export FULLTEXTSEARCH_ENABLED=no
```
szaimen, line 621: todo: also disable in case of upgrade scenario?

PR #2340, opened 2023-04-14
Proxy instructions for NodeJS/express

JL102:
Added instructions for how to proxy network requests for AIO when running a NodeJS server. I personally use CloudFlare's proxying, so it technically works just with the httpServer. The instructions are the same for those using httpsServer, and I believe that anyone who was knowledgeable enough to set up a nodejs server will be able to adapt the instructions to their specific use-case.

### DIFF
--- a/reverse-proxy.md
+++ b/reverse-proxy.md
@@ -352,6 +352,89 @@ Of course you need to modify `<your-nc-domain>` to the domain on which you want 
 
 </details>
 
+### Node.js with Express
+
+<details>
+
+<summary>click here to expand</summary>
+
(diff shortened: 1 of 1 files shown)


szaimen:
@JL102 thanks a lot for your PR! Sorry for coming so late back to you! I rebased it and added some small changes. 

In general it looks good to me but please fix DCO. See https://github.com/nextcloud/all-in-one/pull/2340/checks?check_run_id=13146744413 for how to fix it. We need that so that we can merge your change! Thanks a lot!

JL102:
@szaimen np, done! I configured VS Code to auto signoff my commits now 😉

Issue #2348, opened 2023-04-15
IPv6 not enabled for the custom nextcloud-aio network when ipv6 is enabled for the global network


szaimen:
It Seems like ipv6 need to be enabled manually, however I have not yet found a way to get a working subnet...

szaimen:
@Zoey2936 apparently https://github.com/nextcloud/all-in-one/issues/2041#issuecomment-1438692762 was needed as well to make ipv6 support work in any way. However we would need to add the subnet and gateway manually which I was not able to. 

So honestly after checking this a bit, it feels much safer to just use the default bridge network which enables the things automatically but of course does not allow to configure things anymore and the containers would not be separated form other containers running in the bridge network anymore

szaimen:
I'll tackle this by allowing to define a custom network name. WIP here: https://github.com/nextcloud/all-in-one/pull/2349

Also the ipv6 instructions will get adjusted.

Zoey2936:
one question: what is the goal?  I thought enabling ipv6 inside the deamon.json would be enough to make port 443 accessable over ipv6?

Zoey2936:
why do we need internal ipv6 support in docker?

szaimen:
> I thought enabling ipv6 inside the deamon.json would be enough to make port 443 accessable over ipv6?

Is that so? Let me check this.

Zoey2936:
btw. you could have a lock at the mailcow compose file since this works in combination with the changed deamon.json

szaimen:
> > I thought enabling ipv6 inside the deamon.json would be enough to make port 443 accessable over ipv6?
> 
> Is that so? Let me check this...

It actually works! Really glad that internal ipv6 support is apparently not needed 😅 

So https://github.com/nextcloud/all-in-one/blob/main/docker-ipv6-support.md was all which made it work

PR #2349, opened 2023-04-15
custom network support

szaimen:
WIP and it doesnt work yet.

Close https://github.com/nextcloud/all-in-one/issues/2348

### TODO:
- [ ] when not specified create nextcloud-aio network as usual, add all containers to it, re-add the mastercontainer to the default bridge network.
- [ ] when specified remove the mastercontainer from the bridge network, add it to the custom network, add all containers to the custom network.


szaimen:
https://github.com/nextcloud/all-in-one/issues/2348#issuecomment-1510459719

PR #2454, opened 2023-05-01
allow to define nextcloud_exec_commands in containers definition

szaimen:
TODO: 
- [x] need to remove the value in yaml-update.sh
- [x] maybe look into https://github.com/nextcloud/all-in-one/pull/2455 and https://github.com/nextcloud/all-in-one/pull/2565 for some ideas on implementation?


php/containers.json
```diff
@@ -217,7 +217,11 @@
           "writeable": true
         }
       ],
-      "restart": "unless-stopped"
+      "restart": "unless-stopped",
+      "nextcloud_exec_commands": [
+        "echo 'Activating collabora config...'",
+        "php /var/www/html/occ richdocuments:activate-config"
```
szaimen, line 223: this will not work in manual-install and co. will need to thing about a solution. Maybe actually as env in containers.json?

PR #2455, opened 2023-05-01
allow to include volumes in backup and restore

szaimen:
TODO:
- [x] need to remove that value from yaml-update.sh
- [x] need to add logic to volume injection of borg container
- [x] hardcode default volumes in backup.sh
- [x] need to deduplicate volumes


Containers/borgbackup/backupscript.sh
```diff
@@ -24,6 +24,15 @@ for directory in "${VOLUME_DIRS[@]}"; do
         exit 1
     fi
 done
+# Test if default volumes are there
+DEFAULT_VOLUMES=(nextcloud_aio_apache nextcloud_aio_nextcloud nextcloud_aio_database nextcloud_aio_database_dump nextcloud_aio_elasticsearch nextcloud_aio_nextcloud_data nextcloud_aio_mastercontainer)
+mapfile -t DEFAULT_VOLUMES <<< "$DEFAULT_VOLUMES"
```
szaimen, line 29:
```suggestion
```

php/src/Docker/DockerActionManager.php
```diff
@@ -503,6 +508,24 @@ public function isAnyUpdateAvailable() : bool {
...
+    private function getBackupVolumes(string $id) : array
+    {
+        $container = $this->containerDefinitionFetcher->GetContainerById($id);
+
+        $backupVolumes = $container->GetBackupVolumes();
+
+        foreach ($container->GetDependsOn() as $dependency) {
+            $backupVolumes[] = $this->getBackupVolumes($dependency);
+        }
+        return $backupVolumes;
```
szaimen, line 520: this does not return a flat array but instead a multi-dimensional one...

PR #2476, opened 2023-05-03
manual-install - enable ipv6 by default

szaimen:


Zoey2936:
does this not crash instances with disabled ipv6?

szaimen:
> does this not crash instances with disabled ipv6?

True, but people need to update manually anyway so they hopefully catch the issue directly and disable ipv6 if they should not have it enabled and just comment this section. 

Additionally ipv6 should nowadays get enabled by default if you ask me...

Zoey2936:
maybe it should at least be somewhere (release notes, manual-install-readme and/or somewhere else) be mentioned?

szaimen:
> maybe it should at least be somewhere (release notes, manual-install-readme and/or somewhere else) be mentioned?

Manual-install readme sounds fine. Let me create a PR for that.

szaimen:
Done with https://github.com/nextcloud/all-in-one/commit/0fdcba047d0a95f749551b0bb88fba280f9562d5

PR #2479, opened 2023-05-03
rework docker network to enable ipv6 correctly

szaimen:
Close https://github.com/nextcloud/all-in-one/issues/2041

### DIFF
--- a/php/src/Docker/DockerActionManager.php
+++ b/php/src/Docker/DockerActionManager.php
@@ -678,10 +678,15 @@ class DockerActionManager
                         'Name' => 'nextcloud-aio',
                         'CheckDuplicate' => true,
                         'Driver' => 'bridge',
-                        'Internal' => false,
-                        'Options' => [
-                            'com.docker.network.bridge.enable_icc' => 'true',
-                            'com.docker.network.bridge.enable_ip_masquerade' => 'true'
+                        // Todo retreiive isipv6enabled from the default network and enable based on this
+                        'EnableIPv6' => $this->isIpv6Enabled() ? true : false,
+                        'IPAM' => [
+                            'Config' => [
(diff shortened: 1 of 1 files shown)


php/src/Docker/DockerActionManager.php
```diff
@@ -678,10 +678,15 @@ private function ConnectContainerIdToNetwork(string $id, string $internalPort) :
...
-                            'com.docker.network.bridge.enable_icc' => 'true',
-                            'com.docker.network.bridge.enable_ip_masquerade' => 'true'
+                        // Todo retreiive isipv6enabled from the default network and enable based on this
+                        'EnableIPv6' => $this->isIpv6Enabled() ? true : false,
+                        'IPAM' => [
+                            'Config' => [
+                                // TODO: retreive the ipv4 subnet from the default network and make sure that it does not overlap
+                                [ "Subnet" => '172.30.1.0/24' ],
+                                // TODO: only add this if ipv6 is enabled and also retreive the ipv6 subnet from the default network and make sure that it does not overlap
+                                [ "Subnet" => '2001:db8:2::/64' ],
```
szaimen, line 688: I fear this will be too prone to failure because we would need to also check if some other network is using this subnet already. So I guess it needs to be a manual config.

Zoey2936: what about using default networks subnets and an option to change them over env?

szaimen: I think https://github.com/nextcloud/all-in-one/pull/2481 is the better way

Zoey2936: but does #2476 not the same?

szaimen: in theory yes but we don't need to add variables and an internal logic for this and instead just promote the official way of creating the docker network

PR #2483, opened 2023-05-03
add internal network config to docker-compose file as well

szaimen:


docker-compose.yml
```diff
@@ -47,3 +47,16 @@ services:
   #     - ./data:/data
   #     - ./sites:/srv
   #   network_mode: "host"
+
+# # Optional: If you need ipv6, follow step 1 and 2 of https://github.com/nextcloud/all-in-one/blob/main/docker-ipv6-support.md first and then uncomment the below config in order to activate ipv6 for the internal nextcloud-aio network
+# networks:
```
szaimen, line 52: do we need to add network:external somewhere or so?

docker-compose.yml
```diff
@@ -47,3 +47,17 @@ services:
   #     - ./data:/data
   #     - ./sites:/srv
   #   network_mode: "host"
+
+# # Optional: If you need ipv6, follow step 1 and 2 of https://github.com/nextcloud/all-in-one/blob/main/docker-ipv6-support.md first and then uncomment the below config in order to activate ipv6 for the internal nextcloud-aio network
+# # Inspired by https://github.com/mailcow/mailcow-dockerized/blob/master/docker-compose.yml
+# networks:
+#   nextcloud-aio:
+#     driver: bridge
```
Zoey2936, line 55:
```suggestion
#     name: nextcloud-aio
#     driver: bridge
```

docker-compose.yml
```diff
@@ -47,3 +47,17 @@ services:
...
+#     driver_opts:
+#       com.docker.network.bridge.name: nextcloud-aio
+#     enable_ipv6: true
+#     ipam:
+#       driver: default
+#       config:
+#         - subnet: ${IPV4_NETWORK:-172.22.1}.0/24
```
Zoey2936, line 62:
```suggestion
```

Zoey2936: can be also removed in manual install

docker-compose.yml
Zoey2936: see: https://docs.docker.com/compose/compose-file/compose-file-v3/#name-1

szaimen: okay! do we want to add external: true to the config or rather not?

docker-compose.yml
szaimen: okay nice!

docker-compose.yml
Zoey2936: I think `external: true` is not needed

szaimen: apparently rather not

docker-compose.yml
```diff
@@ -47,3 +47,17 @@ services:
...
   #   network_mode: "host"
+
+# # Optional: If you need ipv6, follow step 1 and 2 of https://github.com/nextcloud/all-in-one/blob/main/docker-ipv6-support.md first and then uncomment the below config in order to activate ipv6 for the internal nextcloud-aio network
+# # Inspired by https://github.com/mailcow/mailcow-dockerized/blob/master/docker-compose.yml
+# networks:
+#   nextcloud-aio:
+#     name: nextcloud-aio
+#     driver: bridge
+#     driver_opts:
+#       com.docker.network.bridge.name: nextcloud-aio
```
Zoey2936, line 58: is this needed?

szaimen: probably not

Zoey2936:
```suggestion
```

Zoey2936:
can you maybe also add these suggestions to the manual install network config?

Zoey2936 (approved):

szaimen:
> can you maybe also add these suggestions to the manual install network config?

done with https://github.com/nextcloud/all-in-one/commit/beffb879dc45a74e5f9ad3e8ce8745bd8cfb57d4

szaimen:
Thanks for the review! :)

Zoey2936:
one problem:
- docker-compose only creates the network if at least one service in the container uses the network

Zoey2936:
and:
- maybe the default recommended network should not be the same as mailcows

Zoey2936:
so the mastercontainer would maybe need:
```
    network_mode: bridge
    networks:
      - nextcloud-aio
```
and all compose files (normal & manual install) should use the same ipv6 subnet as the ipv6 docs

szaimen:
> * docker-compose only creates the network if at least one service in the container uses the network

I see. So we would need to add it possinly commented to the mastercontainer. Do you mind creating a PR with that?

> maybe the default recommended network should not be the same as mailcows

What do you mean?

szaimen:
Ah yes, probably we should change that...

Zoey2936:
and something other, does the ipv6 docs use public ipv6 adresses?

szaimen:
Apparently to be private it needs to start with fd:
https://en.m.wikipedia.org/wiki/Private_network#IPv6
Screenshot_20230503_192303_Brave

szaimen:
Lets use fd12:3456:789a:1::/64 and 	fd12:3456:789a:2::/64 then

PR #2493, opened 2023-05-03
optimize collabora Dockerfile

Zoey2936:


szaimen (changes requested):
I would keep it like this

Zoey2936:
and why?

szaimen:
becaause DEBIAN_FRONTEND is alreaddy set during hte command so we dont need another arg right?

Zoey2936:
> becaause DEBIAN_FRONTEND is alreaddy set during hte command so we dont need another arg right?

yes, but I prefer using ARG for settings envs while building, but if you want you can close this PR

szaimen:
> > becaause DEBIAN_FRONTEND is alreaddy set during hte command so we dont need another arg right?
> 
> yes, but I prefer using ARG for settings envs while building, but if you want you can close this PR

I see but I think it is not necessary so I would close it

PR #2495, opened 2023-05-03
optimze elastic Dockerfile

Zoey2936:


Containers/fulltextsearch/Dockerfile
```diff
@@ -1,15 +1,15 @@
 # Probably from here https://github.com/elastic/elasticsearch/blob/main/distribution/docker/src/docker/Dockerfile
 FROM elasticsearch:7.17.10
 
-RUN elasticsearch-plugin install --batch ingest-attachment
-
+ARG DEBIAN_FRONTEND=noninteractive
 RUN set -ex; \
     \
     apt-get update; \
```
szaimen, line 7:
```suggestion
    export DEBIAN_FRONTEND=noninteractive; \
    apt-get update; \
```

Containers/fulltextsearch/Dockerfile
```diff
@@ -1,15 +1,15 @@
 # Probably from here https://github.com/elastic/elasticsearch/blob/main/distribution/docker/src/docker/Dockerfile
 FROM elasticsearch:7.17.10
 
-RUN elasticsearch-plugin install --batch ingest-attachment
-
+ARG DEBIAN_FRONTEND=noninteractive
```
szaimen, line 4:
```suggestion
```

Zoey2936: why not using ARG?

szaimen: because it is another step which could be done in the run step, no?

Zoey2936: using ARG/export don't change the end result, but I prefer ARG, because then I have a list of build time envs at the top of the file

szaimen (approved):
LGTM, thanks!

PR #2504, opened 2023-05-03
improve ipv6 docs

Zoey2936:


Zoey2936:
I've tried the compose file and it works:
grafik

docker-compose.yml
```diff
@@ -16,6 +16,9 @@ services:
       - 80:80 # Can be removed when running behind a web server or reverse proxy (like Apache, Nginx and else). See https://github.com/nextcloud/all-in-one/blob/main/reverse-proxy.md
       - 8080:8080
       - 8443:8443 # Can be removed when running behind a web server or reverse proxy (like Apache, Nginx and else). See https://github.com/nextcloud/all-in-one/blob/main/reverse-proxy.md
+    # network_mode: bridge # Is needed when you want to create the nextcloud-aio network usinh this file
```
Zoey2936, line 19: not sure if this line is needed, it works with it and I use it in most of my compose files

szaimen: I would not add this here

Zoey2936: do you mean below the envs?

Zoey2936: or do you mean the network_mode line?

szaimen: Network_mode should get removed, the rest I would indeed move below the envs

Zoey2936: done

szaimen:
Maybe also add a hint to the network config that to get enabled the nextcloud-aio network needs to get uncommented above?

Also probably we should state that the name of the network should not be changed

Zoey2936:
> Maybe also add a hint to the network config that to get enabled the nextcloud-aio network needs to get uncommented above?

Do you mean the services or the networks part? because the I've added something like this to the networks part
> 
> Also probably we should state that the name of the network should not be changed

I will add it

docker-compose.yml
```diff
@@ -33,6 +33,8 @@ services:
       # - NEXTCLOUD_ENABLE_DRI_DEVICE=true # This allows to enable the /dev/dri device in the Nextcloud container which is needed for hardware-transcoding. See https://github.com/nextcloud/all-in-one#how-to-enable-hardware-transcoding-for-nextcloud
       # - TALK_PORT=3478 # This allows to adjust the port that the talk container is using.
       # - WATCHTOWER_DOCKER_SOCKET_PATH=/var/run/docker.sock # Needs to be specified if the docker socket on the host is not located in the default '/var/run/docker.sock'. Otherwise mastercontainer updates will fail. For macos it needs to be '/var/run/docker.sock'
+    # networks: # Is needed when you want to create the nextcloud-aio network using this file
+      # - nextcloud-aio # Is needed when you want to create the nextcloud-aio network using this file
```
szaimen, line 37:
```suggestion
    # networks: # Is needed when you want to create the nextcloud-aio network using this file, see the network config at the bottom of the file
      # - nextcloud-aio # Is needed when you want to create the nextcloud-aio network using this file, see the network config at the bottom of the file
```

docker-compose.yml
```diff
@@ -48,14 +50,14 @@ services:
   #     - ./sites:/srv
   #   network_mode: "host"
 
-# # Optional: If you need ipv6, follow step 1 and 2 of https://github.com/nextcloud/all-in-one/blob/main/docker-ipv6-support.md first and then uncomment the below config in order to activate ipv6 for the internal nextcloud-aio network
+# # Optional: If you need ipv6, follow step 1 and 2 of https://github.com/nextcloud/all-in-one/blob/main/docker-ipv6-support.md first and then uncomment the below config in order to activate ipv6 for the internal nextcloud-aio network. Please make sure to uncomment also the networking lines of the mastercontainer above.
```
szaimen, line 53:
```suggestion
# # Optional: If you need ipv6, follow step 1 and 2 of https://github.com/nextcloud/all-in-one/blob/main/docker-ipv6-support.md first and then uncomment the below config in order to activate ipv6 for the internal nextcloud-aio network.
# # Please make sure to uncomment also the networking lines of the mastercontainer above in order to actually create the network with docker-compose
```

docker-compose.yml
```diff
@@ -48,14 +50,14 @@ services:
...
 # # Inspired by https://github.com/mailcow/mailcow-dockerized/blob/master/docker-compose.yml
 # networks:
 #   nextcloud-aio:
-#     name: nextcloud-aio
+#     name: nextcloud-aio # This line is not allowed to be changed
```
szaimen, line 57:
```suggestion
#     name: nextcloud-aio # This line is not allowed to be changed as otherwise the created network will not be used by the other containers of AIO
```

docker-ipv6-support.md
```diff
@@ -3,12 +3,12 @@
 Before you can use IPv6 in Docker containers or swarm services, you need to enable IPv6 support in the Docker daemon. Afterward, you can choose to use either IPv4 or IPv6 (or both) with any container, service, or network.
 
 ## Docker on Linux and Docker-rootless
-1.  Edit `/etc/docker/daemon.json` (or `~/.config/docker/daemon.json` in case of docker-rootless), set the `ipv6` key to `true` and the `fixed-cidr-v6` key to your IPv6 subnet. In this example we are setting it to `2001:db8:1::/64`. Additionally set `experimental` to `true` and `ip6tables` to `true` as well.
+1.  Edit `/etc/docker/daemon.json` (or `~/.config/docker/daemon.json` in case of docker-rootless), set the `ipv6` key to `true` and the `fixed-cidr-v6` key to your IPv6 subnet. In this example we are setting it to `fd12:3456:789a:1::/64`. Additionally set `experimental` to `true` and `ip6tables` to `true` as well. If you enabled IPv6 with the update.sh from mailcow, you can keep their daemon.json, it will work too.
```
szaimen, line 6:
```suggestion
1.  Edit `/etc/docker/daemon.json` (or `~/.config/docker/daemon.json` in case of docker-rootless), set the `ipv6` key to `true` and the `fixed-cidr-v6` key to your IPv6 subnet. In this example we are setting it to `fd12:3456:789a:1::/64`. Additionally set `experimental` to `true` and `ip6tables` to `true` as well. If you are using mailcow and enabled IPv6 with the update.sh, you can keep their daemon.json, it will work too.
```

docker-ipv6-support.md
```diff
@@ -21,22 +21,22 @@ Before you can use IPv6 in Docker containers or swarm services, you need to enab
...
 ## Docker Desktop (Windows and macOS)
 On Windows and macOS which use Docker Desktop, you need to go into the settings, and select `Docker Engine`. There you should see the currently used daemon.json file. 
 
-1. You need to now adjust this json file by setting the `ipv6` key to `true` and the `fixed-cidr-v6` key to your IPv6 subnet. In this example we are setting it to `2001:db8:1::/64`. Additionally set `experimental` to `true` and `ip6tables` to `true` as well.
+1. You need to now adjust this json file by setting the `ipv6` key to `true` and the `fixed-cidr-v6` key to your IPv6 subnet. In this example we are setting it to `fd12:3456:789a:1::/64`. Additionally set `experimental` to `true` and `ip6tables` to `true` as well.
 
     ```
     "ipv6": true,
-    "fixed-cidr-v6": "2001:db8:1::/64",
+    "fixed-cidr-v6": "fd12:3456:789a:1::/64s",
```
szaimen, line 33:
```suggestion
    "fixed-cidr-v6": "fd12:3456:789a:1::/64",
```

szaimen:
> I've tried the compose file (and mailcow daemon.json) and it works:

Nice 🎉

Thanks for testing! :)

docker-compose.yml
```diff
@@ -33,6 +33,8 @@ services:
       # - NEXTCLOUD_ENABLE_DRI_DEVICE=true # This allows to enable the /dev/dri device in the Nextcloud container which is needed for hardware-transcoding. See https://github.com/nextcloud/all-in-one#how-to-enable-hardware-transcoding-for-nextcloud
       # - TALK_PORT=3478 # This allows to adjust the port that the talk container is using.
       # - WATCHTOWER_DOCKER_SOCKET_PATH=/var/run/docker.sock # Needs to be specified if the docker socket on the host is not located in the default '/var/run/docker.sock'. Otherwise mastercontainer updates will fail. For macos it needs to be '/var/run/docker.sock'
+    # networks: # Is needed when you want to create the nextcloud-aio network using this file, see the network config at the bottom of the file
+      # - nextcloud-aio # Is needed when you want to create the nextcloud-aio network using this file, see the network config at the bottom of the file
```
szaimen, line 37:
```suggestion
    # networks: # Is needed when you want to create the nextcloud-aio network with ipv6-support using this file, see the network config at the bottom of the file
      # - nextcloud-aio # Is needed when you want to create the nextcloud-aio network with ipv6-support using this file, see the network config at the bottom of the file
```

szaimen (approved):
LGTM, thanks! :)

philwo:
Thanks for improving the docs for IPv6!

I'm curious about one part of this change: The prefix `2001:db8::` is the correct one to use for documentation and examples according to [RFC 3849](https://datatracker.ietf.org/doc/html/rfc3849).

Why did we change it to `fd12:3456:789a:1::`, which is a valid IPv6 unique local address?

I'm not too familiar with how Docker uses IPv6 and how to configure it correctly - does that mean we *have* to use a site-local address here and then setup a reverse proxy to make Nextcloud reachable via the internet? 

Or would it be OK to specify the globally reachable subnet already assigned to the host? If that even works, would it mean that all containers are accidentally reachable via the internet unless I put a firewall in place?

Zoey2936:
ULAs are from my perspective private IPs under IPv6 and because we want natting, we need to use ULAs

PR #2519, opened 2023-05-08
add root user check for mastercontainer

Zoey2936:


Containers/mastercontainer/start.sh
```diff
@@ -20,6 +20,11 @@ case "${1}" in
 esac
 }
 
+if [ "$(whoami)" != "root" ] || [ "$(id -u)" != "0" ] || [ "$(id -g)" != "0" ]; then
+    print_red "Container does not run as root. Cannot continue."
```
szaimen, line 24:
```suggestion
# Check if running as root user
if [ "$EUID" != "0" ]; then
    print_red "Container does not run as root user. This is not supported."
```

Zoey2936: I've added this changed and I didn't see the change of the if condition, is EUID even a valid env?

szaimen: Yes, it is. Just try it :)

Zoey2936: what is there the difference to id -u / id -g / whoami?

szaimen: Both would work but using the variable seems cleaner to me :)

Zoey2936: but what if the group is different?

szaimen: I think that doesnt matter since you have root permissions anyway?

Zoey2936: not sure...

szaimen (approved):

Issue #2566, opened 2023-05-19
apache does not start after v5.2.0 Update


gerhard-thompson:
### Steps to reproduce
1. upgrade from v5.1.0 to v5.2.0 with APACHE_PORT different than 443

### Expected behavior 
apache should start

### Actual behavior 
apache does not start

### Host OS 
Ubuntu 22.04

#### Nextcloud AIO version 
v5.2.0

#### Current channel 
beta

#### Other valuable info 
Error Message:
`Waiting for Nextcloud to start...
Waiting for Nextcloud to start...
Waiting for Nextcloud to start...
{"level":"info","ts":1684516223.1517866,"msg":"using provided configuration","config_file":"/Caddyfile","config_adapter":""}
Error: adapting config using caddyfile: parsing caddyfile tokens for 'servers': /Caddyfile:9 - Error during parsing: getting module named 'http.ip_sources.private_ranges': module not registered: http.ip_sources.private_ranges
[Fri May 19 17:10:23.161254 2023] [mpm_event:notice] [pid 37:tid 139855306758984] AH00489: Apache/2.4.57 (Unix) configured -- resuming normal operations
[Fri May 19 17:10:23.161281 2023] [core:notice] [pid 37:tid 139855306758984] AH00094: Command line: '/usr/local/apache2/bin/httpd -D FOREGROUND'
{"level":"info","ts":1684516224.19098,"msg":"using provided configuration","config_file":"/Caddyfile","config_adapter":""}
Error: adapting config using caddyfile: parsing caddyfile tokens for 'servers': /Caddyfile:9 - Error during parsing: getting module named 'http.ip_sources.private_ranges': module not registered: http.ip_sources.private_ranges
{"level":"info","ts":1684516226.2237773,"msg":"using provided configuration","config_file":"/Caddyfile","config_adapter":""}
Error: adapting config using caddyfile: parsing caddyfile tokens for 'servers': /Caddyfile:9 - Error during parsing: getting module named 'http.ip_sources.private_ranges': module not registered: http.ip_sources.private_ranges
{"level":"info","ts":1684516229.2550886,"msg":"using provided configuration","config_file":"/Caddyfile","config_adapter":""}
Error: adapting config using caddyfile: parsing caddyfile tokens for 'servers': /Caddyfile:9 - Error during parsing: getting module named 'http.ip_sources.private_ranges': module not registered: http.ip_sources.private_ranges
`

my startup command:
[...]

PR #2569, opened 2023-05-19
caddy - a bugfix and some improvements

szaimen:
Fix https://github.com/nextcloud/all-in-one/issues/2566


szaimen:
Todo: think about leaving nextcloud in internal dir and not copying it to the import dir. Then maybe also create an empty file in therr in order to sipence the warning?

arunas-smala (approved):
it works

DO6FP (approved):
Fixes the issue #2566 successfully.

Zoey2936:
> Or maybe revert back to one caddyfile and one import statement?

I think that would be the best way, from my perspective, but beside this, do you think, you can release this patch today?

szaimen:
> I think that would be the best way, from my perspective, but beside this, do you think, you can release this patch today?

Thanks for the feedback! I'll try...

PR #2576, opened 2023-05-20
enable compression for js files in Nextcloud

szaimen:
Close https://github.com/nextcloud/all-in-one/discussions/2574

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -28,7 +28,7 @@ Included are:
 - PHP and web server timeouts set to 3600s, [adjustable](https://github.com/nextcloud/all-in-one#how-to-adjust-the-max-execution-time-for-nextcloud) (important for big file uploads)
 - Defaults to a max of 512 MB RAM per PHP process, [adjustable](https://github.com/nextcloud/all-in-one#how-to-adjust-the-php-memory-limit-for-nextcloud)
 - Automatic TLS included (by using Let's Encrypt)
-- Brotli compression enabled by default for javascript files which reduces load times
+- Brotli compression enabled by default for javascript files which reduces Nextcloud load times
 - HTTP/2 and HTTP/3 enabled
 - "Pretty URLs" for Nextcloud are enabled by default (removes the index.php from all links)
 - Video previews work out of the box and when Imaginary is enabled, many recent image formats as well!
--- a/readme.md
+++ b/readme.md
@@ -28,6 +28,7 @@ Included are:
 - PHP and web server timeouts set to 3600s, [adjustable](https://github.com/nextcloud/all-in-one#how-to-adjust-the-max-execution-time-for-nextcloud) (important for big file uploads)
 - Defaults to a max of 512 MB RAM per PHP process, [adjustable](https://github.com/nextcloud/all-in-one#how-to-adjust-the-php-memory-limit-for-nextcloud)
 - Automatic TLS included (by using Let's Encrypt)
+- Brotli compression enabled by default for javascript files which reduces load times
 - HTTP/2 and HTTP/3 enabled
 - "Pretty URLs" for Nextcloud are enabled by default (removes the index.php from all links)
 - Video previews work out of the box and when Imaginary is enabled, many recent image formats as well!
--- a/Containers/apache/Dockerfile
+++ b/Containers/apache/Dockerfile
@@ -45,6 +45,7 @@ RUN set -ex; \
             -e 's/^#\(LoadModule .*mod_authz_core.so\)/\1/' \
             -e 's/^#\(LoadModule .*mod_alias.so\)/\1/' \
             -e 's/^#\(LoadModule .*mod_mpm_event.so\)/\1/' \
+            -e 's/^#\(LoadModule .*mod_brotli.so\)/\1/' \
             -e 's/\(LoadModule .*mod_mpm_worker.so\)/#\1/' \
             -e 's/\(LoadModule .*mod_mpm_prefork.so\)/#\1/' \
         /usr/local/apache2/conf/httpd.conf; \
--- a/Containers/apache/nextcloud.conf
+++ b/Containers/apache/nextcloud.conf
@@ -10,6 +10,13 @@ Listen 8000
     <FilesMatch "\.php$">
         SetHandler "proxy:fcgi://${NEXTCLOUD_HOST}:9000"
     </FilesMatch>
+
+    # Enable Brotli compression for js files
+    <IfModule mod_brotli.c>
+        AddOutputFilterByType BROTLI_COMPRESS text/javascript application/javascript application/x-javascript
+        BrotliCompressionQuality 0
+    </IfModule>
+
     # Nextcloud dir
     DocumentRoot /var/www/html/
     <Directory /var/www/html/>


Containers/apache/Caddyfile
```diff
@@ -59,6 +59,13 @@
         reverse_proxy localhost:8000
     }
 
+    # Enable content encoding for javascript files
+    encode zstd gzip {
+        match {
+            header Content-Type application/javascript*
+        }
+    }
+
```
szaimen, line 68: @Zoey2936 unfortunately this does not work inside a route block where I would want to enable this specifically to keep the possible side-effects as low as possible. So I need to enable this globally unfortunately. 

Can you please check if this also works behind NPM or if we need to disable this when AIO is running behind a reverse proxy? Thank you!

Zoey2936: I've added `encode zstd gzip` above `# Collabora`, restartet the nextcloud-aio-apache container, checked if compression worked with `curl -L http://127.0.0.1:11000/login --header "Host: cloud.comain.de"` (it worked) and then tried using firefox to access the cloud behind my npm and yes it worked

Zoey2936:
but why do you only want to enable it for JavaScript files?

szaimen:
> but why do you only want to enable it for JavaScript files?

Because I don't want to risk any possible side-effects. E.g. I could imagine that you might run into problems when the desktop client is not prepared for this kind of encoding on webdav responses and I actually also don't want to find out currently XD

Containers/apache/Caddyfile
szaimen: Nice! Thanks for verifying! Did you make sure to clean the browser cache after applying the setting?

Zoey2936:
> > but why do you only want to enable it for JavaScript files?
> 
> Because I don't want to risk any possible side-effects. E.g. I could imagine that you might run into problems when the desktop client is not prepared for this kind of encoding on webdav responses and I actually also don't want to find out currently XD

Not sure... compression (gzip and also brotli) is basically a default in the internet

Zoey2936:
But beside this, what about the mastercontainer and/or enabling compression in apache (mastercontainer/apache conatiner)?

szaimen:
> But beside this, what about the mastercontainer and/or enabling compression in apache (mastercontainer/apache conatiner)? https://httpd.apache.org/docs/2.4/mod/mod_deflate.html https://httpd.apache.org/docs/2.4/mod/mod_brotli.html (both are inside the httpd:alpine image, not sure if they are inside the alpine package repo)

Since the mastercontainer uses mostly server-side rendering, I don't think that it needs compression. 

Enabling compression in Apache would only make sense if it comes with an additional size benefit but since gzip already reduces the size from 13.3MB to 3.5 MB, I don't think there is a lot of benefit using Apache for content-encoding.

Containers/apache/Caddyfile
Zoey2936: I've tried in Firefox "private" mode, I think a new private season has no cache, if I am right

szaimen:
Just verified, brotli even decreses it to 2.0 MB so it makes sense. Also since we enable this in apache, it only enables it for Nextcloud which should limit side-effects.

Zoey2936:
do you only want to add brotli support? maybe also support gzip/deflate as fallback

Containers/apache/Caddyfile
szaimen: private mode has indeed no cache. Thanks for verifying! :)

szaimen:
> do you only want to add brotli support? maybe also support gzip/deflate as fallback

How would I do that?

Zoey2936:
Enable the deflate module and use the same configuartion, but for deflate/gzip, but maybe brotli is enough, all modern browsers support it now: https://caniuse.com/brotli

szaimen:
> but maybe brotli is enough, all modern browsers support it now: https://caniuse.com/brotli

indeed, nice!

Zoey2936:
Another question, why does nextcloud not provide precompressed (gzip/brotli) files? that would be nice, since then the webserver does not need to compress files and can just sent the precompressed files, the only thing is that they are in a binary format?

szaimen:
> Another question, why does nextcloud not provide precompressed (gzip/brotli) files? that would be nice, since then the webserver does not need to compress files and can just sent the precompressed files, the only thing is that they are in a binary format?

Asked this already in https://github.com/nextcloud/server/issues/36728#issuecomment-1556038420

szaimen:
Apparentl, brotly compression on the fly is slow: https://github.com/nextcloud/server/issues/36728#issuecomment-1556068213 so I guess reverting back to gzip is the way to go for now.

Zoey2936:
Not sure: https://blog.cloudflare.com/results-experimenting-brotli even the lowest compression level of brotli (0) seems to be better than the strongest of gzip and has the same speed then gzip level 4, in my NPM fork I decided to use gzip level 1 and brotli level 0.

szaimen:
Apparently the problem is rather that compressing on the fly on the server side is pretty slow. Better would be if they would be pre-compressed. However gzip compression on the fly is faster iirc.

szaimen:
Ill make some tests the coming days.

Zoey2936:
what compression level did you used?

szaimen:
> what compression level did you used?

4, the same like configured in this PR

Zoey2936:
can you try with a compression level of 0 (brotli) and 1 (gzip), I got with the lowest level the best result, because higher level could cause high cpu usage, if there many requests at the same time

szaimen:
> can you try with a compression level of 0 (brotli) and 1 (gzip), I got with the lowest level the best result, because higher level could cause high cpu usage, if there many requests at the same time

brotli 2.5s, 3.4 MB
deflate 2.7s, 3.7 MB

So looks like brotli on level 0 is the way to go then :)

Zoey2936 (approved):

readme.md
```diff
@@ -28,6 +28,7 @@ Included are:
 - PHP and web server timeouts set to 3600s, [adjustable](https://github.com/nextcloud/all-in-one#how-to-adjust-the-max-execution-time-for-nextcloud) (important for big file uploads)
 - Defaults to a max of 512 MB RAM per PHP process, [adjustable](https://github.com/nextcloud/all-in-one#how-to-adjust-the-php-memory-limit-for-nextcloud)
 - Automatic TLS included (by using Let's Encrypt)
+- Brotli compression enabled by default for javascript files which reduces load times
```
szaimen, line 31:
```suggestion
- Brotli compression enabled by default for javascript files which reduces Nextcloud load times
```

Zoey2936:
> > > but why do you only want to enable it for JavaScript files?
> > 
> > 
> > Because I don't want to risk any possible side-effects. E.g. I could imagine that you might run into problems when the desktop client is not prepared for this kind of encoding on webdav responses and I actually also don't want to find out currently XD
> 
> Not sure... compression (gzip and also brotli) is basically a default in the internet

And also will compression only be done, if the client sends the `Accept-Encoding` HTTP header, which needs to containe "br".

Zoey2936:
do you know if there is something, with providing precompressed files? If no waht about compressing as a build step in the dockerfile? Ans also to my last comment above, what about enabling compression for more mime-types?

szaimen:
> If no waht about compressing as a build step in the dockerfile? 

I guess this will lead to a failed integrity check in Nextcloud so this is no option.

> Ans also to my last comment above, what about enabling compression for more mime-types?

I heard something about desktop client or other software bugging out when compresisng stuff. Js files seemed to be by far the largest assets so it makes the biggest impact.

Zoey2936:
> I heard something about desktop client or other software bugging out when compresisng stuff. Js files seemed to be by far the largest assets so it makes the biggest impact.

but why should they bug? They only will get compressed files, if they send the accept-encoding header...

szaimen:
So if you already verified that there are no problems with the client applications, then I would be fine to enable this for further file types. Which would you recommend and can you create a PR?

Zoey2936:
If you try these two commands:
```
curl -LH 'Accept-encoding: br' https://cloud.domain.de/dist/core-common.js
curl -L https://cloud.domain.de/dist/core-common.js
```
You can see that compression will only be done if the client requests it, so it can't cause problems...
On my NPM, I use Brotli on all mime-types https://github.com/ZoeyVid/nginx-proxy-manager/blob/397e79eff9dae022ab4d616f3d60f20f6d18539f/rootfs/usr/local/nginx/conf/conf.d/include/brotli.conf#L2. And I had never problems with it (Nextcloud and all other websites proxies by it). Maybe html/json/css/php/usw. (or all) could be enabled?

szaimen:
I saw that some assets might be compressed already on the server side (works for css out of the box as I found out) if gzip encoding is accepted. See https://github.com/search?q=repo%3Anextcloud%2Fserver+gzip+language%3APHP&type=code&l=PHP.
Would we not risk double-encoding in that case which would slow-down the request instead of making it faster?

If no, would the following mimetypes be fine? `text/html text/plain text/xml text/css text/javascript application/javascript application/x-javascript application/json application/xml` as recommended in https://www.brotli.pro/enable-brotli/servers/apache/#configure-brotli-compression?

Zoey2936:
maybe also php?

szaimen:
I think php files are never send to the clients, no? (php gets executed and interpreted on the server)

Zoey2936:
> I think php files are never send to the clients, no? (php gets executed and interpreted on the server)

oh yes, your correct

Zoey2936:
tested, yes brotli and gzip on css gets compressed, so we don't need to compress css

szaimen:
Can you also check propfind xml responses? They should also be compressed by default IIRC.

Zoey2936:
no idea where a xml file is located for testing

szaimen:
If yes, only these are valid iirc: text/html text/plain text/javascript application/javascript application/x-javascript application/json

szaimen:
image

szaimen:
So xml seems to be encoded as well

szaimen:
image

szaimen:
text/html as well

Zoey2936:
but why js not?

szaimen:
`image/svg+xml` is also not

szaimen:
image

szaimen:
> but why js not?

good question. Cannot tell you why

szaimen:
So after testing a bit, everything seems to be compressed by default except js files and image/svg+xml files

szaimen:
Actually text/css is also not compressed in my testing. I thoughts so but it is not the case.

szaimen:
See https://github.com/nextcloud/all-in-one/pull/2793

szaimen:
> but why js not?

I think the reason lies here: https://github.com/nextcloud/server/blob/bd3f3afcf715ee283f61a45e4a26e566f949a7a5/lib/private/AppFramework/Middleware/CompressionMiddleware.php#L64-L78

Zoey2936:
https://github.com/nextcloud/server/blob/bd3f3afcf715ee283f61a45e4a26e566f949a7a5/lib/private/Template/JSCombiner.php#L197 does this mean it is compressed using level 9? Is level 1 not better?

Zoey2936:
and do you know if brotli is also supported by nc itself?

szaimen:
> https://github.com/nextcloud/server/blob/bd3f3afcf715ee283f61a45e4a26e566f949a7a5/lib/private/Template/JSCombiner.php#L197 does this mean it is compressed using level 9?

yes

szaimen:
> and do you know if brotli is also supported by nc itself?

I dont think so

Zoey2936:
> > https://github.com/nextcloud/server/blob/bd3f3afcf715ee283f61a45e4a26e566f949a7a5/lib/private/Template/JSCombiner.php#L197 does this mean it is compressed using level 9?
> 
> yes

should it maybe be set down to level 1?

szaimen:
> should it maybe be set down to level 1?

Possibly but it does not look like it is even used anymore....

Zoey2936:
I thought nc does compression by itself under some conditions?

szaimen:
In appdata, I only found these files:
image

szaimen:
It does not seem to be used apart from that...

Zoey2936:
https://github.com/nextcloud/server/blob/bd3f3afcf715ee283f61a45e4a26e566f949a7a5/lib/private/AppFramework/Middleware/CompressionMiddleware.php#L88
https://github.com/nextcloud/server/blob/bd3f3afcf715ee283f61a45e4a26e566f949a7a5/apps/dav/lib/Connector/Sabre/PropfindCompressionPlugin.php#L68
and maybe the compression level could be set lower in these lines?

szaimen:
Actually I think compression level 9 is okay for the jscombiner since it is only created once and then cached IIRC. However for the others there should indeed be a low compression level to keep the load low.

Zoey2936:
I think this should be done, and it could also improve the performance a bit

szaimen:
As for the responses I think level 6 is used if not specified...

szaimen:
> I think this should be done, and it could also improve the performance a bit

you mean for the compressionmiddleware and compressionplugin?

szaimen:
Which level would you recommend for these two?

Zoey2936:
> As for the responses I think level 6 is used if not specified...

yes, default seems to be 6

Zoey2936:
> > I think this should be done, and it could also improve the performance a bit
> 
> you mean for the compressionmiddleware and compressionplugin?

yes

Zoey2936:
> Which level would you recommend for these two?

not sure, needs testing, but I think it can be the lowest/a very low value

szaimen:
So level 1 for example? level 0 is no compression IIRC

szaimen:
However I read this in the zlib manual: `Z_DEFAULT_COMPRESSION requests a default compromise between speed and compression (currently equivalent to level 6).`
reducing to level 1 will reduce the load on the server but obviously also increase the size and load time. So not sure if it really makes sense to adjust the level?

szaimen:
I've just found this btw: 
image
https://github.com/szaimen/TurboBench/tree/patch-1#benchmark

Zoey2936:
> However I read this in the zlib manual: `Z_DEFAULT_COMPRESSION requests a default compromise between speed and compression (currently equivalent to level 6).` reducing to level 1 will reduce the load on the server but obviously also increase the size and load time. So not sure if it really makes sense to adjust the level?


that is the reason why it think it should be tested

philiprenich:
Hey, should I be seeing gzip'd assets coming through with the latest version of AIO? Status is merged, but I'm not sure about release cycles. I'm on AIO v7.2.1 which is Nextcloud Hub 6 and version 27.1.0
Would love to not have to download nearly 50mb for my dashboard 😅

PR #2580, opened 2023-05-22
Update reverse-proxy.md

craigkh:
I suggest adding the name and location of Caddyfile as there is no instructions for how to create a Caddyfile.
Failure to create a Caddyfile prior to starting the container results in a directory called Caddyfile being created.


szaimen (approved):
LGTM, thanks!

PR #2601, opened 2023-05-26
set NEXTCLOUD_MOUNT to rshared

szaimen:
Close https://github.com/nextcloud/all-in-one/discussions/2585

Address https://github.com/nextcloud/all-in-one/discussions/2927?converting=1

### DIFF
--- a/php/src/Docker/DockerActionManager.php
+++ b/php/src/Docker/DockerActionManager.php
@@ -211,9 +211,16 @@ class DockerActionManager
 
     public function CreateContainer(Container $container) : void {
         $volumes = [];
-        foreach($container->GetVolumes()->GetVolumes() as $volume) {
+        foreach ($container->GetVolumes()->GetVolumes() as $volume) {
+            // NEXTCLOUD_MOUNT gets added via bind-mount later on
+            if ($container->GetIdentifier() === 'nextcloud-aio-nextcloud') {
+                if ($volume->name === $this->configurationManager->GetNextcloudMount()) {
+                    continue;
+                }
+            }
+
             $volumeEntry = $volume->name . ':' . $volume->mountPoint;
-            if($volume->isWritable) {
+            if ($volume->isWritable) {
                 $volumeEntry = $volumeEntry . ':' . 'rw';
             } else {
                 $volumeEntry = $volumeEntry . ':' . 'ro';
@@ -226,7 +233,7 @@ class DockerActionManager
             'Image' => $this->BuildImageName($container),
         ];
 
-        if(count($volumes) > 0) {
+        if (count($volumes) > 0) {
             $requestBody['HostConfig']['Binds'] = $volumes;
         }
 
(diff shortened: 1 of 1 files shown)


php/src/Docker/DockerActionManager.php
```diff
@@ -425,10 +409,18 @@ public function CreateContainer(Container $container) : void {
             $requestBody['HostConfig']['SecurityOpt'] = ["apparmor:unconfined"];
         }
 
+        $mounts = [];
+        foreach($container->GetVolumes()->GetVolumes() as $volume) {
+            if (!str_starts_with($volume->name, '/')) {
+                $mounts[] = ["Type" => "volume", "Source" => $volume->name, "Target" => $volume->mountPoint, "ReadOnly" => !$volume->isWritable];
+            } else {
+                $mounts[] = ["Type" => "bind", "Source" => $volume->name, "Target" => $volume->mountPoint, "ReadOnly" => !$volume->isWritable, "Propagation" => "rshared"];
```
szaimen, line 417: This means it will bug out if a directory is not existing irrc. need to test this.

Also test if this works on docker desktop

szaimen, line 417: Should try if that works and creates not existing dirs as well
```suggestion
                $mounts[] = ["Type" => "volume", "Source" => $volume->name, "Target" => $volume->mountPoint, "ReadOnly" => !$volume->isWritable, "Propagation" => "rshared"];
```

szaimen: Or maybe just check first if it creates nor existing dirs first

szaimen: indeed this is not going to work.

szaimen: @chatziko we have a problem here. using mounts instead of volumes will bug out if you specify a directory as target that is not existing yet which we definitely don't want to happen. So we can either have this behaviour kept of not bugging out in that case or setting the propagation to rshared. Not sure how to solve this. WDYT?

chatziko: So from what I understand (I did a few tests myself), the mount directory inside the container is always created. But if the one in the host does not exist:
- `--mount type=bind,source=/foo,target=/foo` will fail with an error
- `--volume /foo:/foo` will create an empty dir `/foo` in the host

Here's an idea: if `--mount` fails, catch the error and re-run the container with `--volume`.
- This will be completely transparent to the user.
- It will only happen the first time, since on future restarts the directory will exist (it gets created).

Another option would be to add a new env variable that controls the propagation, and explain that it can only be set if the mount dir exists.

szaimen: > if `--mount` fails, catch the error and re-run the container with `--volume`.

I fear this is not really possible as checking for that specific error is not going to work....

Actually I am also not sure about an additional config value for this. Will need to think about a good solution to this.

szaimen:
Hm... maybe the documentation is actually wrong? (would need to be tested)
https://github.com/docker/for-win/issues/11958#issuecomment-1406555066

szaimen:
One possible way would be to only use this logic for NEXTCLOUD_MOUNT and not for others.

reply-nick:
@szaimen after this change, I am getting this error:
```Message: Server error: `POST http://localhost/v1.41/containers/nextcloud-aio-nextcloud/start` resulted in a `500 Internal Server Error` response:
{&quot;message&quot;:&quot;path /mnt is mounted on / but it is not a shared mount&quot;}```

my data folder is on a NFS share. 
Are there any workarounds for this?

PR #2611, opened 2023-05-26
validate json against json schema

szaimen:
For https://github.com/nextcloud/all-in-one/discussions/1581

### DIFF
--- a/php/src/ContainerDefinitionFetcher.php
+++ b/php/src/ContainerDefinitionFetcher.php
@@ -12,6 +12,7 @@ use AIO\Container\State\RunningState;
 use AIO\Data\ConfigurationManager;
 use AIO\Data\DataConst;
 use AIO\Docker\DockerActionManager;
+use JsonSchema\Validator;
 
 class ContainerDefinitionFetcher
 {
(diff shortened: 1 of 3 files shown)


php/src/ContainerDefinitionFetcher.php
```diff
@@ -40,12 +41,25 @@ public function GetContainerById(string $id): Container
...
+        }
+    }
+
     /**
      * @return array
      */
     private function GetDefinition(bool $latest): array
     {
         $data = json_decode(file_get_contents(__DIR__ . '/../containers.json'), true);
+        $this->validateJson($data);
```
szaimen, line 62: possible that this needs to be validated as object and not as array, so would simplay need to get the file and not json_decode it...

php/src/ContainerDefinitionFetcher.php
```diff
@@ -40,12 +41,25 @@ public function GetContainerById(string $id): Container
         throw new \Exception("The provided id " . $id . " was not found in the container definition.");
     }
 
+    private function validateJson(array $data): void {
+        // Validate against json schema
+        $validator = new Validator;
+        $validator->validate($data, [file_get_contents(__DIR__ . '/../containers.json')]);
```
szaimen, line 47:
```suggestion
        $validator->validate($data, (array)[file_get_contents(__DIR__ . '/../containers.json')]);
```

PR #2660, opened 2023-06-03
Make sure to disable apps that could not get enabled after an update

szaimen:


Containers/nextcloud/entrypoint.sh
```diff
@@ -362,6 +362,7 @@ DATADIR_PERMISSION_CONF
                         if [ "${APPSTORAGE[$app]}" != "no" ]; then
                             echo "Enabling $app..."
                             if ! php /var/www/html/occ app:enable "$app" >/dev/null; then
+                                php /var/www/html/occ app:disable "$app" >/dev/null
```
szaimen, line 365: If ! Rm app

PR #2665, opened 2023-06-03
add Docker Lint and update some workflows

szaimen:
Close https://github.com/nextcloud/all-in-one/issues/2662


.github/workflows/docker-lint.yml
```diff
@@ -0,0 +1,38 @@
...
+
+    name: docker-lint
+
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v3
+
+      - name: install npm
+        run: |
+          sudo apt-get install nodejs npm -y
```
Zoey2936, line 28: why not https://github.com/actions/setup-node?

szaimen: Good point, thanks!

szaimen: This broke the logic so reverted for now.

Issue #2450, opened 2023-05-01
docker desktop - connecting to docker socket inside container with different user than root is broken after updating to 4.19


szaimen:
This is a Docker Desktop 4.19 bug.
See https://github.com/docker/for-win/issues/13447
and https://github.com/docker/for-mac/issues/6823

Workaround for now: revert back to Docker Desktop 4.18. See https://github.com/docker/for-win/issues/13447#issuecomment-1579399130

bobloadmire:
I'm running docker for windows, the installer won't let me downgrade to 4.18, just says installation is up to date. How are you downgrading?

EwanBSmith:
This appears to be fixed with docker desktop 4.21 (at least on windows)

moracabanas:
Still having issues to access docker sock:

### Container logs
```error
If SELinux is enabled on your host, see https://github.com/nextcloud/all-in-one#are-there-known-problems-when-selinux-is-enabled
If you are on TrueNas SCALE, see https://github.com/nextcloud/all-in-one#can-i-run-aio-on-truenas-scale
If you are on Docker Desktop v4.19 or higher, see https://github.com/nextcloud/all-in-one/issues/2450
Cannot connect to the docker socket. Cannot proceed.
```


### Docker desktop version
```info
Server: Docker Desktop 4.21.1 (114176)
 Engine:
  Version:          24.0.2
  API version:      1.43 (minimum version 1.12)
  Go version:       go1.20.4
  Git commit:       659604f
  Built:            Thu May 25 21:52:17 2023
  OS/Arch:          linux/amd64
  Experimental:     false
 containerd:
  Version:          1.6.21
  GitCommit:        3dce8eb055cbb6872793272b4f20ed16117344f8
 runc:
  Version:          1.1.7
  GitCommit:        v1.1.7-0-g860f061
 docker-init:
  Version:          0.19.0
  GitCommit:        de40ad0
  ```

### Docker compose 
```
version: '3'
services:
  nextcloud-aio-mastercontainer:
    image: nextcloud/all-in-one:latest
    container_name: nextcloud-aio-mastercontainer
    privileged: true
    security_opt: ["label=disabled"]
    environment:
      - SKIP_DOMAIN_VALIDATION=true
    restart: always
    ports:
      - '80:80'
      - '8080'
      - '8443'
    volumes:
      - nextcloud_aio_mastercontainer:/mnt/docker-aio-config
[...]

PR #2670, opened 2023-06-05
add note about SELinux and Docker Desktop

szaimen:
For https://github.com/nextcloud/all-in-one/discussions/485#discussioncomment-6088261
and https://github.com/nextcloud/all-in-one/issues/2450

### DIFF
--- a/Containers/mastercontainer/start.sh
+++ b/Containers/mastercontainer/start.sh
@@ -62,7 +62,7 @@ fi
 if ! sudo -u www-data docker info &>/dev/null; then
     print_red "Cannot connect to the docker socket. Cannot proceed."
     echo "If you are on Docker Desktop v4.19 or higher, see https://github.com/nextcloud/all-in-one/issues/2450"
-    echo "If SELinux is enabled on your host, see https://github.com/nextcloud/all-in-one#how-to-change-the-default-location-of-nextclouds-datadir#are-there-known-problems-when-selinux-is-enabled"
+    echo "If SELinux is enabled on your host, see https://github.com/nextcloud/all-in-one#are-there-known-problems-when-selinux-is-enabled"
     exit 1
 fi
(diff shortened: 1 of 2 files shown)


Containers/mastercontainer/start.sh
```diff
@@ -61,6 +61,8 @@ fi
 # Check if api version is supported
 if ! sudo -u www-data docker info &>/dev/null; then
     print_red "Cannot connect to the docker socket. Cannot proceed."
+    echo "If you are on Docker Desktop v4.19 or higher, see https://github.com/nextcloud/all-in-one/issues/2450"
+    echo "If SELinux is enabled on your host, see https://github.com/nextcloud/all-in-one#how-to-change-the-default-location-of-nextclouds-datadir#are-there-known-problems-when-selinux-is-enabled"
```
Zoey2936, line 65:
```suggestion
    echo "If SELinux is enabled on your host, see https://github.com/nextcloud/all-in-one#are-there-known-problems-when-selinux-is-enabled"
```

Zoey2936:
maybe also add this option commented to the compose file

Zoey2936:
also not sure if privileged or security opt is better

szaimen:
> also not sure if privileged or security opt is better

I think privileged is the last restore so I would say security opt is more secure.

Containers/mastercontainer/start.sh
szaimen: thanks, fixed!

szaimen:
> maybe also add this option commented to the compose file

I would leave it out for now. If the question comes up more often after adding this doc, we can still adjust the compose.yaml later on.

szaimen:
> > also not sure if privileged or security opt is better
> 
> I think privileged is the last resort so I would say security opt is more secure.

image

From https://danwalsh.livejournal.com/74095.html

Zoey2936:
ahh I see

Zoey2936 (approved):

PR #2645, opened 2023-05-31
add talk-recording container

Zoey2936:
For https://github.com/nextcloud/all-in-one/discussions/2527

I've added recording.conf to the update script to see, when there are changes done to it, also backend.secret needs to be changed in the config and added to all other configs in aio
@szaimen If you want just add all other recording changes to this PR

- [ ] Needs https://github.com/nextcloud-releases/all-in-one/pull/40

PR #2674, opened 2023-06-05
allow to disable/enable talk-recording

szaimen:
For https://github.com/nextcloud/all-in-one/discussions/2527

- [x] needs https://github.com/nextcloud/all-in-one/pull/2645


Containers/nextcloud/entrypoint.sh
```diff
@@ -583,6 +583,17 @@ else
     fi
 fi
 
+# Talk recording
+if [ -d "/var/www/html/custom_apps/spreed" ]; then
```
szaimen, line 587: possibly need to wait for the recording server to become available...

PR #2679, opened 2023-06-06
Move vars to lets and fix some things with recording.conf

szaimen:
Follow-up for talk-recording

### DIFF
--- a/Containers/talk-recording/Dockerfile
+++ b/Containers/talk-recording/Dockerfile
@@ -17,6 +17,7 @@ RUN set -ex; \
         git \
         wget \
         shadow \
+        pulseaudio \
         openssl; \
     # chromium chromium-chromedriver?
     apk add --no-cache geckodriver --repository http://dl-cdn.alpinelinux.org/alpine/edge/testing; \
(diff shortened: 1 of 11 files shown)


Containers/talk-recording/Dockerfile
```diff
@@ -17,6 +17,7 @@ RUN set -ex; \
         git \
         wget \
         shadow \
+        pulseaudio \
```
Zoey2936, line 20: is libpulse not enough?

szaimen: Pulseaudio is recommended: https://github.com/nextcloud/spreed/pull/8788/files#diff-3e64353968025a91591a94eb471fd10febcb2b16fb933d84c460f2132029d9bbR39

Containers/talk-recording/Dockerfile
```diff
@@ -28,13 +29,14 @@ RUN set -ex; \
     python3 -m pip install /src/recording/src; \
     rm -rf /src; \
     chown recording:recording -R \
-        /tmp; \
+        /tmp /etc/recording.conf; \
```
Zoey2936, line 32: maybe instead of copying the file as step do a touch to create the file and them run chown?

szaimen: Sounds good!

Zoey2936: can you please do it?

Containers/talk-recording/Dockerfile
Zoey2936: so it requires the bin file? since for launchen the lib was enough

Containers/talk-recording/Dockerfile
szaimen: Done with https://github.com/nextcloud/all-in-one/commit/a797f624ec1d0456dd80f99bc3825aab2ffadd77

Containers/talk-recording/Dockerfile
Zoey2936: then libpulse can be removed

PR #2703, opened 2023-06-07
put notify-push into its own container

szaimen:
Close https://github.com/nextcloud/all-in-one/issues/2695


Containers/notify-push/Dockerfile
```diff
@@ -0,0 +1,22 @@
+FROM alpine:3.18.1
```
szaimen, line 1: not available yet...

PR #2710, opened 2023-06-07
rename some secrets

szaimen:
- [ ] TODO: find out how to keep the current secrets in the backend


szaimen:
I think there is no good way. Dropping this for now

PR #2725, opened 2023-06-10
add run_upgrade_if_needed_due_to_app_update

szaimen:
- [x] needs https://github.com/nextcloud/server/pull/38735
- [x] todo: https://github.com/nextcloud/all-in-one/discussions/2726
- [x] todo: disalke integrity check in the function before running occ upgrade. Also enable it before thr next regular update
- [x] close https://github.com/nextcloud/all-in-one/discussions/1787

### DIFF
--- a/Containers/nextcloud/entrypoint.sh
+++ b/Containers/nextcloud/entrypoint.sh
@@ -10,6 +10,13 @@ directory_empty() {
     [ -z "$(ls -A "$1/")" ]
 }
 
+run_upgrade_if_needed_due_to_app_update() {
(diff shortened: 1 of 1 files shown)


Containers/nextcloud/entrypoint.sh
```diff
@@ -404,6 +420,7 @@ DATADIR_PERMISSION_CONF
         UPDATED_APPS="$(php /var/www/html/occ app:update --all)"
         # Update all apps again and try to prevent something like https://github.com/nextcloud/polls/issues/2793 from happening
         php /var/www/html/occ app:update --all
```
szaimen, line 422: Can be removed now

Issue #1992, opened 2023-02-15
Finish Talk migration to alpine


Zoey2936:
Things that should be done after #1583 is finished

coturn:
- [x] https://github.com/coturn/coturn/pull/1156 to be merged
- [x] new coturn image release (https://github.com/coturn/coturn/releases)
- [x] change `FROM alpine:3.17.x` to latest `FROM coturn/cotrun:x.x.x-rx-alpine` (https://hub.docker.com/layers/coturn/coturn/4.6.2-alpine/images/sha256-52c5e7ec35da4f20b434883909fcafb30a4aa0c7de2f7112836e22eb33a04e58?context=explore)
- [x] remove coturn from normal `apk add` line

janus-gateway:
- [ ] alpine v3.18.1 release
- [ ] [coturn](https://github.com/coturn/coturn/blob/master/docker/coturn/alpine/Dockerfile) image updated to v3.18.1
- [ ] move janus gateway to normal `apk add` line and remove `apk add repository edge/community` line
- [ ] upgrade luarocks / lua-dev to 5.4
- [ ] wget needs to be tmp added and can be removed after luarocks installed everything

Zoey2936:
maybe we could ask here; https://gitlab.alpinelinux.org/alpine/aports/-/issues what is missing that janus-gateway becomes available outside testing?

szaimen:
Sounds good to me 👍

Zoey2936:
https://gitlab.alpinelinux.org/alpine/aports/-/issues/14661

Zoey2936:
that was a fast awnser (https://gitlab.alpinelinux.org/alpine/aports/-/issues/14661#note_293079)

szaimen:
Oh wow!

Zoey2936:
There is already a PR: https://gitlab.alpinelinux.org/alpine/aports/-/merge_requests/44537

szaimen:
Nice 🚀

Zoey2936:
If they continue with this speed, janus-gateway is faster inside v3.17 then signaling is released and coturn is based on alpine 3.17

Zoey2936:
https://gitlab.alpinelinux.org/alpine/aports/-/issues/14661#note_293460

szaimen:
Wow, that was really fast! Okay, so it will be included in alpine 3.18, nice🎉

Zoey2936:
[...]

PR #2768, opened 2023-06-15
update all container to alpine v3.18.3

Zoey2936:
Close https://github.com/nextcloud/all-in-one/discussions/2533

Close https://github.com/nextcloud/all-in-one/issues/1992


Containers/imaginary/Dockerfile
```diff
@@ -1,6 +1,6 @@
-FROM golang:1.20.5-alpine3.17 as go
+FROM golang:1.20.5-alpine3.18 as go
 
-ENV IMAGINARY_HASH b632dae8cc321452c3f85bcae79c580b1ae1ed84
+ENV IMAGINARY_HASH b632dae8cc321452c3f85bcae79c580b1ae1ed84
```
Zoey2936, line 3: noidea what it did here...

Containers/talk/Dockerfile
```diff
@@ -1,6 +1,6 @@
 FROM nats:2.9.17-scratch as nats
 FROM strukturag/nextcloud-spreed-signaling:1.1.2 as signaling
-FROM coturn/coturn:4.6.2-r0-alpine
+FROM coturn/coturn:4.6.2-r3-alpine
```
Zoey2936, line 3:
```suggestion
FROM coturn/coturn:4.6.2-alpine
```

Containers/imaginary/Dockerfile
szaimen: probably again line break added back. It is fine.

Containers/talk/Dockerfile
szaimen: Maybe we could ask them if they can add the alpine version into the tag so that we can pin the alpine version?

szaimen (approved):
LGTM, thanks! :)

Containers/talk/Dockerfile
Zoey2936: > I would keep that

not sure if dependabot understands that

Zoey2936: > Maybe we could ask them if they can add the alpine version into the tag so that we can pin the alpine version?

can you please do this?

szaimen: https://github.com/coturn/coturn/issues/1226

PR #2787, opened 2023-06-17
show AIO password

Zoey2936:
Would look like this:
grafik

### DIFF
--- a/php/public/forms.js
+++ b/php/public/forms.js
@@ -39,6 +39,12 @@
     document.getElementById('overlay').classList.remove('loading');
   }
 
+  function mirrorPassword() {
+    var Password = document.getElementById("Password");
+    var mirroredPassword = document.getElementById("mirroredPassword");
+    mirroredPassword.innerText = Password.value;
+  }
+
   function initForm(form) {
     function submit(event)
     {
--- a/php/public/forms.js
+++ b/php/public/forms.js
@@ -43,7 +43,7 @@
     let passwordField = document.getElementById(id);
     if (passwordField.type === "password" && passwordField.value !== "") {
       passwordField.type = "text";
-    } else if (passwordField.value === "" && passwordField.type !== "password") {
+    } else if (passwordField.type === "text" && passwordField.value === "") {
       passwordField.type = "password";
     }
   }
--- a/php/public/forms.js
+++ b/php/public/forms.js
@@ -41,7 +41,7 @@
 
   function showPassword(id) {
     let passwordField = document.getElementById(id);
-    if (passwordField.value !== "" && passwordField.type !== "text") {
+    if (passwordField.type === "password" && passwordField.value !== "") {
       passwordField.type = "text";
     } else if (passwordField.value === "" && passwordField.type !== "password") {
       passwordField.type = "password";
--- a/php/public/forms.js
+++ b/php/public/forms.js
@@ -41,7 +41,7 @@
 
   function showPassword() {
     var Password = document.getElementById("Password");
-    if (Password.value !== "" && x.type === "password") {
+    if (Password.value !== "" && Password.type === "password") {
       Password.type = "text";
     } else {
       Password.type = "password";
--- a/php/public/forms.js
+++ b/php/public/forms.js
@@ -39,12 +39,12 @@
     document.getElementById('overlay').classList.remove('loading');
   }
 
-  function showPassword() {
-    var Password = document.getElementById("Password");
-    if (Password.value !== "" && Password.type === "password") {
-      Password.type = "text";
-    } else {
-      Password.type = "password";
+  function showPassword(id) {
+    let passwordField = document.getElementById(id);
+    if (passwordField.value !== "" && passwordField.type !== "text") {
+      passwordField.type = "text";
+    } else if (passwordField.value === "" && passwordField.type !== "password") {
+      passwordField.type = "password";
     }
   }
 
--- a/php/public/forms.js
+++ b/php/public/forms.js
@@ -39,10 +39,13 @@
     document.getElementById('overlay').classList.remove('loading');
   }
 
-  function mirrorPassword() {
+  function showPassword() {
     var Password = document.getElementById("Password");
-    var mirroredPassword = document.getElementById("mirroredPassword");
-    mirroredPassword.innerText = Password.value;
+    if (Password.value !== "" && x.type === "password") {
+      Password.type = "text";
+    } else {
+      Password.type = "password";
+    }
   }
 
   function initForm(form) {
--- a/php/templates/login.twig
+++ b/php/templates/login.twig
@@ -8,9 +8,10 @@
             {% if is_login_allowed == true %}
                 <p>Log in using your Nextcloud AIO password:</p>
                 <form method="POST" action="/api/auth/login" class="xhr">
-                    <input type="text" autocomplete="off" name="password" placeholder="Password" />
+                    <input name="password" placeholder="Password" autocomplete="on" type="password" wfd-id="id0" id="Password" oninput="mirrorPassword()">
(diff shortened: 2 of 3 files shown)


szaimen:
Actually I think this is probably more confusion than simply going with the show password button. So lets indeed use implementation from https://www.w3schools.com/howto/howto_js_toggle_password.asp

Zoey2936:
> Actually I think this is probably more confusing than simply going with the show password checkbox. So lets just use implementation from https://www.w3schools.com/howto/howto_js_toggle_password.asp

this was just an alternative Idea, do it the way you want

szaimen:
Would you mind adjusting the PR?

Zoey2936:
in which way?

szaimen:
To use the solution mentioned in https://www.w3schools.com/howto/howto_js_toggle_password.asp

szaimen:
> or another idea: what about setting it to a password field and when the first letter/number/symbol ist filled in change it to a text field?

If this works, probably it would be the best solution. Thanks for the idea! :)

Zoey2936:
I will try it later

Zoey2936:
works now

php/public/forms.js
```diff
@@ -39,6 +39,15 @@
     document.getElementById('overlay').classList.remove('loading');
   }
 
+  function showPassword() {
+    var Password = document.getElementById("Password");
+    if (Password.value !== "" && Password.type === "password") {
+      Password.type = "text";
+    } else {
```
Zoey2936, line 46: maybe and else if would be better

php/public/forms.js
```diff
@@ -39,6 +39,15 @@
     document.getElementById('overlay').classList.remove('loading');
   }
 
+  function showPassword() {
+    var Password = document.getElementById("Password");
+    if (Password.value !== "" && Password.type === "password") {
```
Zoey2936, line 44:
```suggestion
    if (Password.value !== "" && Password.type !== "text") {
```

php/public/forms.js
Zoey2936:
```suggestion
    } else if (Password.value === "" && Password.type !== "password") {
```

Zoey2936:
maybe do the same here:
https://github.com/nextcloud/all-in-one/blob/d0cae686365409daab32c71b962a7debd8e5262c/php/templates/containers.twig#L512
https://github.com/nextcloud/all-in-one/blob/d0cae686365409daab32c71b962a7debd8e5262c/php/templates/containers.twig#L5123

Zoey2936:
@szaimen what do you think, should I also add this to the password change field?

php/public/forms.js
```diff
@@ -39,6 +39,15 @@
...
+      Password.type = "text";
+    } else {
+      Password.type = "password";
+    }
+  }
```
szaimen, line 49:
```suggestion
function showPassword(id) {
    let passwordField = document.getElementById(id);
    if (passwordField) {
        if (passwordField .value === "") {
          passwordField .type = "password";
        } else {
            if (passwordField .type === "password") {
                passwordField .type = "text";
            }
        }
    }
}
```

php/templates/login.twig
```diff
@@ -8,7 +8,7 @@
             {% if is_login_allowed == true %}
                 <p>Log in using your Nextcloud AIO password:</p>
                 <form method="POST" action="/api/auth/login" class="xhr">
-                    <input type="text" autocomplete="off" name="password" placeholder="Password" />
+                    <input type="password" autocomplete="current-password" name="password" placeholder="Password" wfd-id="id0" id="Password" oninput="showPassword()">
```
szaimen, line 11:
```suggestion
                    <input type="passwordInput" autocomplete="current-password" name="password" placeholder="Password" id="Password" oninput="showPassword('passwordInput')">

```

szaimen:
> @szaimen what do you think, should I also add this to the password change field?

Yes, can you also use the same function in all of the fields 

> @szaimen what do you think, should I also add this to the password change field?

yes, probably makes sense 👍

php/public/forms.js
Zoey2936, line 49:
```suggestion
  function showPassword(id) {
    let passwordField = document.getElementById(id);
    if (passwordField.value !== "" && passwordField.type !== "text") {
      passwordField.type = "text";
    } else if (passwordField.value === "" && passwordField.type !== "password") {
      passwordField.type = "password";
    }
  }
```

php/templates/login.twig
Zoey2936: why do you change the type? you need to change the id

Zoey2936:
```suggestion
                    <input type="password" autocomplete="current-password" name="password" placeholder="Password" id="Password" oninput="showPassword('Password')">

```

szaimen:
```suggestion
                    <input type="password" autocomplete="current-password" name="password" placeholder="Password" 
                    id="passwordInput" oninput="showPassword('passwordInput')">

```

Zoey2936:
```suggestion
                    <input type="password" autocomplete="current-password" name="password" placeholder="Password" id="passwordInput" oninput="showPassword('passwordInput')">

```

php/public/forms.js
```diff
@@ -39,6 +39,15 @@
     document.getElementById('overlay').classList.remove('loading');
   }
 
+  function showPassword(id) {
+    let passwordField = document.getElementById(id);
+    if (passwordField.value !== "" && passwordField.type !== "text") {
```
szaimen, line 44:
```suggestion
    if (passwordField.type === "password" && passwordField.value !== "") {
```

php/public/forms.js
```diff
@@ -39,6 +39,15 @@
...
+      passwordField.type = "text";
+    } else if (passwordField.value === "" && passwordField.type !== "password") {
```
szaimen, line 46:
```suggestion
    } else if (passwordField.type === "text" && passwordField.value === "") {
```

php/templates/login.twig
```diff
@@ -8,7 +8,8 @@
             {% if is_login_allowed == true %}
                 <p>Log in using your Nextcloud AIO password:</p>
                 <form method="POST" action="/api/auth/login" class="xhr">
-                    <input type="text" autocomplete="off" name="password" placeholder="Password" />
+                    <input type="password" autocomplete="current-password" name="password" placeholder="Password" id="passwordInput" oninput="showPassword('passwordInput')">
```
szaimen, line 11:
```suggestion
                    <input type="password" autocomplete="current-password" name="password" placeholder="Password" id="master-password" oninput="showPassword('master-password')">

```

php/templates/login.twig
```diff
@@ -8,7 +8,8 @@
             {% if is_login_allowed == true %}
                 <p>Log in using your Nextcloud AIO password:</p>
                 <form method="POST" action="/api/auth/login" class="xhr">
-                    <input type="text" autocomplete="off" name="password" placeholder="Password" />
+                    <input type="password" autocomplete="current-password" name="password" placeholder="Password" id="passwordInput" oninput="showPassword('passwordInput')">
+
```
szaimen, line 12:
```suggestion
```

php/public/forms.js
Zoey2936: and why?

szaimen: because it is more performant and I think the logic is more understandable by so

php/public/forms.js
szaimen: why?

szaimen (approved):
LGTM, thank you! :)

szaimen:
@Zoey2936 I fear I get this issue: 
image

Zoey2936:
sorry, can't check this at the moment

PR #2788, opened 2023-06-17
Update spellcheck.yml

szaimen:


szaimen:
It finds more typos than the other 🎉

PR #2807, opened 2023-06-18
coturn container - pin alpine version manually

szaimen:


Zoey2936:
so you want the build to fail if the version is different, right?

szaimen:
> so you want the build to fail if the version is different, right?

Yes

Zoey2936:
maybe then remove the -r3 since I think dependabot doesn't understand that

szaimen:
> maybe then remove the -r3 since I think dependabot doesn't understand that

good idea

szaimen:
https://github.com/nextcloud/all-in-one/pull/2817

Issue #2506, opened 2023-05-04
Think about switching some containers to readonlyrootfs


szaimen:
in order to improve the security

szaimen:
Possible:
- [ ] Apache (in theory but Caddyfile would need to not get changed then by sed)
- [ ] borgbackup (yes, needs testing)
- [ ] clamav (yes, needs testing)
- [x] ~~collabora (probably no as they copy some stuff internally)~~
- [ ] domaincheck (yes? /var/www/domaincheck/index.html would need to be static then)
- [ ] fulltextsearch (yes? needs testing)
- [ ] imaginary (yes? needs testing)
- [x] ~~mastercontainer (no)~~
- [ ] redis 
- [ ] talk: need

szaimen:
Redis is now read-only which is released with v6.1.0 Beta. Testing and feedback is welcome! See https://github.com/nextcloud/all-in-one#how-to-switch-the-channel

szaimen:
Apache, domaincheck, postgresql, talk and talk-recording are now also read-only which is released with v6.3.0 Beta. Testing and feedback is welcome! See https://github.com/nextcloud/all-in-one#how-to-switch-the-channel

sriccio:
Hmm, by curiositry, what's the goal for the containers to be read only ?

szaimen:
See e.g. https://www.techtarget.com/searchitoperations/answer/Is-read-only-mode-a-viable-approach-to-container-hardening#:~:text=Setting%20containers%20as%20read%2Donly,anyone%20else%20from%20doing%20so.

sriccio:
Hello @szaimen,

Oh, thanks for the link. Sounds great indeed.

Kind regards

PR #2812, opened 2023-06-19
Make borgbackup read-only

szaimen:
Address part of https://github.com/nextcloud/all-in-one/issues/2506

- [x] needs https://github.com/nextcloud/all-in-one/pull/2811

### DIFF
--- a/php/containers-schema.json
+++ b/php/containers-schema.json
@@ -141,7 +141,7 @@
 						"type": "array",
 						"items": {
 							"type": "string",
-							"pattern": "^/[a-z/]$"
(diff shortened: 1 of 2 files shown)


Zoey2936:
you could also add `VOLUME /tmp` to the dockerfile

szaimen:
> you could also add `VOLUME /tmp` to the dockerfile

I tried this. Unfortunately it does not play well when converting to kubernetes. So better to use tmpfs for that because it works out-of-the-box and is intended for that purpose IIRC :)

PR #2813, opened 2023-06-19
make clamav read-only

szaimen:
Address part of https://github.com/nextcloud/all-in-one/issues/2506


szaimen:
This does not work currently as tmpfs is set to 755 by default inside the container IIRC. See https://github.com/docker/compose/issues/3425#issuecomment-423091370 for a workaround. However we might need to switch to mount syntax internally in order to specify the permission mode. See https://docs.docker.com/storage/tmpfs/

However this does not work with docker-compose. So not sure how to proceed here.

Containers/clamav/Dockerfile
```diff
@@ -12,6 +12,7 @@ RUN set -ex; \
     chmod 770 -R /var/run/clamav /run/clamav /var/log/clamav /var/lock /run/lock
 
 VOLUME /var/lib/clamav
+VOLUME /var/lock
```
szaimen, line 15:
```suggestion
```

PR #2826, opened 2023-06-20
mastercontainer - disable http3

szaimen:
Reason: http3 seems overkill for the AIO interface since it is server-side-rendered and should transfer only very few bytes.

### DIFF
--- a/Containers/mastercontainer/Caddyfile
+++ b/Containers/mastercontainer/Caddyfile
@@ -12,7 +12,7 @@
     }
 
     servers {
-        protocols h1 h2
+        protocols h1 h2 h2c
     }
 }
 
--- a/Containers/mastercontainer/Caddyfile
+++ b/Containers/mastercontainer/Caddyfile
@@ -10,6 +10,10 @@
     log {
         level ERROR
     }
+
+    servers {
+        protocols h1 h2
+    }
 }
 
 http://:80 {


Containers/mastercontainer/Caddyfile
```diff
@@ -10,6 +10,10 @@
     log {
         level ERROR
     }
+
+    servers {
+        protocols h1 h2
+    }
```
Zoey2936, line 16:
```suggestion
    servers :80 {
        protocols h1 h2c
    }
```

szaimen: so that would only be disabled then for port 80? I want to disable it for all connections, see the reasoning above.

Zoey2936: but the caddy file only uses port 80?

Zoey2936: then at least also add h2c to this line

szaimen: and 8443, no?

Zoey2936: ohh yes

Zoey2936: I adjusted the suggestion

Zoey2936 (approved):

Croydon:
Was there a disadvantage of having it enabled?

Zoey2936:
> Was there a disadvantage of having it enabled?

it would require binding an additional port on udp

PR #2828, opened 2023-06-20
apache - expose udp for http3 to work

szaimen:
Close https://github.com/nextcloud/all-in-one/discussions/2825

### DIFF
--- a/php/containers.json
+++ b/php/containers.json
@@ -16,6 +16,11 @@
           "ip_binding": "%APACHE_IP_BINDING%",
           "port_number": "%APACHE_PORT%",
           "protocol": "tcp"
+        },
+        {
+          "ip_binding": "%APACHE_IP_BINDING%",
+          "port_number": "%APACHE_PORT%",
+          "protocol": "udp"
         }
       ],
       "internal_port": "%APACHE_PORT%",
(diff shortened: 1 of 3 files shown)


Zoey2936:
I just want to say that WebKit sometimes throws an error when the alt-srv header is sent, but http3 is not reachable

szaimen:
> I just want to say that WebKit sometimes throws an error when the alt-srv header is sent, but http3 is not reachable

so what to do then?

Zoey2936:
> > I just want to say that WebKit sometimes throws an error when the alt-srv header is sent, but http3 is not reachable
> 
> so what to do then?

it is a very rare issue, the only thing would be to disable http3 or opening the ports

szaimen:
Will Webkit bug completely out if the port 443/udp is not available and http3 enabled or is it a minor issue like a cosmetic one?

Zoey2936:
> Will Webkit bug completely out if the udp is not available and http3 enabled or is it a minor issue like a cosmetic one?

sometimes it will show that the webserver is unavailable, but as I said this is very rare

szaimen:
> sometimes it will show that the webserver is unavailable, but as I said this is very rare

I see. then lets hope that they fix it in a future update :)

Zoey2936 (approved):

PR #2854, opened 2023-06-24
switch to cronie

Zoey2936:
Close https://github.com/nextcloud/all-in-one/discussions/2829

### DIFF
--- a/Containers/mastercontainer/Dockerfile
+++ b/Containers/mastercontainer/Dockerfile
@@ -36,7 +36,8 @@ RUN set -ex; \
         sudo \
         netcat-openbsd \
         curl \
-        grep; \
+        grep \
+        cronie; \
     \
     apk add --no-cache --virtual .build-deps \
         autoconf \
@@ -109,7 +110,8 @@ RUN set -ex; \
           /etc/apache2/conf.d/info.conf; \
     \
     mkdir /var/log/supervisord; \
-    mkdir /var/run/supervisord;
+    mkdir /var/run/supervisord; \
+    echo '* * * * * /cron.sh' | tee /etc/crontabs/root
 
 COPY --chmod=775 *.sh /
 COPY --chmod=664 Caddyfile /Caddyfile
--- a/Containers/nextcloud/Dockerfile
+++ b/Containers/nextcloud/Dockerfile
@@ -24,10 +24,12 @@ RUN set -ex; \
     addgroup -g 33 -S www-data; \
     adduser -u 33 -D -S -G www-data www-data; \
     \
-# entrypoint.sh and cron.sh dependencies
+# entrypoint.sh and cron dependencies
     apk add --no-cache \
         rsync \
+        cronie \
     ; \
+    echo '*/5 * * * * php -f /var/www/html/cron.php' | tee /etc/crontabs/www-data; \
 # install the PHP extensions we need
 # see https://docs.nextcloud.com/server/stable/admin_manual/installation/source_installation.html
     apk add --no-cache --virtual .build-deps \
--- a/Containers/nextcloud/cron.sh
+++ b/Containers/nextcloud/cron.sh
@@ -1,7 +1,0 @@
-#!/bin/bash
-set -eu
-
-while true; do
-    php -f /var/www/html/cron.php &
-    sleep 5m
-done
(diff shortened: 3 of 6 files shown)


Containers/mastercontainer/Dockerfile
```diff
@@ -109,7 +110,8 @@ RUN set -ex; \
           /etc/apache2/conf.d/info.conf; \
     \
     mkdir /var/log/supervisord; \
-    mkdir /var/run/supervisord;
+    mkdir /var/run/supervisord; \
+    echo '* * * * * /cron.sh' | tee /etc/crontabs/root
```
Zoey2936, line 114: maybe each line of the cron.sh should be translated into cron syntax

szaimen:
Thanks a lot for the PR @Zoey2936! However since the mastercontainer relies on only spawning one cron.sh at the time and other reasons I would recommend to only migrate the Nextcloud cronjob to use cronie. Would that be fine for you?

Containers/nextcloud/Dockerfile
```diff
@@ -24,10 +24,12 @@ RUN set -ex; \
     addgroup -g 33 -S www-data; \
     adduser -u 33 -D -S -G www-data www-data; \
     \
-# entrypoint.sh and cron.sh dependencies
+# entrypoint.sh and cron dependencies
     apk add --no-cache \
         rsync \
+        cronie \
     ; \
+    echo '*/5 * * * * php -f /var/www/html/cron.php' | tee /etc/crontabs/www-data; \
```
szaimen, line 32: I wonder will cronie keep environmental varialbes? Otherwise php will probably run into an issue?

Zoey2936: the cron.php requires env?

szaimen: Yes due to https://github.com/nextcloud/all-in-one/blob/27dbeec29d720eb19b99c2e47e251e3cb92c2882/Containers/nextcloud/Dockerfile#L105

Zoey2936: but thats not an env?

szaimen: It is -> `'` is used and not `"`

szaimen: image

Zoey2936: ahhh now I understand, then lets close this PR, even if I think that cronie would be a better implementation then looping shell script in a while true loop in combination with sleep, if you understand what I mean

szaimen: > even if I think that cronie would be a better implementation

Yes I agree

szaimen:
> then lets close this PR,

okay

Containers/nextcloud/cron.sh
```diff
@@ -1,7 +0,0 @@
-#!/bin/bash
-set -eu
-
-while true; do
-    php -f /var/www/html/cron.php &
-    sleep 5m
-done
```
Zoey2936, line 7: but one thing, why is there a "&"?

szaimen: Because it shall spawn a new process every 5min (even if they run in parallel then)

Zoey2936: I don't understand why the & changes the logic, beside running both commands at the same time

szaimen: Because the command would be blocking otherwise so onky after it being finished counting the 5min and then starting the next process iirc

Zoey2936: so you mean that in the time of the five minutes the time cron.php needs to run is ignored?

szaimen: Yes

Zoey2936: but is there not a `& wait` missing? since if there is none them php command would be stopped?

szaimen: What would be the advantage? Can you propose a PR?

szaimen: I see. Can you propose a PR?

Zoey2936: can you please do it? i don't like to create prs with my mobile phone

szaimen: https://github.com/nextcloud/all-in-one/pull/2974

PR #2856, opened 2023-06-24
Add Temp Dir Variable 

ralphte:
This pull request would allow the option to change the Nextcloud temp directory outside the data folder. This can improve performance in certain situations.


szaimen:
> This can improve performance in certain situations.

In which situation?

ralphte:
> > This can improve performance in certain situations.
> 
> In which situations?

Well, primarily during upload when the temp folder is on a slower drive like a spinning HD. The upload function will usually segment large files into much smaller files. The segments could be in the 10's of thousands of small files during a few GB upload. While these files are quickly deleted after reassembly, this IOPS-intensive task is better suited for a SSD than a spinning HD. Moving the temp folder to an SSD helped sync speed during testing. Additionally, this pull request is strictly opt-in, and anyone not wanting this change would not be affected in any way.

szaimen:
Hi, I would rather not make this adjustable if we can find a good default for this. What do you think about https://github.com/nextcloud/all-in-one/pull/2868?

ralphte:
> Hi, I would rather not make this adjustable if we can find a good default for this. What do you think about #2868?

That sounds good. While it's nice to have options, the temp folder should be on the fastest storage possible by default. and anyone uploading very large files should plan their OS/Nextcloud hardrive location according.

szaimen:
Superceded by https://github.com/nextcloud/all-in-one/pull/2868

PR #2868, opened 2023-06-26
revert adjustment of tempdirectory

szaimen:
Supercedes https://github.com/nextcloud/all-in-one/pull/2856

Motivation: we introduced this tempdir setting initially in order to make big file uploads more reliable. However later on we found out that inside containers usually no tmpfs is used by default which would come with limitations in regard to the max size. Additionally, setting this to the datadir comes with drawbacks like performance problems and symlink problems. So better to just revert this setting and use the default.

### DIFF
--- a/Containers/nextcloud/entrypoint.sh
+++ b/Containers/nextcloud/entrypoint.sh
@@ -444,12 +444,14 @@ if [ -z "$OBJECTSTORE_S3_BUCKET" ] && [ -z "$OBJECTSTORE_SWIFT_URL" ]; then
         exit 1
     fi
 
-    # Configure tempdirectory
-    mkdir -p "$NEXTCLOUD_DATA_DIR/tmp/"
-    if ! grep -q upload_tmp_dir /usr/local/etc/php/conf.d/nextcloud.ini; then
-        echo "upload_tmp_dir = $NEXTCLOUD_DATA_DIR/tmp/" >> /usr/local/etc/php/conf.d/nextcloud.ini
+    # Delete formerly configured tempdirectory as the default is usually faster (if the datadir is on a HDD or network FS)
+    if [ "$(php /var/www/html/occ config:system:get tempdirectory)" = "$NEXTCLOUD_DATA_DIR/tmp/" ]; then
+        php /var/www/html/occ config:system:delete tempdirectory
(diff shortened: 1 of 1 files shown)


ralphte:
> @ralphte does this look good to you? :)

Looks good here

szaimen:
Thanks for the review! :)

CoderTobi:
Hi, so if I understand this right, the tmpdir will be inside the container again in future versions, right? But I think that might cause some problems again during big uploads over the windows client. From what I have read this was also the reason why it was placed in the datadir in the first place.

ralphte:
> Hi, so if I understand this right, the tmpdir will be inside the container again in future versions, right? But I think that might cause some problems again during big uploads over the windows client. From what I have read this was also the reason why it was placed in the datadir in the first place.

This all depends on how big of a file you upload. The problem is that the temp dir is used for more than just uploads; more importantly, it creates and deletes many small files. The temp folder should be on an SSD and not a platter disk. This is only an issue if you have two drives, one for your OS and one for your files. In any case, you should have enough available storage on the OS drive with docker to support your largest single file upload. The Nextcloud will look to move that file to the primary storage as soon as it can. If your biggest upload file is 10GB then 15GB of extra space should easily handle the task. This amount of extra space should not be a big deal.

JMarcosHP:
I would recommend to maintain the temp directory inside the container as default, but write a doc file explaining how to manually change this because many users have different use cases and only suggest this in case the users need to deploy the AIO for production or a dedicated server with hundred connections/uploads per day.

Setting the temp directory inside the container makes sence for a personal use case or deployment for like 5 or 10 users.

But forcing this configuration to all kind of situations is not a good idea.

Dedicated servers always have separate drives for user data and fast drives for cache and temp files. In this case, having this extra documentation makes more sence.

To make this possible, currently the Mastercontainer has an environment variable to mount extra drives `NEXTCLOUD_MOUNT` the user just need to specify the directory where all his mounted drives reside including the dedicated drive for cache/temp files. Then make a directory inside that drive and set the respective configuration inside the nextcloud container on `config/config.php`, `/usr/local/etc/php/conf.d/nextcloud.ini` and `/usr/local/etc/php-fpm.d/www.conf`. Finally add the respective ownership to www-data in the mounted folder and restart the container.

PR #2873, opened 2023-06-26
adjust wording for backup time

szaimen:
Address https://github.com/nextcloud/all-in-one/discussions/2860


szaimen:
@emelbardis does this look good to you? :)

emelbardis (approved):
Looks great. I think I was afraid to disable backups to test thinking that when I re-enabled, I'd lose my previous backups

szaimen:
All right! Thanks for the review and the feedback! :)

PR #2880, opened 2023-06-27
talk-recording - set allow_all and skip_verify via env

szaimen:


Zoey2936:
can I ask for a reason?

szaimen:
> can I ask for a reason?

@SystemKeeper asked me if he could reuse the container in @juliushaertl docker-dev. For that he needs to adjust these values in order to run it locally.

juliushaertl (approved):

SystemKeeper (changes requested):
No sorry this is not working like this, I’ll add some more details later

SystemKeeper:
So, back again. Again, very sorry for not finding the time earlier and now rushing into this :-(.

I noticed 2 things here I wanted to discuss:
1. Currently we have the secret set under `backend` with a TODO to remove it, after a issue in the recording server is fixed
This is now a problem, because when we use `ALLOW_ALL` we need to have this secret.

2. Right now we expose the signaling server directly, that is without a directory. We introduced the protocol and the domain separately, but that still wouldn't work, because for the dev environment we don't have the `/standalone-signaling/` path.

So what I thought was to do something like this:

```diff
diff --git a/Containers/talk-recording/start.sh b/Containers/talk-recording/start.sh
index 991ea69..f38a7a9 100644
--- a/Containers/talk-recording/start.sh
+++ b/Containers/talk-recording/start.sh
@@ -12,6 +12,15 @@ elif [ -z "$INTERNAL_SECRET" ]; then
     exit 1
 fi
 
+if [ -z "$SIGNALING_URL" ]; then
+    SIGNALING_URL="https://${NC_DOMAIN}/standalone-signaling/"
+fi
+
+# TODO: Enable if-clause below when https://github.com/nextcloud/spreed/issues/9580 is fixed
+#if [ "$ALLOW_ALL" = true ]; then
+    ALLOW_ALL_SECRET="secret = ${RECORDING_SECRET}"
+#fi
+
 cat << RECORDING_CONF > "/etc/recording.conf"
 [logs]
 # 30 means Warning
@@ -22,8 +31,7 @@ listen = 0.0.0.0:1234
 
 [backend]
 allowall = ${ALLOW_ALL}
-# TODO: remove secret below when https://github.com/nextcloud/spreed/issues/9580 is fixed
-secret = ${RECORDING_SECRET}
+${ALLOW_ALL_SECRET}
 backends = backend-1
 skipverify = ${SKIP_VERIFY}
 maxmessagesize = 1024
@@ -32,7 +40,7 @@ videoheight = 1080
 directory = /tmp
 
 [backend-1]
-url = ${HPB_PROTOCOL}://${NC_DOMAIN}
+url = https://${NC_DOMAIN}
 secret = ${RECORDING_SECRET}
 skipverify = ${SKIP_VERIFY}
 
@@ -40,7 +48,7 @@ skipverify = ${SKIP_VERIFY}
 signalings = signaling-1
 
 [signaling-1]
-url = ${HPB_PROTOCOL}://${NC_DOMAIN}/standalone-signaling/
+url = ${SIGNALING_URL}
 internalsecret = ${INTERNAL_SECRET}
 
 [ffmpeg]
```

For the first issue we just check if `ALLOW_ALL` is set and then additionally add the secret under `backend`. Right now the if-clause is commented, because of the issue mentioned above.

For the second issue we check if a `SIGNALING_URL` is provided and if not, we construct it as it was before. This way we could remove the `HPB_PROTOCOL` again. Alternatively we could of course just add a `HPB_PATH` as a variable.

szaimen:
> Alternatively we could of course just add a `HPB_PATH` as a variable.

I guess this would be the simplest approach then.

SystemKeeper (approved):
Tested and works! Thank you very much!

We need to keep in mind that the TODO about the global secret is not true anymore when we set `ALLOW_ALL` to `true`.

Also, it does not work with latest master of talk, because of https://github.com/nextcloud/spreed/pull/9940 (but no backport yet).

szaimen:
This is now released with v6.4.0 Beta. Testing and feedback is welcome! See https://github.com/nextcloud/all-in-one#how-to-switch-the-channel

PR #2888, opened 2023-06-29
documentation: misspelling

t-h2o:
# English error

## diff

```diff
- envorinmental
+ environment
```

## Command used to fix

```bash
sed -i \
's/envorinmental/environment/g' \
$(find . -type f)
```


szaimen (approved):
Thanks!

PR #1313, opened 2022-10-26
rework NGINX reverse proxy docs to support CODE

TheManchineel:
As per PR #1236, applied the same documentation change to the NGINX configuration, so as to pass the full URI of each HTTP request to the Collabora Online backend. This fixes #834 but for NGINX rather than Apache. It might be necessary to make similar changes to other proxy software, but NGINX is what I am familiar with.

Signed-off-by: TheManchineel <37479927+TheManchineel@users.noreply.github.com>

PR #2915, opened 2023-07-09
Update reverse-proxy.md

zuman:
$request_uri is not needed. Tested on my private cloud -> cloud.zuman.one


Zoey2936 (changes requested):
it is required

Zoey2936:
collabora can nake problems without it

Zoey2936:
see #1313

PR #2974, opened 2023-07-18
wait for cronjob

szaimen:
address https://github.com/nextcloud/all-in-one/pull/2854#discussion_r1265616468

### DIFF
--- a/Containers/nextcloud/cron.sh
+++ b/Containers/nextcloud/cron.sh
@@ -4,4 +4,5 @@ set -eu
 while true; do
     php -f /var/www/html/cron.php &
     sleep 5m
+    wait
 done
--- a/php/containers.json
+++ b/php/containers.json
@@ -201,6 +201,7 @@
         "TALK_RECORDING_HOST=nextcloud-aio-talk-recording"
       ],
       "restart": "unless-stopped",
+      "stop_grace_period": 600,
       "devices": [
         "/dev/dri"
       ],


Containers/nextcloud/cron.sh
```diff
@@ -5,3 +5,6 @@ while true; do
     php -f /var/www/html/cron.php &
     sleep 5m
 done
+
+# Wait for last command to end
+wait $!
```
szaimen, line 10: @Zoey2936 I have no clue if this actually works...

php/containers.json
```diff
@@ -201,6 +201,7 @@
         "TALK_RECORDING_HOST=nextcloud-aio-talk-recording"
       ],
       "restart": "unless-stopped",
+      "stop_grace_period": 600,
```
szaimen, line 204: @Zoey2936 do you think 10min are enought to wait for cronjobs to stop?

Containers/nextcloud/cron.sh
Zoey2936: https://stackoverflow.com/a/3004814 I meant that

szaimen: I see and where would I add this inside this script?

Zoey2936: see like I did it here in the last 2 lines https://github.com/ZoeyVid/nginx-proxy-manager/blob/develop/rootfs/bin/launch.sh and this: https://stackoverflow.com/a/68931259

Zoey2936: not sure if it makes a difference to use 
command1 &
command2 &
wait

or

command1 & 
command2
wait

php/containers.json
szaimen: @Zoey2936 if wait only waits for the sleep command, 5min should be enough?

Containers/nextcloud/cron.sh
```diff
@@ -4,4 +4,5 @@ set -eu
 while true; do
     php -f /var/www/html/cron.php &
     sleep 5m
+    wait
```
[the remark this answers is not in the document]
szaimen: I see but we want the next php command to start already even if the old one is still running. Is that possible?

php/containers.json
Zoey2936: no

Containers/nextcloud/cron.sh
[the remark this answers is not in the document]
Zoey2936: does the cron php support thst?

php/containers.json
szaimen: https://docs.docker.com/compose/compose-file/compose-file-v3/#stop_grace_period

Containers/nextcloud/cron.sh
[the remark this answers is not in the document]
Zoey2936: i will check the behaivior later again

php/containers.json
Zoey2936: and why is this needed here?

Containers/nextcloud/cron.sh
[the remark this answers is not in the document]
szaimen: Thanks!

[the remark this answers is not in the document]
Zoey2936: So it seems the script continues first if the second (sleep 5m) finished and it ignores if the first command has finished or not in the current implementation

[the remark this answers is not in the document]
Zoey2936: so this PR can be closed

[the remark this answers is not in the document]
szaimen: > so this PR can be closed

All right

PR #2981, opened 2023-07-19
nextcloud container - make /tmp a tmpfs

szaimen:
to follow https://github.com/nextcloud/recognize#tmp

### DIFF
--- a/Containers/nextcloud/Dockerfile
+++ b/Containers/nextcloud/Dockerfile
@@ -209,6 +209,7 @@ RUN set -ex; \
     chown www-data:root -R /usr/src && \
     chown www-data:root -R /usr/local/etc/php/conf.d && \
     chown www-data:root -R /usr/local/etc/php-fpm.d && \
+    chmod -R 777 /tmp; \
     rm -r /usr/src/nextcloud/apps/updatenotification; \
     \
     mkdir -p /nc-updater; \
--- a/php/containers.json
+++ b/php/containers.json
@@ -209,6 +209,9 @@
       ],
       "networks": [
         "nextcloud-aio"
+      ],
+      "tmpfs": [
+        "/tmp"
       ]
     },
     {


Zoey2936 (approved):

szaimen:
This is now released with v6.4.0 Beta. Testing and feedback is welcome! See https://github.com/nextcloud/all-in-one#how-to-switch-the-channel

ricopt5:
`/tmp` doesn't have execute permissions, even with the `chmod 777` since by default docker mounts `tmpfs` with the `noexec` flag. 
This causes a problem for some apps like Memories since they need to put executables in `/tmp` and execute them.

In docker cli, we can just do `--tmpfs /mnt:exec`, but here I guess we need to make changes to `DockerActionManager.php` to support making `tmpfs` mounts executable.

szaimen:
Hi, the tmp directory is executable by everyone one my system. See 
image
Which docker version are you running and can you try to update your docker daemon to see if that resolves the issue?

ricopt5:
The direction has the executable bit set for owner, group and others, no issue there. But if you put an executable there and try to execute it, you will get a permission denied error.

But, for information `Docker version 24.0.2, build cb74dfcd85` on Arch Linux

szaimen:
I see. Thanks for the feedback!

szaimen:
PR is in https://github.com/nextcloud/all-in-one/pull/3047

szaimen:
This is now released with v7.0.0 Beta. Testing and feedback is welcome! See https://github.com/nextcloud/all-in-one#how-to-switch-the-channel

verement:
This directory is also used to store temporary files during uploads. Unfortunately this makes uploading large files difficult, as they are now limited to half the available memory.

How can we choose a different temporary directory for uploaded files, or otherwise choose not to use the tmpfs?

szaimen:
> This directory is also used to store temporary files during uploads. Unfortunately this makes uploading large files difficult, as they are now limited to half the available memory.

i have tested this and it wasnt the case on my test instance so you might run into a different iasue here.

verement:
Thanks for your reply. Could you elaborate? Are temporary files for uploads stored somewhere else in your test instance?

I installed Nextcloud AIO v7.0.0 fresh today, with a reverse proxy configuration. I created an upload-only shared folder, and attempted to upload files into it using the share link. When uploads of large files started failing, my investigation led me to discover the creation of temporary files in `/tmp` within the `nextcloud-aio-nextcloud` container, which I found is a tmpfs. The uploads failed just as the capacity of `/tmp` was exceeded, which is half the amount of available memory, and much less than the otherwise available disk space.

The Nextcloud log shows:

```
[PHP] Error: fread(): Write of 4096 bytes failed with errno=28 No space left on device at /var/www/html/3rdparty/icewind/streams/src/Wrapper.php#55
```

My question remains; how can we avoid temporary files for uploads being placed in the tmpfs?

szaimen:
Thanks for the heads-up! I just tried it myself again and came to the same conclusion. I will thus revert this change because uploads are more important than the speed benefit this brings.

szaimen:

kce68:
This finally made my day. Just tested with Beta7.2.0: This solved another issue i had with copying 4,5GB-Files from a federated share to my aio-Nextcloud instance. Thx!

verement:
v7.2.0 Beta indeed solves the upload issue I encountered, thanks. However, I will note that nextcloud/server#26179 is still an issue and I have several large-ish .part files remaining from each previously failed upload.

szaimen:
> owever, I will note that [nextcloud/server#26179](https://github.com/nextcloud/server/issues/26179) is still an issue and I have several large-ish .part files remaining from each previously failed upload.

This is unrelated to AIO and needs to be fixed in nextcloud/server.

PR #2984, opened 2023-07-19
change BrotliCompressionQuality to 1

szaimen:
To match https://github.com/nextcloud-snap/nextcloud-snap/blob/391dc6390f915a6974ce902388181bceb8a6cffd/src/apache/conf/httpd.conf#L59


Zoey2936:
but why? even the snap config links to the cloudflare blog which says 

> Although the quality setting in the Brotli implementation states the possible values are 0 to 11, we didn't see any differences between 0 and 1, or between 10 and 11, therefore only quality settings 1 to 10 are reported.

so why setting the level higher if the result is the same?

PR #3010, opened 2023-07-22
nextcloud container - add cache_dir

szaimen:
Close https://github.com/nextcloud/all-in-one/discussions/2924

- [ ] Todo: make some performancd tests. On aio-testing it should hopefully not become slower to upload a big file (also look at assembling times.) (sven better would be to test with datadir on different ssd).


szaimen:
Closing as speed advantage is not assured in all situation as the files need to be copied from the cache_dir to the datadir if both are not on thr same drive instead of just moving them which will be faster in some situation.

szaimen:
But reopening to at least make some tests

PR #3017, opened 2023-07-23
Update manual-upgrade.md shebang

sheikhmishar:
There was an exclamation (!) missing from shebang. As a result, execution of this script could result in unexpected behaviour


szaimen (approved):
Thanks!

PR #3018, opened 2023-07-23
mastercontainer - run supervisord directly and not as cmd option

szaimen:
Close https://github.com/nextcloud/all-in-one/discussions/3014

### DIFF
--- a/Containers/mastercontainer/start.sh
+++ b/Containers/mastercontainer/start.sh
@@ -318,5 +318,3 @@ chmod 777 /root
 
 # Start supervisord
 /usr/bin/supervisord -c /supervisord.conf
-
-exec "$@"
--- a/Containers/mastercontainer/start.sh
+++ b/Containers/mastercontainer/start.sh
@@ -26,6 +26,12 @@ if [ "$EUID" != "0" ]; then
     exit 1
 fi
 
+# Check that the CMD is not overwritten nor set
+if [ "$*" != "" ]; then
+    print_red "Docker run command for AIO is incorrect as a CMD option was given which is not expected."
+    exit 1
+fi
+
 # Check if socket is available and readable
 if ! [ -a "/var/run/docker.sock" ]; then
     print_red "Docker socket is not available. Cannot continue."
@@ -309,5 +315,8 @@ caddy fmt --overwrite /Caddyfile
 
 # Fix caddy log 
 chmod 777 /root
+
+# Start supervisord
+/usr/bin/supervisord -c /supervisord.conf
 
 exec "$@"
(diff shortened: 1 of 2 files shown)


Containers/mastercontainer/start.sh
```diff
@@ -26,6 +26,12 @@ if [ "$EUID" != "0" ]; then
     exit 1
 fi
 
+# Check that the given command is not set
+if [ "$*" != "" ]; then
```
szaimen, line 30: Not completely sure if that works. It should check the options that are given to start.sh

szaimen: should work after googling what it does.

Zoey2936: but why this check? if you remove exec at the end of the file, the given cmd will be ignored anyway

szaimen: yes it will be ignored and then the user complains that the setting does not work even though they used the wrong command which is why I would rather add this check. However we could still remove the exec at the end of the file. WDYT?

Zoey2936: the exec at the end should be removed, but I still don't understand this check

szaimen: The initial report in https://github.com/nextcloud/all-in-one/discussions/3014 must have happened because they must have used something like 

```
sudo docker run \
--sig-proxy=false \
--name nextcloud-aio-mastercontainer \
--restart always \
--publish 80:80 \
--publish 8080:8080 \
--publish 8443:8443 \
--volume nextcloud_aio_mastercontainer:/mnt/docker-aio-config \
--volume /var/run/docker.sock:/var/run/docker.sock:ro \
nextcloud/all-in-one:latest --env NEXTCLOUD_DATADIR="/mnt/ncdata/nextcloud"
```

If we now simply ignore this option without reporting back that it is incorrect, users will wonder why setting NEXTCLOUD_DATADIR does not work and will report a bug or open a thread. Reporting to them directly that this is wrong should hopefully work around this problem.

Zoey2936: I see then keep the check

Zoey2936: but exec at the end still need to be removed

szaimen: done

Zoey2936 (approved):

szaimen:
This is now released with v7.0.0 Beta. Testing and feedback is welcome! See https://github.com/nextcloud/all-in-one#how-to-switch-the-channel

PR #3049, opened 2023-07-28
remove export for HPB domain

enoch85:
It's not needed to export the `HPB_DOMAIN` as that would mean it's the same as `NC_DOMAIN`, which is not the case if `HPB_DOMAIN` is used. 

Was a bit to quick when reviewing, sorry.

In the VM we use both domains, as they are seperate: https://github.com/nextcloud/vm/blob/master/apps/talk.sh#L536-L537

### DIFF
--- a/Containers/talk-recording/start.sh
+++ b/Containers/talk-recording/start.sh
@@ -12,10 +12,6 @@ elif [ -z "$INTERNAL_SECRET" ]; then
     exit 1
 fi
 
-if [ -z "$HPB_DOMAIN" ]; then
-    export HPB_DOMAIN="$NC_DOMAIN"
-fi
-
 cat << RECORDING_CONF > "/conf/recording.conf"
 [logs]
 # 30 means Warning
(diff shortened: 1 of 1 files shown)


enoch85:
cc @szaimen

Zoey2936:
It is needed, since if HPB_DOMAIN is unset and only NC_DOMAIN is set, like it is the case for AIO, the config will be generated wrong

Containers/talk-recording/start.sh
```diff
@@ -13,7 +13,7 @@ elif [ -z "$INTERNAL_SECRET" ]; then
 fi
 
 if [ -z "$HPB_DOMAIN" ]; then
-    export HPB_DOMAIN="$NC_DOMAIN"
+    export HPB_DOMAIN
```
Zoey2936, line 16: I don't understand the sense of this, you want to say, if $HPB_DOMAIN is not set, then export it, without setting a value for it?

Zoey2936: the current logic is: if $HPB_DOMAIN is unset, it will be set to the same value as NC_DOMAIN

enoch85:
OK, I see what you mean now. Sorry, been in front of the computer for that past 14 hours today. Tired. :sleepy:

Containers/talk-recording/start.sh
enoch85: --- closed.

PR #3095, opened 2023-08-06
Update `depends_on:` of latest.yml

GuillaumeHullin:
Related discussion: #3085

Starting from docker-compose version v2.19.1 and newer, there is an enforced behavior where services listed under `depends_on:` are now **required** by default. This change cause issues when running a manual install without utilizing all the profiles. More information about this behavior can be found in the official documentation: https://docs.docker.com/compose/compose-file/05-services/#depends_on

To address this, I rewrote the `depends_on:` section using the long syntax and set `required: false`, excluding the services named _nextcloud-aio-nextcloud_ and _nextcloud-aio-database_ which I kept required.

Additionally, it might be worth considering setting `condition: service_healthy` for certain services, but I'm uncertain about the implications of doing so.

### DIFF
--- a/manual-install/latest.yml
+++ b/manual-install/latest.yml
@@ -1,11 +1,21 @@
 services:
   nextcloud-aio-apache:
     depends_on:
-      - nextcloud-aio-onlyoffice
-      - nextcloud-aio-collabora
-      - nextcloud-aio-talk
-      - nextcloud-aio-nextcloud
-      - nextcloud-aio-notify-push
+      nextcloud-aio-onlyoffice:
+        condition: service_started
+        required: false
+      nextcloud-aio-collabora:
+        condition: service_started
+        required: false
+      nextcloud-aio-talk:
+        condition: service_started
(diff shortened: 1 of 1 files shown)


Zoey2936:
we create the latest.yml using a script, your changes will be overriden, I've created https://github.com/nextcloud/all-in-one/pull/3096 to fix this, maybe you can review it

Zoey2936:
> Additionally, it might be worth considering setting `condition: service_healthy` for certain services, but I'm uncertain about the implications of doing so.

not needed, the containers check by them self if required containers are healthy

GuillaumeHullin:
Cool! I also adapted the yaml to have only `required: false`

szaimen:
Done with https://github.com/nextcloud/all-in-one/pull/3096. Thanks for the contribution though @GuillaumeHullin ! :)

GuillaumeHullin:
@szaimen sure, you're welcome!

PR #3118, opened 2023-08-09
update update-yaml.sh follow-up 

Zoey2936:


Zoey2936:
not sure if it works, then compose file has linebreaks I didn't expected https://github.com/nextcloud/all-in-one/pull/3119

szaimen:
This is how it would look like: https://github.com/nextcloud/all-in-one/pull/3120/files. Is the linebreak intended?

Zoey2936:
> This is how it would look like: https://github.com/nextcloud/all-in-one/pull/3120/files. Is the linebreak intended?

no they are not, but no Idea why they exist

szaimen:
> no they are not, but no Idea why they exist

All right, shall we try to find a workaround for this or is it fine?

szaimen:
Found the solution.

szaimen (approved):

PR #3128, opened 2023-08-09
Use .Values in Helm templates

bustapipes:
```
Error: template: nextcloud-aio-helm-chart/templates/nextcloud-aio-talk-recording-service.yaml:6:65: executing "nextcloud-aio-helm-chart/templates/nextcloud-aio-talk-recording-service.yaml" at <values>: wrong number of args for values: want 1 got 0
```


Zoey2936:
The helm chart is created using a script: https://github.com/nextcloud/all-in-one/blob/main/nextcloud-aio-helm-chart/update-helm.sh please modify this, otherwise your changes to the files will be removed with the next automatic update

Zoey2936:
I think szaimen will review this later, but what I can tell you, is that you should probably fix DCO

szaimen (approved):
Thank you!

PR #3140, opened 2023-08-10
add init flag to all containers

szaimen:
Close https://github.com/nextcloud/all-in-one/discussions/3052

Linked with https://github.com/nextcloud-releases/all-in-one/pull/42

### DIFF
--- a/php/containers.json
+++ b/php/containers.json
@@ -599,7 +599,7 @@
       "container_name": "nextcloud-aio-fulltextsearch",
       "display_name": "Fulltextsearch",
       "image": "nextcloud/aio-fulltextsearch",
-      "init": true,
+      "init": false,
       "expose": [
         "9200"
       ],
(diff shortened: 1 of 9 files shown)


php/containers.json
```diff
@@ -585,6 +599,7 @@
       "container_name": "nextcloud-aio-fulltextsearch",
       "display_name": "Fulltextsearch",
       "image": "nextcloud/aio-fulltextsearch",
+      "init": true,
```
szaimen, line 602: I wonder do we need this here since fts uses tiny as entrypoint...

Zoey2936: yes, not needed https://github.com/elastic/dockerfiles/blob/cf5707cefdfc1dba3e95f7d15931aeba5db93f0f/elasticsearch/Dockerfile#L160

szaimen:
```suggestion
```

PR #3153, opened 2023-08-14
add check for timezone

szaimen:
Close https://github.com/nextcloud/all-in-one/discussions/3147

### DIFF
--- a/Containers/mastercontainer/start.sh
+++ b/Containers/mastercontainer/start.sh
@@ -256,12 +256,15 @@ fi
 # Check that no changes have been made to timezone settings since AIO only supports running in UTC timezone
 if [ -n "$TZ" ]; then
     print_red "The environmental variable TZ has been set which is not supported by AIO since it only supports running in the default UTC timezone!"
+    echo "The correct timezone can be set in the AIO interface later on!"
     exit 1
 elif mountpoint -q /etc/localtime; then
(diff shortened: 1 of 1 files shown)


Zoey2936:
maybe also mention, that the timezone should be set inside the aio panel

Zoey2936:
and also is doing exit 1 a good idea?

szaimen:
> maybe also mention, that the timezone should be set inside the aio panel

Good idea!

> and also is doing exit 1 a good idea?

Yes, I don't think many instances have set this. If there should be many reports, we can still remove the exit later.

captaindyson:
This change has caused issues on Unraid. the os adds a TZ to all docker run commands by default, hence preventing the running of the container.
Manualy adding a TZ variable with no value seems to override this, but is not ideal, or at least will require clearly communicating.

szaimen:
Interesting! I guess we will need to fix this then in order to not break all unraid instances.

szaimen:
> This change has caused issues on Unraid. the os adds a TZ to all docker run commands by default, hence preventing the running of the container.

This is now fixed with v7.1.1 Beta. Testing and feedback is welcome! See https://github.com/nextcloud/all-in-one#how-to-switch-the-channel

captaindyson:
Can confirm this has fixed the issue, warning remains in the logs, but the container starts anyway.

PR #3154, opened 2023-08-14
add check for unsupported environmental variables

szaimen:


Containers/mastercontainer/start.sh
```diff
@@ -253,6 +253,17 @@ if [ "$?" = 6 ]; then
     exit 1
 fi
 
+# Check if unsupported env are set (but dont exit as it would break many instances)
```
Zoey2936, line 256:
```suggestion
# Check if unsupported env are set (but don't exit as it would break many instances)
```

Zoey2936 (approved):

PR #3207, opened 2023-08-18
nextcloud - improve healthcheck

szaimen:
Close part of https://github.com/nextcloud/all-in-one/discussions/3197

### DIFF
--- a/.github/workflows/docker-lint.yml
+++ b/.github/workflows/docker-lint.yml
@@ -32,10 +32,10 @@ jobs:
           sudo apt-get update
           sudo apt-get install nodejs npm -y --no-install-recommends
           npm install -g dockerfilelint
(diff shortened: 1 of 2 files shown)


szaimen:
@frederikb96 does this look good to you?

frederikb96:
> @frederikb96 does this look good to you? :)

Look's nice, thanks a lot for such a fast reply and fix :)

Zoey2936 (changes requested):
I would instead change the start-period https://docs.docker.com/engine/reference/builder/#healthcheck

szaimen:
> I would instead change the start-period https://docs.docker.com/engine/reference/builder/#healthcheck

Good idea, thanks!

Zoey2936 (approved):

szaimen:
Hm... I wonder how to fix the docker Lint workflow...

szaimen:
Seems like we need another patch: https://github.com/replicatedhq/dockerfilelint/pull/167/files

PR #3214, opened 2023-08-21
nextcloud - add stop_grace_period of 600s and wait for cron to finish during that time

szaimen:
Fix https://github.com/nextcloud/recognize/discussions/821 partially?

### DIFF
--- a/Containers/nextcloud/cron.sh
+++ b/Containers/nextcloud/cron.sh
@@ -1,7 +1,20 @@
 #!/bin/bash
 set -eu
 
+wait_for_cron() {
+    set -x
+    while [ -n "$(pgrep -f /var/www/html/cron.php)" ]; do
+        echo "Waiting for cron to stop..."
+        sleep 5
+    done
+    echo "Cronjob successfully exited."
+    set +x
+}
+
+trap wait_for_cron SIGINT SIGTERM
+
 while true; do
     php -f /var/www/html/cron.php &
-    sleep 5m
+    sleep 5m &
+    wait $!
 done
--- a/php/containers.json
+++ b/php/containers.json
@@ -205,6 +205,7 @@
         "TALK_RECORDING_HOST=nextcloud-aio-talk-recording",
         "FULLTEXTSEARCH_PASSWORD=%FULLTEXTSEARCH_PASSWORD%"
       ],
+      "stop_grace_period": 600,
       "restart": "unless-stopped",
       "devices": [
         "/dev/dri"


Containers/nextcloud/cron.sh
```diff
@@ -1,7 +1,11 @@
...
 
+php -f /var/www/html/cron.php &
+
 while true; do
-    php -f /var/www/html/cron.php &
     sleep 5m
+    php -f /var/www/html/cron.php &
 done
+
+wait $!
```
szaimen, line 11: @Zoey2936 I suppose this is not going to work?

Zoey2936: no

szaimen: Could you imagine a syntax that we could use here that would fix the problem? Possibly adding a wrapper script or so?

szaimen:
@Zoey2936 do you think this could work now? :)

Containers/nextcloud/cron.sh
```diff
@@ -1,7 +1,10 @@
...
-    php -f /var/www/html/cron.php &
-    sleep 5m
-done
+trap 'true' SIGINT SIGTERM
+
+exec /cronjob.sh &
+
+wait $!
+
+echo "Cronjob successfully exited."
```
Zoey2936, line 10: what does wait $! does diffrent to just wait?

szaimen: Honestly not sure

Zoey2936: and where do you got this?

szaimen: From here: https://github.com/nextcloud/all-in-one/blob/722a96828e0a5b69cb9fcda5c14b75df3035a0c8/Containers/postgresql/start.sh#L169

Zoey2936: and where does this come from?

szaimen: I think somehwere from stackoverflow or so...

szaimen: but did not note that down...

Zoey2936: from what I know is wait either waits for a specified pid to exit or "all" (I think all launched by the shell script)

szaimen: okay, but that would be what we want here or not?

szaimen:
I tried it locally and it doesnt seem to work yet

Containers/nextcloud/cron.sh
foxhoundv: > I think somehwere from stackoverflow or so...

https://stackoverflow.com/questions/33375924/what-does-wait-mean

The answer was "($!) Expands to the process ID of the job most recently placed into the background, whether executed as an asynchronous command or using the bg builtin"

foxhoundv: I don't know if this will help but...

https://superuser.com/questions/959508/cron-job-does-not-wait-end-of-process

They are suggesting to remove "wait " because it will be obsolete. Instead to use "sleep 2".

FlattusBlastus: I hope this will work in Docker Desktop for Windows. As far as I can tell, DDfW gets TASKKILL and says "Ok" instead of wait x number of seconds. Might have to send this up to Docker.

ShapeShifter499:
Is the idea to wait for all cron jobs to complete successfully?  Does this mean automatic updates will be stalled till the cron jobs complete?

Containers/nextcloud/cron.sh
marcelklehr: This seems more appropriate: https://stackoverflow.com/questions/356100/how-to-wait-in-bash-for-several-subprocesses-to-finish-and-return-exit-code-0

szaimen: Will have a look, thanks!

marcelklehr: You could then put those wait calls in a cleanup function and have it called by the trap

szaimen:
This seems to work now finally :)

Containers/nextcloud/cron.sh
```diff
@@ -1,7 +1,20 @@
...
+    set +x
+}
+
+trap wait_for_cron SIGINT SIGTERM
+
 while true; do
     php -f /var/www/html/cron.php &
-    sleep 5m
+    sleep 5m &
+    wait $!
```
marcelklehr, line 19: This line seems not necessary. Worse, if a cron job takes a long time, it will block others from getting executed

szaimen: I think this wait only waits for the sleep to be finished. We need this as otherwise the wait_for_cron is only executed after the last sleep 5m was finished...

szaimen: See https://stackoverflow.com/a/33375971

marcelklehr (approved):

PR #3232, opened 2023-08-23
Fix OO

szaimen:
Close https://github.com/nextcloud/all-in-one/discussions/3224


Containers/nextcloud/entrypoint.sh
```diff
@@ -583,6 +583,7 @@ if [ "$ONLYOFFICE_ENABLED" = 'yes' ]; then
     php /var/www/html/occ config:system:set onlyoffice jwt_header --value="AuthorizationJwt"
     php /var/www/html/occ config:app:set onlyoffice DocumentServerUrl --value="https://$NC_DOMAIN/onlyoffice"
     php /var/www/html/occ config:system:set allow_local_remote_servers --type=bool --value=true
+    php /var/www/html/occ onlyoffice:documentserver --check
```
szaimen, line 586: Needs to be done after everything started up... So like with the richdocuments app...

Issue #3239, opened 2023-08-24
disable selinux for watchtower and docker-socket-proxy


szaimen:
https://github.com/nextcloud/all-in-one#are-there-known-problems-when-selinux-is-enabled

_Originally posted by @Zoey2936 in https://github.com/nextcloud/all-in-one/pull/3235#discussion_r1304314326_

PR #3240, opened 2023-08-24
disable selinux for watchtower and docker-socket-proxy

szaimen:
Close https://github.com/nextcloud/all-in-one/issues/3239

### DIFF
--- a/php/src/Docker/DockerActionManager.php
+++ b/php/src/Docker/DockerActionManager.php
@@ -499,6 +499,9 @@ class DockerActionManager
                 }
                 $mounts[] = ["Type" => "bind", "Source" => $volume->name, "Target" => $volume->mountPoint, "ReadOnly" => !$volume->isWritable, "BindOptions" => [ "Propagation" => "rshared"]];
             }
+        // Special things for the watchtower and docker-socket-proxy container which should not be exposed in the containers.json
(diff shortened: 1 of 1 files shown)


php/src/Docker/DockerActionManager.php
```diff
@@ -499,6 +499,9 @@ public function CreateContainer(Container $container) : void {
                 }
                 $mounts[] = ["Type" => "bind", "Source" => $volume->name, "Target" => $volume->mountPoint, "ReadOnly" => !$volume->isWritable, "BindOptions" => [ "Propagation" => "rshared"]];
             }
+        // Special things for the watchtower and docker-socket-proxy container which should not be exposed in the containers.json
```
Zoey2936, line 502: can I ask why not using the conatiners.json?

Zoey2936: and why not doing that?

szaimen: because usually I am only adding stuff there that could be useful for further components like for https://github.com/nextcloud/all-in-one/discussions/1581. But I dont see this making sense for that...

szaimen: We can also still add it later

Zoey2936 (approved):

PR #3246, opened 2023-08-25
try to fix update-yaml.sh

szaimen:


manual-install/update-yaml.sh
```diff
@@ -21,7 +21,7 @@ OUTPUT="$(echo "$OUTPUT" | jq 'del(.services[] | select(.container_name == "next
 OUTPUT="$(echo "$OUTPUT" | jq 'del(.services[] | select(.container_name == "nextcloud-aio-domaincheck"))')"
 OUTPUT="$(echo "$OUTPUT" | jq 'del(.services[] | select(.container_name == "nextcloud-aio-borgbackup"))')"
 OUTPUT="$(echo "$OUTPUT" | jq 'del(.services[] | select(.container_name == "nextcloud-aio-docker-socket-proxy"))')"
-OUTPUT="$(echo "$OUTPUT" | jq '.services[].depends_on |= if contains(["nextcloud-aio-docker-socket-proxy"]) then del(.[index("nextcloud-aio-docker-socket-proxy")])')"
+OUTPUT="$(echo "$OUTPUT" | jq '.services[].depends_on |=  if contains(["nextcloud-aio-docker-socket-proxy"]) then del(.[index("nextcloud-aio-docker-socket-proxy")])')"
```
szaimen, line 24: @Zoey2936 do you know how to fix this logic? This somehow still does not seem to work...

Zoey2936: I think there is a "else ." missing

szaimen: Oh it requires an else? 🤦‍♂️

Zoey2936: yes

Zoey2936: and an "end", so it needs to end with "else . end'"

szaimen: let me try this. thanks!

szaimen: Still seems to fail :/

image

szaimen: I need to probably first check if depends_on is there and then check if it contains something?

szaimen:
Fixed it :)

PR #3265, opened 2023-08-27
backupscript.sh: slight rewording for consistency

sjjh:
Added some full stops, changed some wording, changed capitalisation, ...in some "echo"-statements  to improve consistency for user outputs.

### DIFF
--- a/Containers/borgbackup/backupscript.sh
+++ b/Containers/borgbackup/backupscript.sh
@@ -47,7 +47,7 @@ fi
 
 # Do not continue if this file exists (needed for simple external blocking)
 if [ -f "$BORG_BACKUP_DIRECTORY/aio-lockfile" ]; then
-    echo "Not continuing because aio-lockfile exists -- it seems like a script is externally running which is locking the backup archive."
+    echo "Not continuing because aio-lockfile exists – it seems like a script is externally running which is locking the backup archive."
     echo "If this should not be the case, you can fix this by deleting the 'aio-lockfile' file from the backup archive directory."
     exit 1
 fi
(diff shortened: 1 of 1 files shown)


Containers/borgbackup/backupscript.sh
```diff
@@ -35,19 +35,19 @@ done
...
 if [ "$BORG_MODE" != backup ] && [ "$BORG_MODE" != test ] && ! [ -f "$BORG_BACKUP_DIRECTORY/config" ]; then
-    echo "The repository is empty. cannot perform check or restore."
+    echo "The repository is empty. Cannot perform check or restore."
     exit 1
 fi
 
 # Do not continue if this file exists (needed for simple external blocking)
 if [ -f "$BORG_BACKUP_DIRECTORY/aio-lockfile" ]; then
-    echo "Not continuing because aio-lockfile exists - it seems like a script is externally running which is locking the backup archive."
+    echo "Not continuing because aio-lockfile exists -- it seems like a script is externally running which is locking the backup archive."
```
Zoey2936, line 50: i don't understand this additional -

sjjh: Sry, me being a typography nerd here out of habit. According to my knowledge, the correct punctuation mark to be used here is an (spaced) [en-dash](https://en.wikipedia.org/wiki/Dash#En_dash) (or an unspaced em-dash) and not a [hyphen](https://en.wikipedia.org/wiki/Hyphen). At least in LaTeX `--` is replaced by an en-dash (`–`) (and `---` by an em-dash (`—`)), I assumed in markdown this might be the case as well. If utf-8 is used, probably a en-dash (–) can be used directly as well.

Zoey2936: I have no idea. English is not my native language

szaimen:
```suggestion
    echo "Not continuing because aio-lockfile exists – it seems like a script is externally running which is locking the backup archive."
```

szaimen (approved):
Thank you!

Issue #3266, opened 2023-08-27
Enrich documentation about reverse proxy with information about ssl certs for connection between reverse proxy and nextcloud


sjjh:
### Is your feature request related to a problem? Please describe. 

I'm trying to set up a new AIO nextcloud installation behind an existing reverse proxy. The reverse proxy is providing a LE cert for the public domain and terminates the https connection. The connection between the reverse proxy and the AIO NC installation should be secured by https as well, using a self-signed cert (this is how we did setup other services, e.g. discourse as well).
Reading the [documentation about reverse proxy](https://github.com/nextcloud/all-in-one/blob/main/reverse-proxy.md) it states to use `--env APACHE_PORT=11000 \`. The documentation does not describe if the port expects HTTP or HTTPS traffic. The documentation does not describe how to specify the (self-signed) certs to the container.
### Describe the solution you'd like

The documentation should:
- describe if HTTP or HTTPS traffic is expected at the specified port (`--env APACHE_PORT=11000 \`)
- *if HTTP traffic is expected* describe how to configure the container to expect HTTPS traffic
- describe how to specify a (self-signed) cert to the container
### Describe alternatives you've considered

I searched the other documentation of NC AIO without luck and looked at the community forum for help, but did only find one [not matching post](https://help.nextcloud.com/t/false-self-signed-certificate-error-with-reverse-proxy-for-aio/148350).
### Additional context

Context to the environment: NC AOI will run on a raspberry pi. The reverse proxy is an nginx, running as a plug-in on a OPNsense firewall.
For reference/inspiration: With Discourse you can copy the certs to a a folder that gets mounted as a volume inside the docker container. The yml-file can be altered, to specify the volume and file names of the certs.
If my hypothesis that this use case (reverse proxy terminates HTTPS connection, additional HTTPS connection using self-signed certificates between reverse proxy and NC AIO ) is not supported by now (I'd assume that it is a quite common setup), I appologize and we can change the issue from "existing feature/documentation enhancement" to "new feature request".

Zoey2936:
1. http
2. switching to https would break existing instaces
3. there could be an env added to switch to https, but that is nothing I can decide

szaimen:
> additional HTTPS connection using self-signed certificates between reverse proxy and NC AIO ) is not supported by now

Correct. Background is that the reverse prpxy documentation is already complicated and I dont want to make it even more complicated by adding more options.

However there are two workarounds that I see for your use case:
[...]

PR #3273, opened 2023-08-28
Update reverse-proxy.md: clarified that unencrypted HTTP traffic is e…

sjjh:
…xpected

response to #3266
I could imagine it would be even clearer to explicitly state that HTTPS is (currently) not supported and reference to #3268 but I didn't want to make it too verbose (as @szaimen mentioned that the documentation is already quite long).


szaimen (approved):
Fine by me

PR #3298, opened 2023-09-04
do not exit if TZ is set

szaimen:
Address https://github.com/nextcloud/all-in-one/pull/3153#issuecomment-1704402550

### DIFF
--- a/Containers/mastercontainer/start.sh
+++ b/Containers/mastercontainer/start.sh
@@ -253,17 +253,20 @@ if [ "$?" = 6 ]; then
     exit 1
 fi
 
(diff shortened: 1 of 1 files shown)


Zoey2936:
maybe run `unset TZ` (not sure if it was `unset`)

szaimen:
> maybe run `unset TZ` (not sure if it was `unset`)

It gets overwritten here so should be fine?
https://github.com/nextcloud/all-in-one/blob/af10e2822552d9eb361f554bcf3be9fd96bab3b7/Containers/mastercontainer/start.sh#L334

Zoey2936:
ok

PR #3313, opened 2023-09-07
Update check - move to head request instead of get request

szaimen:
Motivation: head requests are not counted in the pull limit afaik.

Also for https://github.com/nextcloud/all-in-one/discussions/2667


Zoey2936 (approved):

SystemKeeper (approved):
Makes sense 👍 
Also see https://docs.docker.com/docker-hub/download-rate-limit/#definition-of-limits for reference about the definition.

szaimen:

Issue #3172, opened 2023-08-16
Make mastercontainer and nextcloud container more secure by setting listen.allowed_clients


szaimen:
- [x] mastercontainer: https://github.com/nextcloud/all-in-one/pull/3317

```
/usr/local/etc/php-fpm.d/www.conf

;listen.allowed_clients = 127.0.0.1,::1
```

- [ ] nextcloud container: https://github.com/nextcloud/all-in-one/pull/3318
```
/usr/local/etc/php-fpm.d/www.conf

;listen.allowed_clients = ipv4-of-apache;ipv6-of-apache,ipv4-of-mastercontainer-ipv6-of-mastercontainer,localhost,notify-push,dsp
```

Zoey2936:
notify push connects to php-fpm?

Zoey2936:
but does it not only do a `nc` check? I think it wouldn't even see if there is no permission

szaimen:
> but does it not only do a `nc` check? I think it wouldn't even see if there is no permission

It does, I tested and it shows even when only doing a nc check

szaimen:
> for the mastercontainer: what about switching to unix sockets?

This would be an alternative idea 👍

szaimen:
a PR would be welcome, btw! :)

szaimen:
This is now released with v7.4.0 Beta. Testing and feedback is welcome! See https://github.com/nextcloud/all-in-one#how-to-switch-the-channel

PR #3317, opened 2023-09-07
mastercontainer - limit access to php-fpm to localhost

szaimen:
address part of https://github.com/nextcloud/all-in-one/issues/3172

### DIFF
--- a/Containers/mastercontainer/Dockerfile
+++ b/Containers/mastercontainer/Dockerfile
@@ -56,6 +56,8 @@ RUN set -ex; \
     sed -i 's/^pm = dynamic/pm = ondemand/' /usr/local/etc/php-fpm.d/www.conf; \
     sed -i 's/^pm.max_children =.*/pm.max_children = 80/' /usr/local/etc/php-fpm.d/www.conf; \
     sed -i 's|access.log = /proc/self/fd/2|access.log = /proc/self/fd/1|' /usr/local/etc/php-fpm.d/docker.conf; \
+    grep -q ';listen.allowed_clients' /usr/local/etc/php-fpm.d/www.conf; \
+    sed -i 's|;listen.allowed_clients.*|listen.allowed_clients = 127.0.0.1,::1|' /usr/local/etc/php-fpm.d/www.conf; \
     \
     apk add --no-cache git; \
     wget https://getcomposer.org/installer -O - | php -- --install-dir=/usr/local/bin --filename=composer; \


Containers/mastercontainer/Dockerfile
```diff
@@ -56,6 +56,8 @@ RUN set -ex; \
     sed -i 's/^pm = dynamic/pm = ondemand/' /usr/local/etc/php-fpm.d/www.conf; \
     sed -i 's/^pm.max_children =.*/pm.max_children = 80/' /usr/local/etc/php-fpm.d/www.conf; \
     sed -i 's|access.log = /proc/self/fd/2|access.log = /proc/self/fd/1|' /usr/local/etc/php-fpm.d/docker.conf; \
+    grep -q ';listen.allowed_clients' /usr/local/etc/php-fpm.d/www.conf; \
+    sed -i 's|;listen.allowed_clients.*|listen.allowed_clients = 127.0.0.1,::1|' /usr/local/etc/php-fpm.d/www.conf; \
```
Zoey2936, line 60:
```suggestion
    grep -q 'listen =' /usr/local/etc/php-fpm.d/www.conf; \
    sed -i 's|listen =.*|listen = /var/run/php.sock|' /usr/local/etc/php-fpm.d/www.conf; \
```

And change the Apache PHP handler to: `SetHandler "proxy:unix:/var/run/php.sock"`

Zoey2936:
also not needed, since php fpm is by default only exposed to 127.0.0.1 if I remember correctly

Zoey2936:
```console
/var/www/html # cat /usr/local/etc/php-fpm.d/www.conf | grep listen | grep -v ";"
listen = 127.0.0.1:9000
```

szaimen:
> ```
> /var/www/html # cat /usr/local/etc/php-fpm.d/www.conf | grep listen | grep -v ";"
> listen = 127.0.0.1:9000
> ```

It is the same in the nextcloud container and there access from a different container is obiously possible so switching to a socket probably makes sense...

Containers/mastercontainer/Dockerfile
szaimen: all right, lets do it like this then

Zoey2936: https://github.com/docker-library/php/blob/e9e47fd185d6b36e491c713811437b0c17944faa/8.2/alpine3.18/fpm/Dockerfile#L249

```suggestion
    grep -q 'listen =' /usr/local/etc/php-fpm.d/www.conf; \
    sed -i 's|listen =.*|;listen = /var/run/php.sock # handled in zz-docker.conf|' /usr/local/etc/php-fpm.d/www.conf; \
    grep -q 'listen =' /usr/local/etc/php-fpm.d/zz-docker.conf; \
    sed -i 's|listen =.*|listen = /var/run/php.sock|' /usr/local/etc/php-fpm.d/zz-docker.conf; \
```

And change the Apache PHP handler to: `SetHandler "proxy:unix:/var/run/php.sock"`

Zoey2936: healthcheck also needs to check if /var/run/php.sock exists

szaimen: or we just actually limit it to 127.0.0.1:::1? would that be fine for you? Sounds like the easier approach...

szaimen: due to the performance improvement?

Zoey2936: https://serverfault.com/questions/124517/what-is-the-difference-between-unix-sockets-and-tcp-ip-sockets unix sockets also seems to be faster

Zoey2936: I don't see why it is simpler to limit ips, instead of using a socket

szaimen: 2 lines of code vs 4? but yeah if you really want we can also use a socket

Zoey2936: I would use the socket

szaimen: @Zoey2936 lets switch to the socket in a follow-up.

I would welcome a PR :)

szaimen:

Issue #3354, opened 2023-09-16
Full text search first indexing failed with "NoNodeAvailableException" 


vidlb:
The installation is working like a charm with fulltextsearch disabled.

### Steps to reproduce

1. Enable FTS for the first time
2. Launch containers
=> Failure after 40K files and ~ 137 Mb usage 

### Expected behavior
Indexing should end without errors, there is enough RAM available in the system.

### Actual behavior
Errors will continue until all users have been processed. The first index can never end completely.

### Host OS
Debian 12 with latest docker, lots of RAM and disk space

#### Nextcloud AIO version
27.1.1 and 27.2.1

#### Current channel
Stable

PR #3355, opened 2023-09-16
FIX: use 1GB for FTS max JVM memory usage

vidlb:
Fix for #3354, this is the change I had to make in order for FTS to complete first indexing successfully (following this [discussion](https://github.com/nextcloud/all-in-one/discussions/2509) to re-index).  
Also, in the docs it is stated that FTS requires 1GB additional RAM, so I guess this -Xmx512M was not intended ?  

Fix https://github.com/nextcloud/all-in-one/discussions/3373

### DIFF
--- a/php/containers.json
+++ b/php/containers.json
@@ -612,7 +612,7 @@
       "internal_port": "9200",
       "environment": [
         "TZ=%TIMEZONE%",
-        "ES_JAVA_OPTS=-Xms512M -Xmx512M",
+        "ES_JAVA_OPTS=-Xms512M -Xmx1024M",
         "bootstrap.memory_lock=true",
         "cluster.name=nextcloud-aio",
         "discovery.type=single-node",


Zoey2936:
> Also, in the docs it is stated that FTS requires 1GB additional RAM, so I guess this -Xmx512M was not intended ?

It is java, it will always use more ram than you specify...

vidlb:
I don't know may be `-Xms256M  -Xmx512M` would do the trick ?  
I mean can identical values for min and max be a problem somehow ?

I tried several things, this change is the only thing that made it work

vidlb:
>It is java, it will always use more ram than you specify...

Indeed I just checked with top the java process is using 2GB, what the hell !  
And it's 3GB in total for the elasticsearch container.

php/containers.json
```diff
@@ -612,7 +612,7 @@
       "internal_port": "9200",
       "environment": [
         "TZ=%TIMEZONE%",
-        "ES_JAVA_OPTS=-Xms512M -Xmx512M",
+        "ES_JAVA_OPTS=-Xms512M -Xmx1024M",
```
szaimen, line 615: I would be fine with this. Wdyt @Zoey2936?
```suggestion
        "ES_JAVA_OPTS=-Xms1024M -Xmx1024M",
```

Zoey2936: we can do this, but it will use a lot of ram

vidlb: @szaimen I don't understand why you would rewrite this with the same value for initial and max heap size, I've always seen this with Xms < Xmx.  
I'm almost sure this is the source of the problem, when java cannot allocate more than initial (reserved) allocated size, it seems elasticsearch cannot spawn new nodes for indexing.  

I could try `-Xms256M -Xmx512M` but it's very long to re-index...
I have a lot of files and groupfolders, but also ignored dirs, yet I'm pretty sure I don't need 1GB or 2GB of RAM to do this. I don't know about the memory usage printed by occ, but I've never seen it exceed 200 or 300MB.

vidlb: > we can do this, but it will use a lot of ram

To avoid high memory usage even if not that much is required, shouldn't we use a smaller value like `-Xms64m -Xmx1024M` ?

vidlb: I have another idea : may be I should close this, in favor of another issue for feature request : for a `ES_JAVA_OPTS` env that can be set in the compose file ?  

In my case the problem is probably also related to groupfolders, which fulltextsearch does not seems understand really well and I wouldn't be surprised if files are indexed as many times as they are users. It's not clears from the logs if there's just an entry per user (file path) or if file is actually parsed every time.  
But I don't think this high memory usage is normal in regard to the number of files I have to index here.

szaimen: Hm... 512m and 512m are the values recommended in https://github.com/R0Wi/elasticsearch-nextcloud-docker#how-to-use-this so not sure increasing the value is the way to go. However I would really like to not introduce a variable here. So maybe the fulltextsearch app needs to be improved instead in order to improve the handling of groupfolders instead?

vidlb: As you want, yet if I were you I wouldn't take anything for granted except if it comes from the official ElasticSearch docs.  

But if this does not happen for users without groupfolders, I agree here isn't the place to make that change.  

Yet I don't understand why want to avoid a variable, since this setting has to be set according to number of files to scan, you might see this exception pop in future issues !  

At least it should be added to the docs which file has to be modified (the `containers.json` in order for the first indexing to succeed. This is not a problem if the container then re-run with Xmx512M, because it's only happening during a first run (or manual `occ fulltextsearch:index`.

vidlb: Well I just gave a quick look at ES docs, they show the following example for memory settings in docker. So yeah I guess my first argument is broken ^^

`docker run -e ES_JAVA_OPTS="-Xms1g -Xmx1g" -e ENROLLMENT_TOKEN="<token>" --name es01 -p 9200:9200 --net elastic -it docker.elastic.co/elasticsearch/elasticsearch:8.9.2`

vidlb: But interestingly, the following is stated, just under this example : 
>The ES_JAVA_OPTS variable overrides all other JVM options. We do not recommend using ES_JAVA_OPTS in production.

Zoey2936: running java without xms/xmx will make java to eat basically all ram

vidlb: Yes, I'm confused by this statement, if we run in docker, isn't it for ES java process to be alone in its env ? So I don't see how ES_JAVA_OPTS could override other JVM options.. how many JVM processes are there in this container 😬

PR #3375, opened 2023-09-18
Added link for how to set chunking on Nextcloud Desktop Client

jahanson:
I was unsure how to chunk uploads and whether or not it was a client or server problem. This bullet may help people in the future trying to upload large files behind a Cloudflare Tunnel.


szaimen:
Hi, thanks for your PR! However I think it would better fit in this section: https://github.com/nextcloud/all-in-one#notes-on-cloudflare-proxytunnel

jahanson:
That makes more sense I went ahead and moved the bullet to a more relevant location.

PR #3377, opened 2023-09-19
drop NET_RAW from all containers

szaimen:
Close https://github.com/nextcloud/all-in-one/issues/3366


php/src/Docker/DockerActionManager.php
```diff
@@ -476,6 +476,8 @@ public function CreateContainer(Container $container) : void {
             $requestBody['HostConfig']['CapAdd'] = $capAdds;
         }
```
szaimen, line 478: Add comment here why this was added

PR #3069, opened 2023-08-02
improve webserver conf mastercontainer

Zoey2936:

PR #3096, opened 2023-08-06
Update update-yaml.sh

Zoey2936:
see: https://github.com/nextcloud/all-in-one/discussions/3085#discussioncomment-6649658

Close https://github.com/nextcloud/all-in-one/issues/3088
Close https://github.com/nextcloud/all-in-one/issues/3085

PR #3416, opened 2023-09-26
mastercontainer - use Caddy for generating self-singed cert

szaimen:
Alternative approach to https://github.com/nextcloud/all-in-one/pull/3069
Supercedes https://github.com/nextcloud/all-in-one/pull/3069

### DIFF
--- a/Containers/mastercontainer/Caddyfile
+++ b/Containers/mastercontainer/Caddyfile
@@ -25,7 +25,9 @@ http://:80 {
 }
 
 https://:8080 {
-    reverse_proxy localhost:8000
+    reverse_proxy localhost:8000 {
+        trusted_proxies private_ranges
+    }
     tls internal {
         on_demand
     }
--- a/Containers/mastercontainer/Caddyfile
+++ b/Containers/mastercontainer/Caddyfile
@@ -21,7 +21,14 @@
 }
 
 http://:80 {
-  redir https://{host}{uri}
+    redir https://{host}{uri}
+}
+
+https://:8080 {
+    reverse_proxy localhost:8000
+    tls internal {
+        on_demand
+    }
 }
 
 https://:8443 {
--- a/Containers/mastercontainer/Dockerfile
+++ b/Containers/mastercontainer/Dockerfile
@@ -23,6 +23,7 @@ RUN set -ex; \
     usermod -u 333 -g 333 xfs; \
     groupmod -g 33 www-data; \
     usermod -u 33 -g 33 www-data; \
+    apk del --no-cache shadow; \
     \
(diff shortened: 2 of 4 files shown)


Containers/mastercontainer/Caddyfile
```diff
@@ -21,7 +21,14 @@
...
 
 http://:80 {
-  redir https://{host}{uri}
+    redir https://{host}{uri}
+}
+
+https://:8080 {
+    reverse_proxy localhost:8000
+    tls internal {
+        on_demand
```
szaimen, line 30: This needs double check if also ip-addresses get accepted here...

szaimen:
@Zoey2936 would this approach be fine with you?

Containers/mastercontainer/Caddyfile
Zoey2936: yes

Zoey2936:
> @Zoey2936 would this approach be fine for you?

what is different to #3096?

Containers/mastercontainer/Caddyfile
szaimen: should be done with https://github.com/nextcloud/all-in-one/pull/3416/commits/9a0fd6d0a21530a54cf196ab00ab7c94ee511164

szaimen:
> > @Zoey2936 would this approach be fine for you?
> 
> what is different to #3096?

Mainly separating the change to caddy from moving to a full-blown apache conf which is better to test

Containers/mastercontainer/Caddyfile
Zoey2936: is this not a global option? https://caddyserver.com/docs/caddyfile/options#trusted-proxies

szaimen: I think it is both...

szaimen: https://caddyserver.com/docs/caddyfile/directives/reverse_proxy#the-http-transport

Zoey2936 (approved):

szaimen:
Thanks :)

szaimen:
I fear this does not work. I get the following issue:

```
automation policy from site block is also default/catch-all policy because of key without hostname, and the two are in conflict: []certmagic.Issuer{(*caddytls.InternalIssuer)(0xc00032c360)} != []certmagic.Issuer{(*caddytls.ACMEIssuer)(0xc0003fce00)}
```

So reverting.

szaimen:
This is now released with v7.4.0 Beta. Testing and feedback is welcome! See https://github.com/nextcloud/all-in-one#how-to-switch-the-channel

PR #3427, opened 2023-09-27
adjust limiting the php-fpm port to specific containers again

szaimen:
Close https://github.com/nextcloud/all-in-one/issues/3172

### DIFF
--- a/Containers/nextcloud/start.sh
+++ b/Containers/nextcloud/start.sh
@@ -131,4 +131,17 @@ if ! sudo -E -u www-data bash /entrypoint.sh; then
     exit 1
 fi
 
+while [ -z "$(dig nextcloud-aio-apache A +short)" ]; do
+    echo "Waiting for nextcloud-aio-apache to start..."
+    sleep 5
+done
+IPv4_ADDRESS_APACHE="$(dig nextcloud-aio-apache A +short | grep '^[0-9.]\+$' | sort | head -n1)"
+IPv6_ADDRESS_APACHE="$(dig nextcloud-aio-apache AAAA +short | grep '^[0-9a-f:]\+$' | sort | head -n1)"
+IPv4_ADDRESS_MASTERCONTAINER="$(dig nextcloud-aio-mastercontainer A +short | grep '^[0-9.]\+$' | sort | head -n1)"
+IPv6_ADDRESS_MASTERCONTAINER="$(dig nextcloud-aio-mastercontainer AAAA +short | grep '^[0-9a-f:]\+$' | sort | head -n1)"
+
+sed -i "s|^;listen.allowed_clients|listen.allowed_clients|" /usr/local/etc/php-fpm.d/www.conf
+sed -i "s|listen.allowed_clients.*|listen.allowed_clients = 127.0.0.1,::1,$IPv4_ADDRESS_APACHE,$IPv6_ADDRESS_APACHE,$IPv4_ADDRESS_MASTERCONTAINER,$IPv6_ADDRESS_MASTERCONTAINER|" /usr/local/etc/php-fpm.d/www.conf
+grep listen.allowed_clients /usr/local/etc/php-fpm.d/www.conf
+
 exec "$@"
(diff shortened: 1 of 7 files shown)


Containers/nextcloud/start.sh
```diff
@@ -28,18 +28,16 @@ fi
 set -x
 IPv4_ADDRESS_APACHE="$(dig nextcloud-aio-apache A +short | grep '^[0-9.]\+$' | sort | head -n1)"
 IPv6_ADDRESS_APACHE="$(dig nextcloud-aio-apache AAAA +short | grep '^[0-9a-f:]\+$' | sort | head -n1)"
```
szaimen, line 30: Unfortunately I missed that we need for apache to start in order to get its ip-address... @Zoey2936 do you have a better idea to fix this? I dont see any unfortunately...

Containers/nextcloud/start.sh
```diff
@@ -28,18 +28,16 @@ fi
...
-IPv4_ADDRESS_NOTIFY_PUSH="$(dig nextcloud-aio-notify-push A +short | grep '^[0-9.]\+$' | sort | head -n1)"
-IPv6_ADDRESS_NOTIFY_PUSH="$(dig nextcloud-aio-notify-push AAAA +short | grep '^[0-9a-f:]\+$' | sort | head -n1)"
-IPv4_ADDRESS_DSP="$(dig nextcloud-aio-docker-socket-proxy A +short | grep '^[0-9.]\+$' | sort | head -n1)"
-IPv6_ADDRESS_DSP="$(dig nextcloud-aio-docker-socket-proxy AAAA +short | grep '^[0-9a-f:]\+$' | sort | head -n1)"
+# IPv4_ADDRESS_MASTERCONTAINER="$(dig nextcloud-aio-mastercontainer A +short | grep '^[0-9.]\+$' | sort | head -n1)"
+# IPv6_ADDRESS_MASTERCONTAINER="$(dig nextcloud-aio-mastercontainer AAAA +short | grep '^[0-9a-f:]\+$' | sort | head -n1)"
+# IPv4_ADDRESS_NOTIFY_PUSH="$(dig nextcloud-aio-notify-push A +short | grep '^[0-9.]\+$' | sort | head -n1)"
+# IPv6_ADDRESS_NOTIFY_PUSH="$(dig nextcloud-aio-notify-push AAAA +short | grep '^[0-9a-f:]\+$' | sort | head -n1)"
+# IPv4_ADDRESS_DSP="$(dig nextcloud-aio-docker-socket-proxy A +short | grep '^[0-9.]\+$' | sort | head -n1)"
+# IPv6_ADDRESS_DSP="$(dig nextcloud-aio-docker-socket-proxy AAAA +short | grep '^[0-9a-f:]\+$' | sort | head -n1)"
```
szaimen, line 36: I would like to do this but then it will log things like
```
ERROR: Connection disallowed: IP address '172.20.0.11' has been dropped.
ERROR: Connection disallowed: IP address '172.20.0.10' has been dropped.
ERROR: Connection disallowed: IP address '172.20.0.11' has been dropped.
ERROR: Connection disallowed: IP address '172.20.0.11' has been dropped.
```
so not sure how to fix this differently...

Containers/nextcloud/start.sh
Zoey2936: maybe a volume which only contains the php socket and is shared between the needed containers?

szaimen: Then we run into permission problems and things like the socket not getting correctly deleted when the process ends...

But I found a solution with a fake service

Containers/nextcloud/start.sh
```diff
@@ -25,6 +25,16 @@ if [ -f "/var/www/html/config/config.php" ]; then
     fi
 fi
 
+# TODO: need to wait for apache to spawn?
```
szaimen, line 28: todo

Zoey2936 (commented):
can't review, I have no idea

szaimen:
> can't review, I have no idea

all right

PR #3459, opened 2023-09-29
allow to set aio_variables from containers.json

szaimen:


php/src/ContainerDefinitionFetcher.php
```diff
@@ -166,9 +168,13 @@ private function GetDefinition(): array
             if (isset($entry['depends_on'])) {
                 $valueDependsOn = $entry['depends_on'];
                 if ($entry['container_name'] === 'nextcloud-aio-apache') {
+                    // Make sure nextcloud-aio-nextcloud is added last so that aio_variables are set before nextcloud-aio-nextcloud starts
+                    $valueDependsOn = array_diff($valueDependsOn, ['nextcloud-aio-nextcloud']);
```
szaimen, line 172: dont do it like this, better add ccontainers first and then later array_merge with the others

Issue #3466, opened 2023-10-02
404 Internal server error !!!


gbertolottiKS:
### Steps to reproduce
1. Use Nextcloud
2. See the server crash

### Actual behavior 
The server crashed, showing us a black screen with the writing "404 Internal server Error". It's not the first time it happens (see https://github.com/nextcloud/all-in-one/discussions/3056), but we hoped it was fixed. It seems again a problem of connection's spam.
Of course, rebooting the OS (and so Docker) solved the problem..

### Host OS 
Ubuntu 22.04

#### Nextcloud AIO version 
7.2.1

#### Current channel 
Latest
#### Other valuable info 
we attach the logs of the:
- system report
- apache container
- nextcloud container
- database container
[2023-10-02.zip](https://github.com/nextcloud/all-in-one/files/12781843/2023-10-02.zip)

szaimen:
Hi, can you for a test disable collabora from the AIO interface and check if that makes it work?

gbertolottiKS:
Hi.
Unfortunately, we can't: it's a live environment, and the colleagues uses it often. 
We can disable it during the night or in the weekend, but I fear it wouldn't be useful: during that time, Nextcloud is barely used.

szaimen:
I am wondering because there were some reports that disabling collabora makes it work. This would mean that this would need to be fixed on the richdocuments app side and not in AIO.

szaimen:
Btw, how much ram do you have and whoch optional addons are enabled in the AIO interface?

gbertolottiKS:
Hi, we took 16 GB of RAM, but usually AIO takes around 4.5 GB

![immagine](https://github.com/nextcloud/all-in-one/assets/115481064/1046b3f7-0856-4a7f-9bc0-839713e49074)

One thing that we noticed is that when the crashes occurred, there is also a spike in the memory usage: during this last crash, the memory usage was around 10 GB…

And these are the installed addons:
![immagine](https://github.com/nextcloud/all-in-one/assets/115481064/ad9ac563-119d-43ce-a292-714f3ca53c88)

So, in short, all except the Talk recording.

szaimen:
Hi, please update to AIO v7.4.0 and report back if it fixes the issue for you. Thanks!

PR #3479, opened 2023-10-03
Revert dbpersistent setting to check if it fixes too many db connections

szaimen:
For https://github.com/nextcloud/all-in-one/issues/3466 and https://github.com/nextcloud/all-in-one/discussions/3288


Zoey2936:
Maybe we should use a static count of max_connections, pgtune recommend at least 200 if you use a web application: https://pgtune.leopard.in.ua but it can be increased, even with low cpu and ram...

Zoey2936:
With 4GB (80 connections) of RAM the max_connections would be bellow the default (100):
grafik

PR #3488, opened 2023-10-05
allow _ping docker api to verify service accessible

andrey18106:
Allow `/_ping` api in docker-socket-proxy required for `app_api` to verify docker service connection.


szaimen (approved):
Fine by me

szaimen:
@andrey18106 @bigcat88 does my commit look good as well?

bigcat88:
> @andrey18106 @bigcat88 does my commit look good as well?

I like it better the way it was.
I don’t know if Docker will accept "... Privileged " - then after this commit it will pass check, when it should not.

But if Docker rejects such request with `space` at the end after `Privileged` word, then I'm fine.

bigcat88:
> @andrey18106 @bigcat88 does my commit look good as well?

never mind, I deleted my previous comment, misunderstood.
Yeah, I am fine, it looks more strict :)

szaimen:
All right, thanks! :)

PR #3492, opened 2023-10-06
optimisations of the reverse proxy and docker rootless docs

### LINKED ISSUE #3399: Optimisations of the installation description for AIO with Docker rootless and Nginx proxy manager

While installing along the given instruction of how to install Nextcloud AIO [with rootless ](https://github.com/nextcloud/all-in-one/blob/main/docker-rootless.md)and behind a [Nginx Proxy Manager](https://github.com/nextcloud/all-in-one/blob/main/reverse-proxy.md#1-add-this-to-your-reverse-proxy-config) I stumbled across different questions. This includes a few minor points before deploying AIO and one bigger issue after the deployment (see #3396) .

- If using Portainer to deploy the AIO the variable $XDG_RUNTIME_DIR is not available. In this case it is necessary to insert the path (e.g. /run/user/1000/docker.sock) directly into the [Docker compose file](https://github.com/nextcloud/all-in-one/blob/main/compose.yaml). If you are not sure how to retrieve the path, you can run on the host: `echo $XDG_RUNTIME_DIR`
- Extend the (Nginx) proxy manager config to avoid triggering brute force prevention like mentioned in #2968 AND/OR
- ...explain how to change parameter like trusted_proxies mentioned in #3396 
- Differentiate more between `localhost` and `ip-address of the host` [in this section](https://github.com/nextcloud/all-in-one/blob/main/reverse-proxy.md#nginx-proxy-manager). Example: 

> Of course you need to modify <your-nc-domain> to the domain on which you want to run Nextcloud.
> 
> **Please note**
> _Running the Proxy Manager on the host_
> The above configuration will only work if your reverse proxy is running directly on the host that is running the docker daemon. If the reverse proxy is running in a docker container, you can use the --network host when starting the reverse proxy container in order to connect the reverse proxy container to the host network. 
> 
> _Running the Proxy Manager in a Docker container_
> If that is not an option or not possible for you (like e.g. on Windows or if the reverse proxy is running on a different host), you can alternatively instead of localhost use the private ip-address of the host that is running the docker daemon. If you are not sure how to retrieve that, you can run: ip a | grep "scope global" | head -1 | awk '{print $2}' | sed 's|/.*||'. If the command returns a public ip-address, use ip a | grep "scope global" | grep docker0 | awk '{print $2}' | sed 's|/.*||' instead (the commands only work on Linux)

none

### Additional context

none

MarvvanPal:
Close #3399
Close https://github.com/nextcloud/all-in-one/issues/3502
See commits for details.


docker-rootless.md
```diff
@@ -8,7 +8,7 @@ You can run AIO with docker rootless by following the steps below.
...
-1. Use the official AIO startup command but use `--volume $XDG_RUNTIME_DIR/docker.sock:/var/run/docker.sock:ro` instead of `--volume /var/run/docker.sock:/var/run/docker.sock:ro` and also add `--env WATCHTOWER_DOCKER_SOCKET_PATH=$XDG_RUNTIME_DIR/docker.sock` to the initial container startup (which is needed for mastercontainer updates to work correctly).
+1. Use the official AIO startup command but use `--volume $XDG_RUNTIME_DIR/docker.sock:/var/run/docker.sock:ro` instead of `--volume /var/run/docker.sock:/var/run/docker.sock:ro` and also add `--env WATCHTOWER_DOCKER_SOCKET_PATH=$XDG_RUNTIME_DIR/docker.sock` to the initial container startup (which is needed for mastercontainer updates to work correctly). For some reasons (e.g. using Portainer to deploy AIO) the variable `$XDG_RUNTIME_DIR` is not available. In this case, it is necessary to manually add the path (e.g. `/run/user/1000/docker.sock`) to the Docker compose file to replace the `$XDG_RUNTIME_DIR` variable. If you are not sure how to get the path, you can run on the host: `echo $XDG_RUNTIME_DIR`.
```
szaimen, line 11:
```suggestion
1. Use the official AIO startup command but use `--volume $XDG_RUNTIME_DIR/docker.sock:/var/run/docker.sock:ro` instead of `--volume /var/run/docker.sock:/var/run/docker.sock:ro` and also add `--env WATCHTOWER_DOCKER_SOCKET_PATH=$XDG_RUNTIME_DIR/docker.sock` to the initial container startup (which is needed for mastercontainer updates to work correctly). When you are using Portainer to deploy AIO, the variable `$XDG_RUNTIME_DIR` is not available. In this case, it is necessary to manually add the path (e.g. `/run/user/1000/docker.sock`) to the Docker compose file to replace the `$XDG_RUNTIME_DIR` variable. If you are not sure how to get the path, you can run on the host: `echo $XDG_RUNTIME_DIR`.

```

reverse-proxy.md
```diff
@@ -18,6 +18,15 @@ In order to run Nextcloud behind a web server or reverse proxy (like Apache, Ngi
...
+ 
+_Running the Proxy Manager in a Docker container_<br>
+The below configuration will only work if your reverse proxy is running directly on the host that is running the docker daemon. If the reverse proxy is running in a docker container, you can use the `--network host` option (or `network_mode: host` for docker-compose) to connect the reverse proxy container to the host network. If you are using a firewall on the server, you need to open ports 80 and 443 for the reverse proxy manually.
+  
+_Running the Proxy Manager on the host_<br>
+If that is not an option or not possible for you (like e.g. on Windows or if the reverse proxy is running on a different host), you can alternatively instead of `localhost` use the private ip-address of the host that is running the docker daemon. If you are not sure how to retrieve that, you can run: `ip a | grep "scope global" | head -1 | awk '{print $2}' | sed 's|/.*||'`. If the command returns a public ip-address, use `ip a | grep "scope global" | grep docker0 | awk '{print $2}' | sed 's|/.*||'` instead (the commands only work on Linux).
```
szaimen, line 28:
```suggestion
**Running the Reverse Proxy on the same server, not in a container**
For this setup, the default sample configurations should work.

**Running the Reverse Proxy in a Docker container on the same server**
For this setup, you can use as target `host.docker.internal:$APACHE_PORT` instead of `localhost:$APACHE_PORT`.<br>
Another option and actually the recommended way in this case is to use `--network host` option (or `network_mode: host` for docker-compose) for the reverse proxy container to connect it to the host network. If you are using a firewall on the server, you need to open ports 80 and 443 for the reverse proxy manually. By doing so, the default sample configurations should work.

**Running the Reverse Proxy on a different server (no matter if in container or not)**
For this setup, you need to use as target the private ip-address of the host that shall be running AIO. So e.g. `private.ip.address.of.aio.host:$APACHE_PORT` instead of `localhost:$APACHE_PORT`.<br>
If you are not sure how to retrieve that, you can run: `ip a | grep "scope global" | head -1 | awk '{print $2}' | sed 's|/.*||'` on the server that shall be running AIO. If the command returns a public ip-address, use `ip a | grep "scope global" | grep docker0 | awk '{print $2}' | sed 's|/.*||'` instead (the commands only work on Linux).
```

szaimen (commented):
Thanks a lot for the attempt! I made some suggestions and commit them :)

Zoey2936 (approved):

hunhejj (approved):
Looks good to me!

szaimen:
Thank you both for the approvals and @MarvvanPal for the PR! :)

PR #3520, opened 2023-10-09
Add NPMplus as community-container

Zoey2936:


community-containers/npmplus/npmplus.json
```diff
@@ -0,0 +1,30 @@
...
+            "restart": "unless-stopped",
+            "environment": [
+                "TZ=%TIMEZONE%",
+                "NC_AIO=true",
+                "NC_DOMAIN=%NC_DOMAIN%",
+                "NPM_LISTEN_LOCALHOST=true"
+            ],
+            "volumes": [
+                {
+                    "source": "nextcloud_aio_npmplus",
```
szaimen, line 19: Does this have some data that should be backed up? 

If so it can be covered like this https://github.com/nextcloud/all-in-one/blob/5dcdb6268ab2b08b58f7f2510de3bfe01bc55bbc/community-containers/plex/plex.json#L36-L38

It will also get restored on a new instance then automatically :)

community-containers/npmplus/readme.md
```diff
@@ -0,0 +1,17 @@
+## NPMplus
+This container contains a fork of the Nginx Proxy Manager, which is a WebUI for nginx. It will also automatically create a config and cert for AIO.
+
+### Notes
+- Make sure that no other service is using port `80 (tcp)`, `443 (tcp/upd)` or `81 (udp)` on your host as otherwise the containers will fail to start. You can check this with `sudo netstat -tulpn | grep "443\|80\|81"` before installing AIO.
```
szaimen, line 5: I guess this will conflict with the default AIO instructions... Maybe there should be an additional hint that the command for AIO itself needs to be adjusted? I suppose there is no way for NPM to retreive certs via the tls-alpn challenge? This would eliminate this problem :)

community-containers/npmplus/readme.md
```diff
@@ -0,0 +1,17 @@
...
+- After the container was started the first time, please check the logs for errors. Then you can open NPMplus on `https://127.0.0.1:81` (`[::1]` also works) and change the password.
```
szaimen, line 6: Maybe also mention that https://ip.address.of.server:81 should work as well or doesnt it?

community-containers/npmplus/readme.md
```diff
@@ -0,0 +1,17 @@
+## NPMplus
+This container contains a fork of the Nginx Proxy Manager, which is a WebUI for nginx. It will also automatically create a config and cert for AIO.
+
+### Notes
+- Make sure that no other service is using port `80 (tcp)`, `443 (tcp/upd)` or `81 (udp)` on your host as otherwise the containers will fail to start. You can check this with `sudo netstat -tulpn | grep "443\|80\|81"` before installing AIO.
+- After the container was started the first time, please check the logs for errors. Then you can open NPMplus on `https://127.0.0.1:81` (`[::1]` also works) and change the password. 
+- If you want to use NPMplus behind a domain and outside localhost just create a new proxy host inside the NPMplus which proxies to `https`, `127.0.0.1` and port `81` - all other settings should be the same as for the AIO host.
+- The default password is `iArhP1j7p1P6TA92FA2FMbbUGYqwcYzxC4AVEe12Wbi94FY9gNN62aKyF1shrvG4NycjjX9KfmDQiwkLZH1ZDR9xMjiG2QmoHXi` and the default email is `admin@example.com`
+- Please change the default login data first, after you can read inside the logs that the default config for AIO is created and there are no errors.
```
szaimen, line 9: do the logs get printed to the container logs as well? Should make debugging a bit easier...

community-containers/npmplus/readme.md
```diff
@@ -0,0 +1,17 @@
+## NPMplus
+This container contains a fork of the Nginx Proxy Manager, which is a WebUI for nginx. It will also automatically create a config and cert for AIO.
```
szaimen, line 2: I guess we should add a hint also to the caddy community container that you cannot use both at the same time?

szaimen (commented):
Thanks a lot for this contribution! Looks already pretty good! :) 

Just a few small comments...

community-containers/npmplus/readme.md
Zoey2936: does not because of NPM_LISTEN_LOCALHOST=true

community-containers/npmplus/readme.md
Zoey2936: yes

community-containers/npmplus/readme.md
Zoey2936: thats why I asked https://github.com/nextcloud/all-in-one/discussions/3505, I used the caddy community container as base to create this community container and there is also no mention of this

community-containers/npmplus/npmplus.json
Zoey2936: yes should be inside the backup

szaimen: all right :)

community-containers/npmplus/readme.md
szaimen: ah I see. Maybe there should be a hint that this will not work then? Also what about situations where you cannot open a browser on the server directly? How the NPM interface then?

community-containers/npmplus/readme.md
szaimen: cool!

community-containers/npmplus/readme.md
Zoey2936: port 80 of mastercontainer always redirects to https, right?

community-containers/npmplus/readme.md
Zoey2936: that is true

community-containers/npmplus/readme.md
szaimen: yes but it is also used for getting the cert for the AIO interface... So if port 80 is removed from there then only port 8080 can be used...

community-containers/npmplus/readme.md
szaimen: all right? So now access via ip.address.of.server.81 should work right?

community-containers/npmplus/readme.md
```diff
@@ -2,13 +2,15 @@
...
 - After the container was started the first time, please check the logs for errors. Then you can open NPMplus on `https://127.0.0.1:81` (`[::1]` also works) and change the password. 
 - If you want to use NPMplus behind a domain and outside localhost just create a new proxy host inside the NPMplus which proxies to `https`, `127.0.0.1` and port `81` - all other settings should be the same as for the AIO host.
 - The default password is `iArhP1j7p1P6TA92FA2FMbbUGYqwcYzxC4AVEe12Wbi94FY9gNN62aKyF1shrvG4NycjjX9KfmDQiwkLZH1ZDR9xMjiG2QmoHXi` and the default email is `admin@example.com`
 - Please change the default login data first, after you can read inside the logs that the default config for AIO is created and there are no errors.
 - The PHP options are not supported, as well as all other env options.
 - See https://github.com/nextcloud/all-in-one/tree/main/community-containers how to add it to the AIO stack
+- Incompatible with the caddy community container
```
szaimen, line 14: Lets move this up to the very top please

szaimen: and maybe add a link to it? something like https://github.com/nextcloud/all-in-one/tree/main/community-containers/caddy

community-containers/npmplus/readme.md
Zoey2936: then I just add DISABLE_HTTP=true and it should work, the users are then forced to use https for all sites

community-containers/npmplus/readme.md
Zoey2936: you can make this change

community-containers/npmplus/readme.md
Zoey2936: yes

community-containers/npmplus/readme.md
szaimen: So NPM can actually get certificates via tls-alpn challenge? :)

community-containers/npmplus/readme.md
szaimen: should we document this?

community-containers/npmplus/readme.md
Zoey2936: no, but the http challenge support https, if http redirects to the same domain on https, it ignores self signed or invalid certs on https

szaimen: interesting, I didnt know! So I guess this should actually make it work then? :)
https://github.com/nextcloud/all-in-one/blob/5dcdb6268ab2b08b58f7f2510de3bfe01bc55bbc/Containers/mastercontainer/Caddyfile#L23-L25

community-containers/npmplus/readme.md
```diff
@@ -0,0 +1,20 @@
+## NPMplus
+This container contains a fork of the Nginx Proxy Manager, which is a WebUI for nginx. It will also automatically create a config and cert for AIO.
+
+### Notes
+- This container is incompatible with the [caddy](https://github.com/nextcloud/all-in-one/tree/main/community-containers/caddy) community container. So make sure that you do not enable both at the same time!
+- Only works on linux since it uses network mode host
+- You can ignore the NPM configuration of the reverse-proxy.md. The NPMplus fork already contains the changes of the advanced tab.
```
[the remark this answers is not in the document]
Zoey2936: I would say yes, since the config for NPMplus is diffrent from NPM

community-containers/npmplus/readme.md
```diff
@@ -0,0 +1,21 @@
...
+- Please change the default login data first, after you can read inside the logs that the default config for AIO is created and there are no errors.
+- After the container was started the first time, please check the logs for errors. Then you can open NPMplus on `https://<ip>:81` and change the password. 
+- The default password is `iArhP1j7p1P6TA92FA2FMbbUGYqwcYzxC4AVEe12Wbi94FY9gNN62aKyF1shrvG4NycjjX9KfmDQiwkLZH1ZDR9xMjiG2QmoHXi` and the default email is `admin@example.com`
+- If you want to use NPMplus behind a domain and outside localhost just create a new proxy host inside the NPMplus which proxies to `https`, `127.0.0.1` and port `81` - all other settings should be the same as for the AIO host.
+- If you want to set env options from this [compose.yaml](https://github.com/ZoeyVid/NPMplus/blob/develop/compose.yaml), please set them inside the `.env` file which you can find in the `nextcloud_aio_npmplus` volume
+- See https://github.com/nextcloud/all-in-one/tree/main/community-containers how to add it to the AIO stack
+- **Important:** you always need to enable https for your hosts, since `DISABLE_HTTP` is set to true
```
szaimen, line 15:
```suggestion
- **Important:** you always need to enable https for your hosts, since `DISABLE_HTTP` is set to true
- See https://github.com/nextcloud/all-in-one/tree/main/community-containers how to add it to the AIO stack
```

community-containers/npmplus/readme.md
Zoey2936: yes

Zoey2936: https://github.com/ZoeyVid/NPMplus/blob/920e4848fa731563dec9535c8cbbb93626af57ad/Caddyfile#L28 I know with this line it works

community-containers/npmplus/readme.md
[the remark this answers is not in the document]
szaimen: all right

community-containers/npmplus/readme.md
szaimen: So maybe permanend should be added?

szaimen: https://github.com/nextcloud/all-in-one/pull/3522

szaimen (approved):
LGTM but we need to wait for the next stable release

szaimen:
This is now released with v7.5.0 Beta. Testing and feedback is welcome! See https://github.com/nextcloud/all-in-one#how-to-switch-the-channel

PR #3550, opened 2023-10-17
Add more compose presets and fix typos

p-fruck:
During my setup of nextcloud AIO behind a reverse proxy I noticed some typos I'd like to fix. Also, I would like to add configuration examples for using a volume mount and running on SELinux to the compose.yaml

### DIFF
--- a/compose.yaml
+++ b/compose.yaml
@@ -32,6 +32,8 @@ services:
       # - WATCHTOWER_DOCKER_SOCKET_PATH=/var/run/docker.sock # Needs to be specified if the docker socket on the host is not located in the default '/var/run/docker.sock'. Otherwise mastercontainer updates will fail. For macos it needs to be '/var/run/docker.sock'
     # networks: # Is needed when you want to create the nextcloud-aio network with ipv6-support using this file, see the network config at the bottom of the file
       # - nextcloud-aio # Is needed when you want to create the nextcloud-aio network with ipv6-support using this file, see the network config at the bottom of the file
+    # # Uncomment the following line when using SELinux
+    # security_opt: ["label:disable"]
 
   # # Optional: Caddy reverse proxy. See https://github.com/nextcloud/all-in-one/blob/main/reverse-proxy.md
   # # You can find further examples here: https://github.com/nextcloud/all-in-one/discussions/588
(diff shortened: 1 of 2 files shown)


compose.yaml
```diff
@@ -4,6 +4,8 @@ services:
     init: true
     restart: always
     container_name: nextcloud-aio-mastercontainer # This line is not allowed to be changed as otherwise AIO will not work correctly
+    # # Uncomment the following line when using SELinux
+    # security_opt: ["label:disable"]
```
szaimen, line 8: Please add this below the environment variables

compose.yaml
```diff
@@ -50,6 +52,12 @@ services:
 volumes:
   nextcloud_aio_mastercontainer:
     name: nextcloud_aio_mastercontainer # This line is not allowed to be changed as otherwise the built-in backup solution will not work
+    # # Optional: Use a volume mount instead of internal docker volume
+    # driver: local
+    # driver_opts:
+    #   type: none
+    #   o: bind
+    #   device: ./nextcloud_data
```
szaimen, line 60: I am not sure if we should expose this. People might think that all data will be stored in that location which is not the case. I would rather point at https://github.com/nextcloud/all-in-one#how-to-store-the-filesinstallation-on-a-separate-drive

szaimen (commented):
Thanks for the PR! Just a few things to discuss. See below

compose.yaml
p-fruck: I think those are two different topics. I personally just dislike docker volumes (because I tend to prune my whole docker system from time to time and don't want to risk loosing my data) and would prefer to mount all persistent data of the nextcloud installation in a local folder so that I can always access those files using the standard file system, just in case. This is not about changing the storage directory of docker. The datafolder and the compose file should be everything required to reproduce the current installation on a new server. Isn't this the case for the AIO installation?

szaimen: > The datafolder and the compose file should be everything required to reproduce the current installation on a new server. Isn't this the case for the AIO installation?

It isnt. See https://github.com/nextcloud/all-in-one#how-does-it-work. Also see https://github.com/nextcloud/all-in-one#backup-solution

szaimen (approved):
LGTM, thanks! :)

Issue #3566, opened 2023-10-19
Docker Socket Proxy fails to start.


ShapeShifter499:
### Steps to reproduce
Install the docker, see log. 

### Expected behavior 
Docker socket proxy should load and work. 

### Actual behavior 
AIO settings webpage states "starting" but it never appears to actually start. I see errors in the log for nextcloud-aio-docker-socket-proxy, log attached at the bottom.

### Host OS 
Manjaro x86_64

#### Current channel 
Nextcloud AIO v7.4.1

#### Other valuable info 
[nextcloud-aio-docker-socket-proxy_log.txt](https://github.com/nextcloud/all-in-one/files/13045687/nextcloud-aio-docker-socket-proxy_log.txt)

szaimen:
This is the issue apparently:
```
[ALERT]    (23) : Not enough memory to allocate 1073741815 entries for fdtab!
[ALERT]    (23) : No polling mechanism available.
  This may happen when using thread-groups with old pollers (poll/select), or
  it is possible that haproxy was built with TARGET=generic and that FD_SETSIZE
  is too low on this platform to support maxconn and the number of listeners
  and servers. You should rebuild haproxy specifying your system using TARGET=
  in order to support other polling systems (poll, epoll, kqueue) or reduce the
  global maxconn setting to accommodate the system's limitation. For reference,
  FD_SETSIZE=1024 on this system, global.maxconn=536870896 resulting in a maximum of
  1073741815 file descriptors. You should thus reduce global.maxconn by 536870396. Also,
  check build settings using 'haproxy -vv'.
```

szaimen:
https://github.com/haproxy/haproxy/issues/2043

szaimen:
Possible fix: https://github.com/matrix-org/sytest/commit/a42e75a3845f339ad41aac9b3cf5fef820c0f7fc

ShapeShifter499:
> Possible fix: [matrix-org/sytest@a42e75a](https://github.com/matrix-org/sytest/commit/a42e75a3845f339ad41aac9b3cf5fef820c0f7fc)

so this would be added to the haproxy used inside the AIO?

szaimen:
> > Possible fix: [matrix-org/sytest@a42e75a](https://github.com/matrix-org/sytest/commit/a42e75a3845f339ad41aac9b3cf5fef820c0f7fc)
> 
> so this would be added to the haproxy used inside the AIO?

yes

sol8712:
[...]

PR #3567, opened 2023-10-19
add maxconn value to haproxy

szaimen:
Fix https://github.com/nextcloud/all-in-one/issues/3566

### DIFF
--- a/Containers/docker-socket-proxy/haproxy.cfg
+++ b/Containers/docker-socket-proxy/haproxy.cfg
@@ -1,5 +1,8 @@
 # Inspiration: https://github.com/Tecnativa/docker-socket-proxy/blob/master/haproxy.cfg
 
+global
+    maxconn 10
+
 defaults
     timeout connect 10s
     timeout client 10s


Containers/docker-socket-proxy/haproxy.cfg
```diff
@@ -1,5 +1,8 @@
 # Inspiration: https://github.com/Tecnativa/docker-socket-proxy/blob/master/haproxy.cfg
 
+global
+    maxconn 2000
```
Zoey2936, line 4: is there are reason for using this number?

szaimen: No

Zoey2936: do you know how many connection Appapi uses?

szaimen: @andrey18106 @bigcat88 do you know?

andrey18106: Is it that necessary? AppAPI uses connection to daemon only during administration actions

szaimen: Apparently maxconn is needed on some system. See https://github.com/nextcloud/all-in-one/issues/3566

andrey18106: It would be enough with reserve set it to 10 I guess.

szaimen: all right. adjusted.

andrey18106 (approved):

PR #3583, opened 2023-10-21
do not catch error during pullcontainer so that it throws and logs if it cannot get the image

szaimen:
Make things like https://github.com/nextcloud/all-in-one/issues/3582 easier to debug

### DIFF
--- a/php/src/Docker/DockerActionManager.php
+++ b/php/src/Docker/DockerActionManager.php
@@ -577,14 +577,8 @@ class DockerActionManager
     public function PullContainer(Container $container) : void
     {
         $url = $this->BuildApiUrl(sprintf('images/create?fromImage=%s', urlencode($this->BuildImageName($container))));
-        try {
-            $this->guzzleClient->post($url);
-        } catch (RequestException $e) {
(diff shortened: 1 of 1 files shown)


php/src/Docker/DockerActionManager.php
```diff
@@ -579,7 +579,8 @@ public function PullContainer(Container $container) : void
         $url = $this->BuildApiUrl(sprintf('images/create?fromImage=%s', urlencode($this->BuildImageName($container))));
         try {
             $this->guzzleClient->post($url);
-        } catch (RequestException $e) {
+        // catching throwable to make sure that something is logged here in any case
+        } catch (\Throwable $e) {
             error_log('Could not get image ' . $this->BuildImageName($container) . ' from docker hub. Probably due to rate limits. ' . $e->getMessage());
```
szaimen, line 584:
```suggestion
            error_log('Could not get image ' . $this->BuildImageName($container) . ' from docker hub: ' . $e->getMessage());
```

szaimen: Or maybe just throw here ans not catch the error?

PR #3594, opened 2023-10-23
fix reference to column in wikipedia timezone list

Habbie:
https://github.com/nextcloud/all-in-one/blob/main/Containers/postgresql/start.sh has the same old text. I've used the GitHub web editor now so I'm not sure I can add that edit, but if preferred, I can `git clone` and fix that one too.

### DIFF
--- a/Containers/postgresql/start.sh
+++ b/Containers/postgresql/start.sh
@@ -31,7 +31,7 @@ fi
 if [ -f "$DUMP_DIR/initialization.failed" ]; then
     echo "The database initialization failed. Most likely was a wrong timezone selected."
     echo "The selected timezone is '$TZ'." 
-    echo "Please check if it is in 'TZ database name' column of the timezone list: https://en.wikipedia.org/wiki/List_of_tz_database_time_zones#List"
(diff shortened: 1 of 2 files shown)


Containers/postgresql/start.sh
```diff
@@ -31,7 +31,7 @@ fi
 if [ -f "$DUMP_DIR/initialization.failed" ]; then
     echo "The database initialization failed. Most likely was a wrong timezone selected."
     echo "The selected timezone is '$TZ'." 
-    echo "Please check if it is in 'TZ database name' column of the timezone list: https://en.wikipedia.org/wiki/List_of_tz_database_time_zones#List"
+    echo "Please check if it is in 'TZ identfier' column of the timezone list: https://en.wikipedia.org/wiki/List_of_tz_database_time_zones#List"
```
Habbie, line 34:
```suggestion
    echo "Please check if it is in 'TZ identifier' column of the timezone list: https://en.wikipedia.org/wiki/List_of_tz_database_time_zones#List"
```

szaimen: thanks!

szaimen:
This is now released with v7.6.0 Beta. Testing and feedback is welcome! See https://github.com/nextcloud/all-in-one#how-to-switch-the-channel

PR #3614, opened 2023-10-26
Nextcloud-DLNA community container

thanek:
Discussion is here: https://github.com/thanek/nextcloud-dlna/issues/1

Cannot be merged before https://github.com/nextcloud/all-in-one/pull/3611

- needs https://github.com/nextcloud/all-in-one/pull/3611

### DIFF
--- a/community-containers/dlna/dlna.json
+++ b/community-containers/dlna/dlna.json
@@ -12,6 +12,8 @@
                 "nextcloud-aio-database"
             ],
             "environment": [
+                "NC_DOMAIN=%NC_DOMAIN%",
+                "NC_PORT=443",
                 "NEXTCLOUD_DLNA_SERVER_PORT=9999",
                 "NEXTCLOUD_DLNA_FRIENDLY_NAME=nextcloud-aio",
                 "NEXTCLOUD_DATA_DIR=/data",
--- a/community-containers/dlna/dlna.json
+++ b/community-containers/dlna/dlna.json
@@ -1,0 +1,38 @@
+{
+    "aio_services_v1": [
+        {
+            "container_name": "nextcloud-aio-dlna",
+            "display_name": "DLNA",
+            "documentation": "https://github.com/nextcloud/all-in-one/tree/main/community-containers/dlna",
+            "image": "thanek/nextcloud-dlna",
+            "image_tag": "latest",
+            "internal_port": "host",
+            "restart": "unless-stopped",
+            "depends_on": [
+                "nextcloud-aio-database",
+                "nextcloud-aio-nextcloud"
+            ],
+            "environment": [
+                "NEXTCLOUD_DLNA_SERVER_PORT=9999",
+                "NEXTCLOUD_DLNA_FRIENDLY_NAME=Nextcloud-AIO",
+                "NEXTCLOUD_DATA_DIR=/data",
+                "NEXTCLOUD_DB_TYPE=postgres",
+                "NEXTCLOUD_DB_HOST=%AIO_DATABASE_HOST%",
+                "NEXTCLOUD_DB_PORT=5432",
+                "NEXTCLOUD_DB_NAME=nextcloud_database",
+                "NEXTCLOUD_DB_USER=nextcloud",
+                "NEXTCLOUD_DB_PASS=%DATABASE_PASSWORD%"
+            ],
+            "secrets": [
+                "DATABASE_PASSWORD"
+            ],
+            "volumes": [
+                {
+                    "source": "%NEXTCLOUD_DATADIR%",
+                    "destination": "/data",
+                    "writeable": false
+                }
+            ]
+        }
+    ]
+}
--- a/community-containers/dlna/dlna.json
+++ b/community-containers/dlna/dlna.json
@@ -9,18 +9,17 @@
             "internal_port": "host",
             "restart": "unless-stopped",
             "depends_on": [
-                "nextcloud-aio-database",
-                "nextcloud-aio-nextcloud"
+                "nextcloud-aio-database"
             ],
             "environment": [
                 "NEXTCLOUD_DLNA_SERVER_PORT=9999",
-                "NEXTCLOUD_DLNA_FRIENDLY_NAME=Nextcloud-AIO",
+                "NEXTCLOUD_DLNA_FRIENDLY_NAME=nextcloud-aio",
                 "NEXTCLOUD_DATA_DIR=/data",
                 "NEXTCLOUD_DB_TYPE=postgres",
                 "NEXTCLOUD_DB_HOST=%AIO_DATABASE_HOST%",
                 "NEXTCLOUD_DB_PORT=5432",
                 "NEXTCLOUD_DB_NAME=nextcloud_database",
-                "NEXTCLOUD_DB_USER=nextcloud",
+                "NEXTCLOUD_DB_USER=oc_nextcloud",
                 "NEXTCLOUD_DB_PASS=%DATABASE_PASSWORD%"
             ],
             "secrets": [
--- a/community-containers/dlna/readme.md
+++ b/community-containers/dlna/readme.md
@@ -2,7 +2,7 @@
 This container bundles DLNA server for your Nextcloud files to be accessible by the clients in your local network. Simply run the container and look for a new media server `nextcloud-aio` in your local network.
 
 ### Notes
-- This is not working on Mac OS since it requires the `host` networking mode in docker, and it doesn't really share the host's network interfaces in this system
+- This is not working with Docker Desktop since it requires the `host` networking mode in docker, and it doesn't really share the host's network interfaces in this system
 - If you have a firewall like ufw configured, you might need to open at least port 9999 TCP and 1900 UDP first in order to make it work.
 
 ### Repository
--- a/community-containers/dlna/readme.md
+++ b/community-containers/dlna/readme.md
@@ -5,7 +5,7 @@ This container bundles DLNA server for your Nextcloud files to be accessible by 
 - This container needs the `nextcloud-aio-nextcloud` container to be up and running to work
 - This is not working on Docker Desktop since it needs `network_mode: host` in order to work correctly
 - This is not working on Mac OS since the `host` networking mode doesn't really share the host's network interfaces in this system
-- It is needed to allow TCP network traffic on port 9999 to the AIO instance and the UDP (port 1900) from it to the local network
+If you have a firewall like ufw configured, you might need to open at least port 9999 TCP Port 1900 UDP first in order to make it work.
 
 ### Repository
 https://github.com/thanek/nextcloud-dlna
--- a/community-containers/dlna/readme.md
+++ b/community-containers/dlna/readme.md
@@ -2,6 +2,7 @@
 This container bundles DLNA server for your Nextcloud files to be accessible by the clients in your local network. Simply run the container and look for a new media server `nextcloud-aio` in your local network.
 
 ### Notes
+- This container will work only if the Nextcloud installation is in your home network, it is not suitable for installations on remote servers.
 - This is not working with Docker Desktop since it requires the `host` networking mode in docker, and it doesn't really share the host's network interfaces in this system
 - If you have a firewall like ufw configured, you might need to open at least port 9999 TCP and 1900 UDP first in order to make it work.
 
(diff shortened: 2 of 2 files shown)


community-containers/dlna/dlna.json
```diff
@@ -0,0 +1,38 @@
...
+            "container_name": "nextcloud-aio-dlna",
+            "display_name": "DLNA",
+            "documentation": "https://github.com/nextcloud/all-in-one/tree/main/community-containers/dlna",
+            "image": "thanek/nextcloud-dlna",
+            "image_tag": "latest",
+            "internal_port": "host",
+            "restart": "unless-stopped",
+            "depends_on": [
+                "nextcloud-aio-database",
+                "nextcloud-aio-nextcloud"
```
szaimen, line 13: This is not going to work as you might think. It only means that the container will be started after the nextcloud container. However the nextcloud container might still not be ready then since it will first start the installation on the first startup if not already done. So instead you should probably add a logic inside your container that first checks maybe in t he database if nextcloud is actually installed.

community-containers/dlna/dlna.json
```diff
@@ -0,0 +1,38 @@
...
+            ],
+            "environment": [
+                "NEXTCLOUD_DLNA_SERVER_PORT=9999",
+                "NEXTCLOUD_DLNA_FRIENDLY_NAME=Nextcloud-AIO",
```
szaimen, line 17: Maybe?
```suggestion
                "NEXTCLOUD_DLNA_FRIENDLY_NAME=nextcloud-aio",
```

community-containers/dlna/dlna.json
```diff
@@ -0,0 +1,38 @@
...
+                "NEXTCLOUD_DATA_DIR=/data",
+                "NEXTCLOUD_DB_TYPE=postgres",
+                "NEXTCLOUD_DB_HOST=%AIO_DATABASE_HOST%",
+                "NEXTCLOUD_DB_PORT=5432",
+                "NEXTCLOUD_DB_NAME=nextcloud_database",
+                "NEXTCLOUD_DB_USER=nextcloud",
```
szaimen, line 23:
```suggestion
                "NEXTCLOUD_DB_USER=oc_nextcloud",
```

community-containers/dlna/readme.md
```diff
@@ -0,0 +1,9 @@
+## DLNA server
+This container bundles DLNA server for your Nextcloud files to be accessible by the clients in your local network. Simply run the container and look for a new media server Nextcloud-AIO in your local network.
```
szaimen, line 2: Possibly?
```suggestion
This container bundles DLNA server for your Nextcloud files to be accessible by the clients in your local network. Simply run the container and look for a new media server `nextcloud-aio` in your local network.
```

community-containers/dlna/readme.md
```diff
@@ -0,0 +1,9 @@
...
+
```
szaimen, line 3: You will need to add some hints here, some ideas what should be written here are in https://github.com/nextcloud/all-in-one/tree/main/community-containers/plex#notes

szaimen (commented):
Thanks for the PR! :)
Some feedback below

community-containers/dlna/readme.md
thanek: OK, how about now? I've also added a note about the existing nextcloud container requirement.

community-containers/dlna/dlna.json
thanek: To be honest, I have no idea how can I check if the nextcloud installation is present or is in the preparation phase, or how the dlna container  should behave when it faces this. So I removed the dependency here and added an entry to the Notes section in the readme of nextcloud-dlna community container.

community-containers/dlna/dlna.json
thanek: Thanks!

community-containers/dlna/readme.md
```diff
@@ -1,5 +1,11 @@
 ## DLNA server
-This container bundles DLNA server for your Nextcloud files to be accessible by the clients in your local network. Simply run the container and look for a new media server Nextcloud-AIO in your local network.
+This container bundles DLNA server for your Nextcloud files to be accessible by the clients in your local network. Simply run the container and look for a new media server `nextcloud-aio` in your local network.
+
+### Notes
+- This container needs the `nextcloud-aio-nextcloud` container to be up and running to work
```
szaimen, line 5: I would remove this notice. As I said in best case there should be a check maybe in the startup command of the container that checks if for example the given nextcloud domain reachable and only start up the service in that case?

community-containers/dlna/readme.md
```diff
@@ -1,5 +1,11 @@
...
+- This is not working on Docker Desktop since it needs `network_mode: host` in order to work correctly
+- This is not working on Mac OS since the `host` networking mode doesn't really share the host's network interfaces in this system
```
szaimen, line 7: I think Docker Desktop includes macOS?

thanek: Those two lines probably mean the same - the `host` mode on Mac OS is unusable for DLNA to work. I'll change that.

community-containers/dlna/readme.md
```diff
@@ -1,5 +1,11 @@
...
+- It is needed to allow TCP network traffic on port 9999 to the AIO instance and the UDP (port 1900) from it to the local network
```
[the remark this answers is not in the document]
szaimen: BTW, now that I am reading this, don't you think using bridged networking and exposing 9999/tcp and 1900/udp should make it work?

community-containers/dlna/dlna.json
szaimen: See https://github.com/nextcloud/all-in-one/pull/3614#discussion_r1376419690

community-containers/dlna/readme.md
szaimen: You could use netcat-openbsd like this for example:
```bash
while ! nc -z "$NC_DOMAIN" 443; do
	echo "Waiting for Nextcloud to become online"
	sleep 5
done
```

szaimen: In that case you would need to add a wrapper bash script as entrypoint and start the dlna server from there.

szaimen: If a domain is given, it could run this check and wait for it to start. If not, it should skip this waiting...

community-containers/dlna/readme.md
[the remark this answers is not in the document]
thanek: Not sure, I'm not an expert in docker network settings. I've just glanced at plex and miniDLNA docker containers as an inspiration ;)
From what I know the requirements are:
* container must be able to make multicast requests to the local network on UDP, port 1900
* container must be able to receive the UDP requests from other machines in the local network
* in the above messages, DLNA messages contain the `PresentationURL` which should be a local network accessible address for TCP (HTTP) requests such as 192.168.0.10:9999.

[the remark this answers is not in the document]
szaimen: Have you already tried using
`"internal_port": "9999",`
and then adding 
```
            "ports": [
                {
                  "ip_binding": "",
                  "port_number": "9999",
                  "protocol": "tcp"
                },
                {
                    "ip_binding": "",
                    "port_number": "1900",
                    "protocol": "udp"
                }
            ],
```

community-containers/dlna/readme.md
thanek: But I would have to put this script into the nextcloud-dlna container, right? I would rather not do that in the nextcloud-dlna container for two reasons:
* nextcloud-dlna doesn't need an access to the Nextcloud APP via HTTP, so checking it and waiting for it in most cases would make no sense (it only needs an access to the DB and `data` folder)
* Nextcloud is not always exposed on 443 port, there is a chance that it's a backend installation using port 80 (or other).

szaimen: I've created a PR: https://github.com/thanek/nextcloud-dlna/pull/17. Let me know what you think :)

thanek: Nice one! I'll take a look at it and let you know, looks pretty good at first glance :)

community-containers/dlna/readme.md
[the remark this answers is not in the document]
thanek: I will check this out once I start my Linux box again :)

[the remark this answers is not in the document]
thanek: OK, I've  tested the `bridge` net mode and standard container net mode and they both failed. Even if the bridged network seems to use the same network interface as the host, since the container has a different address, the `PresentationURL` is unusable for other devices in the local network to reach the DLNA server with it. The container mode not only has a different address issue (just like `bridge`) but also it doesn't share the network interface so the container is unable to listen for responses from the local network devices at all. So, I guess the `host` option is the only way to go. Maybe it would be possible to create a network that would be accessible for both containers and other local network devices, but it doesn't seem to be simple (at least for me) and I suppose It would be easy to implement it in the AIO as a generic solution for every network configuration. But, as I said, I'm not an expert and maybe there is some better solution - I just don't know it.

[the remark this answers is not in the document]
szaimen: > . So, I guess the `host` option is the only way to go.

All right. Thanks for testing!

community-containers/dlna/readme.md
```diff
@@ -0,0 +1,13 @@
+## DLNA server
+This container bundles DLNA server for your Nextcloud files to be accessible by the clients in your local network. Simply run the container and look for a new media server `nextcloud-aio` in your local network.
+
+### Notes
+- This is not working on Mac OS since it requires the `host` networking mode in docker, and it doesn't really share the host's network interfaces in this system
+- If you have a firewall like ufw configured, you might need to open at least port 9999 TCP and 1900 UDP first in order to make it work.
+
```
szaimen, line 7: Should prpbably also add a note here that this is only for home networks and not for a vps for example...

thanek: Yeah, you're right. I've added another note for that :)

community-containers/dlna/readme.md
thanek: Great. I've just released version 1.1.2 of nextcloud-dlna - it contains your wrapper and is aware of nextcloud presentce. Thanks!

community-containers/dlna/dlna.json
```diff
@@ -0,0 +1,37 @@
...
+            "documentation": "https://github.com/nextcloud/all-in-one/tree/main/community-containers/dlna",
+            "image": "thanek/nextcloud-dlna",
+            "image_tag": "latest",
+            "internal_port": "host",
+            "restart": "unless-stopped",
+            "depends_on": [
+                "nextcloud-aio-database"
+            ],
+            "environment": [
+                "NEXTCLOUD_DLNA_SERVER_PORT=9999",
```
szaimen, line 15:
```suggestion
                "NC_DOMAIN=%NC_DOMAIN%",
                "NC_PORT=443",
                "NEXTCLOUD_DLNA_SERVER_PORT=9999",
```

community-containers/dlna/readme.md
szaimen: nice :)

community-containers/dlna/readme.md
```diff
@@ -0,0 +1,13 @@
+## DLNA server
+This container bundles DLNA server for your Nextcloud files to be accessible by the clients in your local network. Simply run the container and look for a new media server `nextcloud-aio` in your local network.
+
+### Notes
+- This is not working on Mac OS since it requires the `host` networking mode in docker, and it doesn't really share the host's network interfaces in this system
```
szaimen, line 5:
```suggestion
- This is not working with Docker Desktop since it requires the `host` networking mode in docker, and it doesn't really share the host's network interfaces in this system
```

community-containers/dlna/readme.md
szaimen: I cannot see that note yet...

Also would be good if you would keep a note like this:
`See https://github.com/nextcloud/all-in-one/tree/main/community-containers#community-containers how to add it to the AIO stack`

thanek: haha, that's because I forgot to push it :) there it goes. OK, I'll add a note you provided.

szaimen: Nice!

szaimen (approved):
LGTM, thanks! :)

szaimen:
This is now released with v7.6.0 Beta. Testing and feedback is welcome! See https://github.com/nextcloud/all-in-one#how-to-switch-the-channel

PR #3631, opened 2023-10-29
Add a config option for the Postgres database port

davidmehren:
Tested using manual deployment. 
I couldn't verify that this doesn't break something with the mastercontainer, but I don't see why it should.

Should the new config option be documented somewhere?

Addresses https://github.com/nextcloud/all-in-one/discussions/3097

### DIFF
--- a/Containers/nextcloud/entrypoint.sh
+++ b/Containers/nextcloud/entrypoint.sh
@@ -227,7 +227,11 @@ if ! [ -f "$NEXTCLOUD_DATA_DIR/skip.update" ]; then
 DATADIR_PERMISSION_CONF
 
             echo "Installing with PostgreSQL database"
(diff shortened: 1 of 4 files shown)


szaimen:
Hi, thanks for the PR!

I still find occurences of 5432 in this container: https://github.com/nextcloud/all-in-one/tree/main/Containers/postgresql

Also please add it as value to the following config: https://github.com/nextcloud/all-in-one/blob/main/php/containers.json

As an example: https://github.com/nextcloud/all-in-one/blob/f4ee3ff6b0281ba02fd57d288957be002fc0031b/php/containers.json#L251

szaimen:
I guess I can close this?

davidmehren:
Sorry this was left alone for so long, I just didn't get around to continue working on this.

Regarding your comments: I didn't add the option for the AIO Postgres container, as it doesn't really make sense to change the port when the internal database is used. The new option is only relevant when an external database is used.

The `containers.json` has been updated.

davidmehren:
@szaimen Could you reopen this PR, or would you prefer for me to open a new one?

szaimen:
> @szaimen Could you reopen this PR, or would you prefer for me to open a new one?

Please open a new one if my review is adressed

PR #3634, opened 2023-10-29
Initial add of jellyfin community-container

burnclouds:
Adding jellyfin community-container


szaimen:
Hi, thanks for the PR! Please fix DCO by following the steps in https://github.com/nextcloud/all-in-one/pull/3634/checks?check_run_id=18163243196. Apart from that you might simply create a new PR directly from the web interface which should do the sign-off automatically. We need this so that we can merge it later. Thanks!

PR #3707, opened 2023-11-08
make sure that notify-push is executable

szaimen:
Prevent things like https://help.nextcloud.com/t/aio-notify-push-container-does-not-start-permission-denied/171600/9
And https://github.com/nextcloud/all-in-one/discussions/3710
And https://github.com/nextcloud/all-in-one/issues/3727


1ubuntuuser:
Beat me to it. was going to raise an issue. I am concerned the permissions fix in the forum would have to be applied at every update.

SophyInBetween:
Looks like yesterdays update did not solve the issue.

szaimen:
> Looks like yesterdays update did not solve the issue.

Yes, it is not released yet

PR #3736, opened 2023-11-13
community containers - add memories

szaimen:
Close https://github.com/pulsejet/memories/issues/834

### DIFF
--- a/community-containers/memories/memories.json
+++ b/community-containers/memories/memories.json
@@ -1,0 +1,33 @@
+{
+    "aio_services_v1": [
+        {
+            "container_name": "nextcloud-aio-memories",
+            "display_name": "Memories Transcoder",
+            "documentation": "https://github.com/nextcloud/all-in-one/tree/main/community-containers/memories",
+            "image": "radialapps/go-vod",
+            "image_tag": "latest",
+            "internal_port": "47788",
+            "restart": "unless-stopped",
+            "environment": [
+                "TZ=%TIMEZONE%",
+                "NEXTCLOUD_HOST=https://%NC_DOMAIN%"
+            ],
+            "volumes": [
+                {
+                    "source": "%NEXTCLOUD_DATADIR%",
+                    "destination": "/mnt/ncdata",
+                    "writeable": false
+                }
+            ],
+            "devices": [
+                "/dev/dri"
+            ],
+            "nextcloud_exec_commands": [
+                "php /var/www/html/occ app:install memories",
+                "php /var/www/html/occ app:enable memories",
+                "php /var/www/html/occ config:system:set memories.vod.external --value true --type bool",
+                "php /var/www/html/occ config:system:set memories.vod.connect --value nextcloud-aio-memories:47788"
+            ]
+        }
+    ]
+}
--- a/community-containers/memories/readme.md
+++ b/community-containers/memories/readme.md
@@ -1,0 +1,12 @@
+## Memories
+This container bundles the hardware-transcoding container of memories and auto-configures it for you.
+
+### Notes
+- In order to actually enable the hardware transcoding, you need to add the following flag to AIO apart from adding this container: https://github.com/nextcloud/all-in-one#how-to-enable-hardware-transcoding-for-nextcloud
+- See https://github.com/nextcloud/all-in-one/tree/main/community-containers#community-containers how to add it to the AIO stack
+
+### Repository
+https://github.com/pulsejet/memories
+
+### Maintainer
+https://github.com/pulsejet


community-containers/memories/memories.json
```diff
@@ -0,0 +1,36 @@
...
+                    "source": "%NEXTCLOUD_DATADIR%",
+                    "destination": "/mnt/ncdata",
+                    "writeable": false
+                }
+            ],
+            "devices": [
+                "/dev/dri"
+            ],
+            "aio_variables": [
+                "nextcloud_enable_dri_device=true"
```
szaimen, line 26: I wonder is the container even useful if no /dev/dri device is present on the host?

pulsejet:
This looks neat, thanks! I'm gonna try setting up AIO on my laptop tonight and report back.

community-containers/memories/memories.json
pulsejet: Actually it is. The ffmpeg image is guaranteed to have a newer version of ffmpeg giving better format support etc.

szaimen: All right. Then I guess remocing this option makes most sense.

szaimen:
> This looks neat, thanks! I'm gonna try setting up AIO on my laptop tonight and report back.

Cool! You might find this useful: https://github.com/nextcloud/all-in-one/discussions/588#discussioncomment-3450343

community-containers/memories/memories.json
```diff
@@ -0,0 +1,36 @@
...
+            ],
```
szaimen, line 27:
```suggestion
```

community-containers/memories/readme.md
```diff
@@ -0,0 +1,11 @@
+## Memories
+This container bundles the hardware-transcoding container of memories and auto-configures it for you.
+
+### Notes
```
szaimen, line 4:
```suggestion
### Notes
- in order to actually enable the hardware transcoding, you need to add the following flag to AIO apart from adding this container: https://github.com/nextcloud/all-in-one#how-to-enable-hardware-transcoding-for-nextcloud
```

community-containers/memories/memories.json
szaimen: Done via https://github.com/nextcloud/all-in-one/pull/3736#discussion_r1391778560

community-containers/memories/memories.json
```diff
@@ -0,0 +1,33 @@
...
+                    "destination": "/mnt/ncdata",
+                    "writeable": false
+                }
+            ],
+            "devices": [
+                "/dev/dri"
+            ],
+            "nextcloud_exec_commands": [
+                "php /var/www/html/occ app:install memories",
+                "php /var/www/html/occ app:enable memories",
```
pulsejet, line 27: These don't seem to execute for me, so I needed to do the config manually. This could be something on my side though because I've a weird setup. Also, I'd prefer not auto-installing memories here, but leave that to the admin to do.

community-containers/memories/memories.json
```diff
@@ -0,0 +1,33 @@
+{
+    "aio_services_v1": [
+        {
+            "container_name": "nextcloud-aio-memories",
+            "display_name": "Transcoding container for memories",
```
pulsejet, line 5: Can we change this to "Memories Transcoder"? It's looking weirdly out of place in the UI.

pulsejet (approved):
I tested this and it works as expected. Some minor comments.

community-containers/memories/memories.json
szaimen: Yes, this is likely due to your test setup. The commands above only get executed if the nextcloud container can reach `https://$NC_DOMAIN:443` which is not the case in your testing setup.

I would actually like to keep it. Is that fine for you?

community-containers/memories/memories.json
szaimen: done

community-containers/memories/memories.json
pulsejet: Ah okay.

> I would actually like to keep it. Is that fine for you?

Sure I don't mind. Maybe worth a line in the description then that it also installs the app? I'm also assuming this is a no-op if already installed.

szaimen: > Sure I don't mind. Maybe worth a line in the description then that it also installs the app?

I may add this as a follow-up 👍

> I'm also assuming this is a no-op if already installed.

Yes

pulsejet:
Thanks for helping out with this 🎉

szaimen:
No problem :)

szaimen:
This is now released with 7.7.0 Beta. Testing and feedback is welcome! See https://github.com/nextcloud/all-in-one#how-to-switch-the-channel

PR #3756, opened 2023-11-16
helm chart - allow to configure additional values

szaimen:
Close https://github.com/nextcloud/all-in-one/discussions/3742


Containers/nextcloud/config/apps.config.php
```diff
@@ -12,4 +12,5 @@
               'writable' => true,
       ),
   ),
+  'appsallowlist' => getenv('APPS_ALLOWLIST') ? explode(" ", getenv('APPS_ALLOWLIST')) : [],
```
szaimen, line 15: I am not completely sure about this syntax. Does this look good?

juliushaertl: 👍

Containers/nextcloud/config/smtp.config.php
```diff
@@ -0,0 +1,20 @@
+<?php
+if (getenv('SMTP_HOST') && getenv('MAIL_FROM_ADDRESS') && getenv('MAIL_DOMAIN')) {
+  $CONFIG = array (
+    'mail_smtpmode' => 'smtp',
```
juliushaertl, line 4: I'm wondering if this could not already work out of the box with https://github.com/nextcloud/server/pull/3966

Containers/nextcloud/config/apps.config.php
szaimen: Thanks! Does the rest of the PR also look good? :) (no testing required)

Containers/nextcloud/config/smtp.config.php
juliushaertl: Just looking at the code and assuming that the kubernetes environment variables are properly passed to the PHP process it could already work with `NC_mail_smtpmode=smtp`

szaimen: Yes possibly but I find it cleaner via config.php. also cia direct env is rather limited in what you can do

szaimen: Btw, I copied this file from https://github.com/nextcloud/docker/blob/master/.config/smtp.config.php

juliushaertl (approved):

szaimen:
Thanks Julius! :)

szaimen:
This is now released with 7.7.0 Beta. Testing and feedback is welcome! See https://github.com/nextcloud/all-in-one#how-to-switch-the-channel

PR #3779, opened 2023-11-20
Fix collabora in manual-install Yaml

rahilarious:
Making it consistent with docs https://docs.nextcloud.com/server/stable/admin_manual/office/example-docker.html

When I ran compose with `--profile collabora` , Fiddling with logs showed me bunch of   "permission denied". Adding CAP_MKNOD fixes it

### DIFF
--- a/manual-install/latest.yml
+++ b/manual-install/latest.yml
@@ -198,6 +198,8 @@ services:
       - server_name=${NC_DOMAIN}
       - DONT_GEN_SSL_CERT=1
     restart: unless-stopped
+    cap_add:
+      - MKNOD
     profiles:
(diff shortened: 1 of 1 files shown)


szaimen:
Hi, this needs to be added to https://github.com/nextcloud/all-in-one/pull/3781 first.

rahilarious:
> Hi, this needs to be added to #3781 first.

so I'm guessing adding it to that yaml is automatically generated from containers.json?

szaimen:
> > Hi, this needs to be added to #3781 first.
> 
> so I'm guessing that yaml is automatically generated from containers.json?

yes :)

PR #3788, opened 2023-11-21
test if hostname makes it work

szaimen:


st3iny (changes requested):
Kubernetes complains when installing the manifest: `Error: UPGRADE FAILED: failed to create resource: Service "no-domain.com" is invalid: metadata.name: Invalid value: "no-domain.com": a DNS-1035 label must consist of lower case alphanumeric characters or '-', start with an alphabetic character, and end with an alphanumeric character (e.g. 'my-name',  or 'abc-123', regex used for validation is '[a-z]([-a-z0-9]*[a-z0-9])?')`

PR #3792, opened 2023-11-21
helm - remove lost+found in nextcloud

szaimen:


nextcloud-aio-helm-chart/update-helm.sh
```diff
@@ -98,6 +98,22 @@ cat << EOL > /tmp/initcontainers.clamav
...
 EOL
+cat << EOL > /tmp/initcontainers.nextcloud
+      initContainers:
+        - name: delete lost+found
+          image: alpine
+          command:
+            - rm
+            - "-rf"
+            - /nextcloud-aio-nextcloud/lost+found
+          volumeMountsInitContainer:
```
szaimen, line 109: TODO: do not delete volumes...

PR #3807, opened 2023-11-23
helm - add livenessprobe to apache

szaimen:


nextcloud-aio-helm-chart/templates/nextcloud-aio-apache-deployment.yaml
```diff
@@ -64,6 +64,13 @@ spec:
               protocol: TCP
             - containerPort: {{ .Values.APACHE_PORT }}
               protocol: UDP
+          livenessProbe:
+            exec:
+              command:
+              - bash
+              - /healthcheck.sh
+            failureThreshold: 1
+            periodSeconds: 10
```
szaimen, line 73: this needs to be added via script. and I guess it will not work yet due to https://github.com/nextcloud/all-in-one/blob/a43f75542ecc4fbef0590cd9737886ee0f45e437/Containers/apache/healthcheck.sh#L3 which is exactly what we want to catch here...

szaimen: Also I wonder if we would basically need liveness probes for all containers so that they get restarted if some other container got missing in the meantime...

szaimen: Another solution would be to put everything into one pod but doesnt sound like the best solution either...

Issue #3908, opened 2023-12-08
Change install_latest_major update logic


szaimen:
```
$ php occ config:app:get core installedat 2>/dev/null
1539092232.9661
$ php occ config:app:set core installedat --value="1539092232.9600" 
Config value installedat for app core set to 1539092232.9600
```

Setting it to 00 allows to work around staging

Also add no-backup flag

PR #3915, opened 2023-12-11
use correct way to do the update to new major versions

szaimen:
Close https://github.com/nextcloud/all-in-one/issues/3908

This is best reviewed like this: https://github.com/nextcloud/all-in-one/pull/3915/files?w=1

### DIFF
--- a/Containers/nextcloud/entrypoint.sh
+++ b/Containers/nextcloud/entrypoint.sh
@@ -272,6 +272,18 @@ DATADIR_PERMISSION_CONF
                 fi
                 # shellcheck disable=SC2016
                 installed_version="$(php -r 'require "/var/www/html/version.php"; echo implode(".", $OC_Version);')"
+                INSTALLED_MAJOR="${installed_version%%.*}"
+                IMAGE_MAJOR="${image_version%%.*}"
+                if ! [ "$INSTALLED_MAJOR" -gt "$IMAGE_MAJOR" ]; then
+                    php /var/www/html/updater/updater.phar --no-interaction --no-backup
+                    if ! php /var/www/html/occ -V || php /var/www/html/occ status | grep maintenance | grep -q 'true'; then
+                        echo "Installation of Nextcloud failed!"
+                        touch "$NEXTCLOUD_DATA_DIR/install.failed"
+                        exit 1
+                    fi
+                    # shellcheck disable=SC2016
+                    installed_version="$(php -r 'require "/var/www/html/version.php"; echo implode(".", $OC_Version);')"
+                fi
                 php /var/www/html/occ app:disable updatenotification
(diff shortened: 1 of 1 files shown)


Containers/nextcloud/entrypoint.sh
```diff
@@ -258,26 +258,21 @@ DATADIR_PERMISSION_CONF
             if [ "$INSTALL_LATEST_MAJOR" = yes ]; then
                 php /var/www/html/occ config:system:set updater.release.channel --value=beta
```
blizzz, line 259: still want to stay on beta?

szaimen: Ah right, forgot to remove that

szaimen: done with https://github.com/nextcloud/all-in-one/pull/3915/commits/6412aa3e0f4262e5fa6b163be7efb3bbd2ed5e2e

Containers/nextcloud/entrypoint.sh
```diff
@@ -256,31 +256,24 @@ DATADIR_PERMISSION_CONF
...
             if [ "$INSTALL_LATEST_MAJOR" = yes ]; then
-                php /var/www/html/occ config:system:set updater.release.channel --value=beta
                 php /var/www/html/occ config:system:set updatedirectory --value="/nc-updater"
-                php /var/www/html/updater/updater.phar --no-interaction
+                INSTALL_DAT="$(php /var/www/html/occ config:app:get core installedat)"
+                if [ -n "$INSTALL_DAT" ]; then
+                    # Set the installdat to 00 which will allow to skip staging and install the next major directly
+                    # shellcheck disable=SC2001
+                    INSTALL_DAT="$(echo "$INSTALL_DAT" | sed "s|[0-9][0-9]$|00|")"
+                    php /var/www/html/occ config:app:set core installedat --value="$INSTALL_DAT"
```
blizzz, line 265:
```suggestion
                INSTALLED_AT="$(php /var/www/html/occ config:app:get core installedat)"
                if [ -n "${INSTALLED_AT}" ]; then
                    # Set the installdat to 00 which will allow to skip staging and install the next major directly
                    # shellcheck disable=SC2001
                    INSTALLED_AT="$(echo "${INSTALLED_AT}" | sed "s|[0-9][0-9]$|00|")"
                    php /var/www/html/occ config:app:set core installedat --value="${INSTALLED_AT}" 
```

blizzz (approved):
with one suggestion

szaimen:
Actually, I just figured that we need the version check in some cases...

szaimen:
> Actually, I just figured that we need the version check in some cases...

done

blizzz (approved):

szaimen:
Thanks Arthur! :)

PR #3972, opened 2023-12-28
Add context to custom skeleton directory howto in readme

### LINKED ISSUE #3968: Custom skeleton directory section of README does not describe what problem this solves

The README documentation file has a section entitled [Custom skeleton directory](https://github.com/nextcloud/all-in-one#custom-skeleton-directory) which details how an administrator can create a custom skeleton directory.  However this section does not say why you would wish to do this - it assumes an understanding that the skeleton directory is used to change the default files and folders created for a new user account.  The reader may not have this prior understanding, meaning the solution is without context.

The section should include a description of why an administrator would wish to change the skeleton directory, similar to the description provided in the [providing default files](https://docs.nextcloud.com/server/latest/admin_manual/configuration_files/default_files_configuration.html) section of the primary Nextcloud documentation.

I intend to submit a pull request proposing a text change.

Administrators can search elsewhere to gain an understanding of why they would wish to change the skeleton directory (i.e. the problem to be solved), but it seems more logical to include the context with the solution.

### Additional context

This could be considered a distant relation of the [XY problem](https://www.perlmonks.org/index.pl?node_id=542341), we get the solution 'Y' but not the problem 'X.'

PreciousChicken:
Currently the README section that describes how to change the custom skeleton directory does not explain why an administrator may wish to do this.  This addition to the documentation provides context: a custom skeleton directory is created to change the default files and folders that ship with Nextcloud.  Fix #3968 .


szaimen (approved):
LGTM, thanks! :)

PR #3999, opened 2024-01-03
community-containers - add facerecognition

szaimen:
Close https://github.com/goodspb/pdlib/issues/56


szaimen:
@matiasdelellis I invited you to the repo so that you can review the PR :)

szaimen:
FYI @pulsejet

community-containers/facerecognition/facerecognition.json
```diff
@@ -0,0 +1,29 @@
...
+                "nextcloud_memory_limit=4096M"
+            ],
+            "nextcloud_exec_commands": [
+                "php /var/www/html/occ app:install facerecognition",
+                "php /var/www/html/occ app:enable facerecognition", 
+                "php /var/www/html/occ config:system:set facerecognition.external_model_url --value nextcloud-aio-facerecognition:5000",
+                "php /var/www/html/occ config:system:set facerecognition.external_model_api_key --value some-super-secret-api-key",
+                "php /var/www/html/occ face:setup -m 5",
+                "php /var/www/html/occ face:setup -M 4G",
+                "php /var/www/html/occ face:background_job &"
```
matiasdelellis, line 25: This last command is not necessary. The administrator should run it periodically.

* https://github.com/matiasdelellis/facerecognition/wiki/Schedule-Background-Task

If you added it to do an initial scan, you should add the `--defer-clustering` option,  but probably don't want this.

community-containers/facerecognition/facerecognition.json
```diff
@@ -0,0 +1,29 @@
...
+                "API_KEY=some-super-secret-api-key"
+            ],
+            "aio_variables": [
+                "nextcloud_memory_limit=4096M"
+            ],
+            "nextcloud_exec_commands": [
+                "php /var/www/html/occ app:install facerecognition",
+                "php /var/www/html/occ app:enable facerecognition", 
+                "php /var/www/html/occ config:system:set facerecognition.external_model_url --value nextcloud-aio-facerecognition:5000",
+                "php /var/www/html/occ config:system:set facerecognition.external_model_api_key --value some-super-secret-api-key",
```
matiasdelellis, line 22: The model_key should be configurable in some way, but since here we run all the services automatically on the same host, we can ignore it in this context.. 🤔

community-containers/facerecognition/facerecognition.json
```diff
@@ -0,0 +1,29 @@
...
+                "php /var/www/html/occ face:setup -m 5",
+                "php /var/www/html/occ face:setup -M 4G",
```
matiasdelellis, line 24: It's probably a good idea to set the temporary image size here. It is not necessary, but in the context of so many automations it would avoid confusion. 🤔 
Again, it would just be to avoid confusion, and I prefer that users know what they are doing so let's ignore this. 🙈

matiasdelellis (commented):
I am still a stranger to nextcloud AIO, therefore I cannot comment more than this.

The only important thing is to remove the background_job command

community-containers/facerecognition/facerecognition.json
szaimen: The --defer-clustering sounds good, thanks! The idea is to actually make it as easy as possible which would mean that the background job should run inifinitely and also do everything automatically...

community-containers/facerecognition/facerecognition.json
szaimen: yes, the plan is to automatically assign a real secret later on but for that some pre-requirements would be needed which is tracked in https://github.com/nextcloud/all-in-one/discussions/3889

community-containers/facerecognition/facerecognition.json
szaimen:
```suggestion
                "php /var/www/html/occ face:setup -M 4G",
                "php /var/www/html/occ config:app:set facerecognition analysis_image_area --value="4320000",
```

szaimen:
Thanks for the review @matiasdelellis! :)
Should all be resolved with https://github.com/nextcloud/all-in-one/pull/3999/commits/8fc41130a0d15691574b87c037b223b517e504e0

xieminghuang:
fgfd

szaimen:
Hi @matiasdelellis does the PR look good to you now? :)
(No testing required as I will take care of this)

szaimen:
Thank you @matiasdelellis! :)

oliverhu:
@szaimen curious on this line - `php /var/www/html/occ config:system:set facerecognition.external_model_api_key --value some-super-secret-api-key`, probably we should make the key field configurable?

szaimen:
> @szaimen curious on this line - `php /var/www/html/occ config:system:set facerecognition.external_model_api_key --value some-super-secret-api-key`, probably we should make the key field configurable?

See https://github.com/nextcloud/all-in-one/pull/3999#discussion_r1442679455

szaimen:
This is now released with v7.10.0 Beta. Testing and feedback is welcome! See https://github.com/nextcloud/all-in-one#how-to-switch-the-channel

PR #4002, opened 2024-01-04
Include caldav/carddav rewrite in Traefik 2 documentation

RlndVt:
I'm no Traefik expert, but I found [1] explaining how to implement the redirect using regex matching. I thought it was a pretty neat way of doing it :). 

Should be a nice addition to the documentation.

[1] https://community.traefik.io/t/nextcloud-with-cal-cardav-routing/5553/3

### DIFF
--- a/reverse-proxy.md
+++ b/reverse-proxy.md
@@ -601,15 +601,21 @@ The examples below define the dynamic configuration in YAML files. If you rather
                         - "X-Forwarded-Host"
                     referrerPolicy: "same-origin"
 
+            nextcloud-dav:
+                replacepathregex:
+                    regex: "^/.well-known/ca(l|rd)dav"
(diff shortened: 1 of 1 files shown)


Zoey2936 (changes requested):
not needed: https://github.com/nextcloud/all-in-one/blob/5ca8025dcdf32ded868628f3501538339f6b89b7/Containers/apache/Caddyfile#L56

RlndVt:
Oh interesting, I thought that the Caddy part of the Apache container was skipped when you ran in reverse proxy mode but I see now it is always used.

[internet] :443: ([reverse-proxy] :11000: ) [caddy] : 8000: [Apache] :9000: [nextcloud].

Makes me wonder why Apache is used at all, doesn't Caddy have a good fpm engine?

PR #4046, opened 2024-01-12
Make port 443 error message more verbose

szaimen:
For https://github.com/nextcloud/all-in-one/discussions/4041

### DIFF
--- a/php/src/Data/ConfigurationManager.php
+++ b/php/src/Data/ConfigurationManager.php
@@ -331,7 +331,7 @@ class ConfigurationManager
             if ($connection) {
                 fclose($connection);
             } else {
-                throw new InvalidSettingConfigurationException("The server is not reachable on Port 443. You can verify this e.g. with 'https://portchecker.co/' by entering your domain there as ip-address and port 443 as port.");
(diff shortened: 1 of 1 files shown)


php/src/Data/ConfigurationManager.php
```diff
@@ -331,7 +331,7 @@ public function SetDomain(string $domain) : void {
             if ($connection) {
                 fclose($connection);
             } else {
-                throw new InvalidSettingConfigurationException("The server is not reachable on Port 443. You can verify this e.g. with 'https://portchecker.co/' by entering your domain there as ip-address and port 443 as port.");
+                throw new InvalidSettingConfigurationException("The domain is not reachable on Port 443 from within this container. Have you opened port 443/tcp in your router/firewall? If yes is it most likely that the router or firewall forbids local access to your domain. You can work around that by setting up a local DNS-server.");
```
szaimen, line 334:
```suggestion
                throw new InvalidSettingConfigurationException("The domain is not reachable on Port 443 from within this container. Have you opened port 443/tcp in your router/firewall? If yes is the problem most likely that the router or firewall forbids local access to your domain. You can work around that by setting up a local DNS-server.");
```

PR #4065, opened 2024-01-15
add further hint for what could be wrong

szaimen:
For https://github.com/nextcloud/all-in-one/discussions/4008

### DIFF
--- a/Containers/mastercontainer/start.sh
+++ b/Containers/mastercontainer/start.sh
@@ -70,6 +70,7 @@ fi
 # Check if api version is supported
 if ! sudo -u www-data docker info &>/dev/null; then
     print_red "Cannot connect to the docker socket. Cannot proceed."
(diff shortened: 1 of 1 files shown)


Containers/mastercontainer/start.sh
```diff
@@ -70,6 +70,7 @@ fi
 # Check if api version is supported
 if ! sudo -u www-data docker info &>/dev/null; then
     print_red "Cannot connect to the docker socket. Cannot proceed."
+    echo "Did you maybe remove group read permissions for the docker socket? AIO needs them in order to access the docker socket."
```
zine999, line 73: > Did you maybe remove group read permissions

This should read:

> Did you maybe remove group write permissions

szaimen: I think read permissions should be enough.

zine999: @szaimen ... did you test it? because I did, and it was the write permission that it needed...

PR #4116, opened 2024-01-26
Drop NET_RAW from all containers in manual

jhesketh:
#3377 drops NET_RAW from all containers, but this doesn't appear to have been adopted into the manual mode.

### DIFF
--- a/manual-install/latest.yml
+++ b/manual-install/latest.yml
@@ -45,6 +45,8 @@ services:
       - /usr/local/apache2/logs
       - /tmp
       - /home/www-data
+    cap_drop:
+      - NET_RAW
 
   nextcloud-aio-database:
     image: nextcloud/aio-postgresql:latest
@@ -68,6 +70,8 @@ services:
     read_only: true
     tmpfs:
       - /var/run/postgresql
+    cap_drop:
+      - NET_RAW
 
   nextcloud-aio-nextcloud:
     depends_on:
@@ -149,6 +153,8 @@ services:
     restart: unless-stopped
     networks:
       - nextcloud-aio
+    cap_drop:
+      - NET_RAW
 
   nextcloud-aio-notify-push:
     image: nextcloud/aio-notify-push:latest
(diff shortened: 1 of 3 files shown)


jhesketh:
As an aside: Are there any checks to detect configuration drift from AIO and Manual modes? Perhaps a github check could help find these kind of bugs?

Also, I wonder if more capabilities could be dropped from all of the containers.

Zoey2936:
you need to add this to the https://github.com/nextcloud/all-in-one/blob/main/manual-install/update-yaml.sh, otherwise your changes will be overridden when the file is regenerated

jhesketh:
@Zoey2936 Just wondering if you could please take another look at this when you get a chance. Thanks!

Zoey2936:
I still think it needs to be added to the update-yaml.sh, since the latest.yaml will be automatically overriden and if you add it to the containers.json the containers will get the value twice which could cause errors

jhesketh:
Right, I see, thanks.

Wouldn't it be better to have the capabilities listed in `containers.json` and drop modifying it from [DockerActionManager.php](https://github.com/nextcloud/all-in-one/blob/main/php/src/Docker/DockerActionManager.php#L529)?

Rationale:

1. `cap_add` [already exists in containers.json](https://github.com/nextcloud/all-in-one/blob/main/php/containers.json#L329), meaning having `cap_drop` is more consistent
2. If I'm reading it correctly, any `cap_drop` (regardless if it is `NET_RAW` or something else) would be removed because [this line](https://github.com/nextcloud/all-in-one/blob/main/php/src/Docker/DockerActionManager.php#L529) overwrites the `cap_drop` array instead of appending a specific term.

Zoey2936:
Not sure @szaimen did this

jhesketh:
Looking a bit closer, it seems this is a little more non-trivial than I expected. Specifically cap_drop isn't supported in containers.json. It would need loading in ContainerDefinitionFetcher.php and passed into the Container() constructor.

All doable, and possibly the neater solution due to the above rationale, but more work than placing a `sed` into update-yaml.sh.

I'll await @szaimen's opinion before bothering to pipe it through the definition fetcher.

szaimen (approved):
LGTM

PR #4119, opened 2024-01-26
downgrade back to 27.1.5

szaimen:
Due to https://github.com/nextcloud/server/issues/43134


pierreozoux:
@szaimen it is not dangerous to downgrade minor?

szaimen:
> @szaimen it is not dangerous to downgrade minor?

If you are already on 27.1.6 you will not be downgraded. But if you were on 27.1.5 you will stay on 27.1.5

PR #4188, opened 2024-02-06
fix: create elasticsearch pvc only if fulltextsearch is enabled

alinefr:
At the moment, this chart is creating elasticsearch pvc even if we don't need it.


szaimen:
This has been done already: https://github.com/nextcloud/all-in-one/blob/main/nextcloud-aio-helm-chart/templates/nextcloud-aio-elasticsearch-persistentvolumeclaim.yaml

PR #4210, opened 2024-02-11
hider server and x-powered-by header

Zoey2936:
I've changed the level of my NPMplus from error to warn today, now the log spams:
`upstream sent duplicate header line: "Server: Apache/2.4.58 (Unix)", previous value: "Server: Caddy", ignored while reading response header from upstream`
To fix, I think the Server header should be completely removed (so the apache and the caddy one). I also think the X-Powered-By header should be hidden, since it doesn't need to be public visible

### DIFF
--- a/Containers/apache/Caddyfile
+++ b/Containers/apache/Caddyfile
@@ -16,6 +16,8 @@
 
 https://{$ADDITIONAL_TRUSTED_DOMAIN}:443,
 {$PROTOCOL}://{$NC_DOMAIN}:{$APACHE_PORT} {
+    header -Server
+    header -X-Powered-By
 
     # Collabora
     route /browser/* {


szaimen:
Hi Zoey, thanks for your PR! I cannot really judge this PR unfortunately :/

Zoey2936:
Currently, caddy sends its own Server header, but also copies the header of upstream Apache, so there are two Server headers and this causes a warn in nginx. To fix this this PR removes both Server headers so nginx and any other reverse proxy will recieve no server header and may at its own server header. Also browser can not see the version of apache running behind the proxy and by hiding the X-Powered-By header clients can also not any more see the php version running in the background.

Altahrim (approved):

szaimen:
This is now released with v7.13.0 Beta. Testing and feedback is welcome! See https://github.com/nextcloud/all-in-one#how-to-switch-the-channel

Issue #4183, opened 2024-02-05
AIO Linode becomes unresponsive after a few days


MariusQuabeck:
Every ~10 days, my Nextcloud instance becomes unresponsive

- Webinterface wont load
- calendar sync stops working
- even on ssh, login hangs and times out
- AIO:8443 login prompt works but never finishes loading after submitting password

I literally can't access the vm or logs without rebooting the VM from Linodes web interface, so I'm not exactly sure how/where to get logs for this.
A forced reboot from Linodes web interface solves the issue for at least a few days

this is a AIO install from their marketplace/store, no custom changes AFAIK

here is what Linode analytics look like while the issue occurs
<img width="1303" alt="image" src="https://github.com/nextcloud/all-in-one/assets/17063564/7804052f-57f2-4532-9826-801223c2fb26">

in this instance, AIO has been unresponsive for almost 3 days before I've noticed
<img width="664" alt="image" src="https://github.com/nextcloud/all-in-one/assets/17063564/37234a44-7db4-46be-880f-9519f535fe5f">

szaimen:
Hi Marius, hope you are doing fine? 👋

Can you please post the output of `sudo docker info` here? :)

MariusQuabeck:
I'm doing great, thank you. I hope we'll run into each other at some event soon

`sudo docker info`

```
Client: Docker Engine - Community
 Version:    24.0.7
 Context:    default
 Debug Mode: false
 Plugins:
  buildx: Docker Buildx (Docker Inc.)
    Version:  v0.11.2
    Path:     /usr/libexec/docker/cli-plugins/docker-buildx
  compose: Docker Compose (Docker Inc.)
    Version:  v2.21.0
    Path:     /usr/libexec/docker/cli-plugins/docker-compose

Server:
 Containers: 11
  Running: 9
  Paused: 0
  Stopped: 2
 Images: 12
 Server Version: 24.0.7
 Storage Driver: overlay2
  Backing Filesystem: extfs
  Supports d_type: true
  Using metacopy: false
  Native Overlay Diff: true
  userxattr: false
 Logging Driver: json-file
 Cgroup Driver: systemd
 Cgroup Version: 2
 Plugins:
  Volume: local
  Network: bridge host ipvlan macvlan null overlay
  Log: awslogs fluentd gcplogs gelf journald json-file local logentries splunk syslog
 Swarm: inactive
 Runtimes: io.containerd.runc.v2 runc
 Default Runtime: runc
[...]

PR #4230, opened 2024-02-16
dissable init for containers again

szaimen:
For https://github.com/nextcloud/all-in-one/issues/4183


szaimen:
This is not the reason for the problem.

PR #4234, opened 2024-02-16
Add extra informations about root-data change for Docker

AiroPi:
This follows https://github.com/nextcloud/all-in-one/discussions/4224


szaimen (approved):
Thanks!

PR #4235, opened 2024-02-16
Fix pi-hole dashboard page url

AiroPi:
The url provided in the readme returns a "403 Forbidden" error. The right page is `/admin`.

https://github.com/pi-hole/web?tab=readme-ov-file#post-installation-access-the-web-interface-and-gain-insight-into-your-networks-activity


szaimen (approved):
Right, thanks!

PR #4249, opened 2024-02-19
fix: Do not apply an empty allow list if none is set

juliushaertl:
Otherwise the empty array could block showing any apps from the store in apps management

### DIFF
--- a/Containers/nextcloud/config/apps.config.php
+++ b/Containers/nextcloud/config/apps.config.php
@@ -12,5 +12,5 @@ $CONFIG = array (
               'writable' => true,
       ),
   ),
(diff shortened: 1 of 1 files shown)


szaimen (approved):
Thanks!

szaimen:
This is now released with v7.13.0 Beta. Testing and feedback is welcome! See https://github.com/nextcloud/all-in-one#how-to-switch-the-channel

corincorvus:
We are on v7.13.0 Beta now with update nextcloud on 28.0.2.
image

Empty Store:
image

szaimen:
Hi @corincorvus I fear the change to this file only gets applied after Nextcloud gets updated. However you can modify the file yourself manually.

corincorvus:
> However you can modify the file yourself manually.

Hey,
its done, it work! Thanks.

PR #4274, opened 2024-02-25
Add information about virtual machines

BraidenPsiuk:
This change adds a guide for provisioning multiple VMs to run their own AIO instances. My idea is to provide a thorough guide that is easy to follow and gets readers up and running quickly, but without going too deep, unless the reader seeks to delve into advanced configuration. I wrote this because I tried going through this process myself, and spent hours struggling to piece everything together.

If the guide is too long or does not fit in with the rest of the material I am happy to provide changes. I also hope that somebody with more knowledge can provide better info for getting Nextcloud Talk to work.


multiple-instances.md
```diff
@@ -3,17 +3,114 @@ It is possible to run multiple instances of AIO on one server.
 
 There are two ways to achieve this: The normal way is creating multiple VMs, installing AIO in [reverse proxy mode](./reverse-proxy.md) in each of them and having one reverse proxy in front of them that points to each VM (you also need to [use a different `TALK_PORT`](https://github.com/nextcloud/all-in-one#how-to-adjust-the-talk-port) for each of them). The second and more advanced way is creating multiple users on the server and using docker rootless for each of them in order to install multiple instances on the same server. 
 
-Below is described more in detail how the the second way works.
+## Run multiple AIO instances on the same server inside their own virtual machines
```
szaimen, line 6: Can we please move this section below the docker rootless section?

multiple-instances.md
```diff
@@ -3,17 +3,114 @@ It is possible to run multiple instances of AIO on one server.
...
...ge the Talk port for a given instance, you must either have created it that way initially (back when you first configured the VM in step 3 above), or you can change it at any time by removing the mastercontainer and re-running the initial docker run command (but modifying the run command <a href="https://github.com/nextcloud/all-in-one?tab=readme-ov-file#how-to-adjust-the-talk-port">like so</a>). Then, the Talk port for EACH instance needs to be forwarded in your router's settings DIRECTLY to the VM hosting the instance (completely bypassing our physical host machine/reverse proxy). And finally, inside an admin-priveleged account (such as the default "admin" account) in each instance, you must visit <pre>https://[DOMAIN_NAME_*]/settings/admin/talk</pre> then find the STUN/TURN Settings, and from there set the proper values. If this is too complicated, it may be easier to use public STUN/TURN servers, but I have not tested any of this, rather I'm just sharing what I have found so far (more info available <a href="https://github.com/nextcloud/all-in-one/discussions/2517">here</a>). If you have figured this out or if any of this information is incorrect, please edit this section!</li>
```
szaimen, line 99: > and each AIO instance inside each VM must be configured with a different Talk port (like 3478, 3479, etc.). To change the Talk port for a given instance, you must either have created it that way initially (back when you first configured the VM in step 3 above), or you can change it at any time by removing the mastercontainer and re-running the initial docker run command (but modifying the run command <a href="https://github.com/nextcloud/all-in-one?tab=readme-ov-file#how-to-adjust-the-talk-port">like so</a>).

indeed, that is why I would usually recommend to set the talk port already during initial install

multiple-instances.md
```diff
@@ -3,17 +3,114 @@ It is possible to run multiple instances of AIO on one server.
...
+1. Log in to your new VM. After it's finished installing, the VM will have rebooted and presented you with a login prompt. Use `root` as the username, and enter the password you chose during the installation process. Then, run this command (**on the VM**):
+   ```shell
+   apt install -y curl && curl -fsSL https://get.docker.com | sh && docker run --init --sig-proxy=false --name nextcloud-aio-mastercontainer --restart always --publish 8080:8080 --env APACHE_PORT=11000 --env APACHE_IP_BINDING=0.0.0.0 --volume nextcloud_aio_mastercontainer:/mnt/docker-aio-config --volume /var/run/docker.sock:/var/run/docker.sock:ro nextcloud/all-in-one:latest
```
szaimen, line 37: this is missing the TALK_PORT variable which needs to be set to different values on each instance

szaimen (commented):
Thanks a lot for this nice instructions! I have some remarks below.

And sorry that it took so long to review this!

multiple-instances.md
BraidenPsiuk: Certainly. Will change.

BraidenPsiuk: Changed in a32f96c72af3d651b9d5e05ca2b2e174eb050b2c

multiple-instances.md
BraidenPsiuk: Makes sense. I'll add the TALK_PORT variable to the run command with a note telling the reader to increment it.

multiple-instances.md
BraidenPsiuk: Will add.

BraidenPsiuk:
This is my first "real" review process, so please feel free to correct me if I do anything that doesn't seem conventional.

No problem about the wait! It's actually good. I've been testing this setup with two VMs/AIO instances in a small business office network (static IP, daily file downloads from a small number of customers via web shares, etc.) for the past month, and these instances have been running very smoothly, including the auto backups/updates. So it's given some time to make sure no weird issues pop up. It also feels pretty easy to manage this setup, not to mention the cost savings and speed boost that come with self-hosting too.

Despite it mostly going well, I do have a few thoughts/decisions for the guide that I've been struggling to decide on my own. I'll list them and maybe you could help decide?

1. I've been testing on a physical host machine with a 4TB SSD. One of my two VMs is very storage-heavy and makes use of ~1.5TB (the other VM only uses ~200GB). I initially allocated 1.7TB to the larger VM, but recently decided it was time to expand this above 2TB. So I tried growing the root partition to something like 2.5TB, but found I was unable to, because the Debian installer initially made the decision to use MBR as the partitioning scheme (and MBR doesn't allow >2TB. GPT seems to be necessary for this).

   I learned that if you initially allocate >=2TB to the VM before running the Debian installer, the installer will use the GPT scheme. But I have a feeling many people will not start off with quite that much storage, but may expand to that at a later point (for example, by adding more drives to their volume group if they used LVM or similar). I also learned that the only way to "force" the installer to use GPT (without initially allocating >=2TB) is to boot the VM in UEFI mode, which requires 1 extra package on the host, `ovmf`, and 1 extra flag, `--boot uefi`, added to the `virt-install` VM creation command ([more info](https://unix.stackexchange.com/a/518378)). Unfortunately there's no easy option to just set this in the installer. It's apparently possible to [change from MBR to GPT](https://unix.stackexchange.com/a/662459) after the initial setup, but it's very risky, and when I tried it left me with an unbootable VM- so not good to recommend this approach.

   *TL;DR: Most readers following the guide currently will not be able grow above 2TB if they didn't already set it up that way. If they store their files in the default docker volume, this will give them trouble if they ever want to grow past 2TB. The use of Virtualization is nice since we can mostly add/remove resources whenever we need, so it would be unfortunate to run into this storage limitation. Changing the guide to include UEFI boot seems to fix this.*

   Therefore: **Should the guide use UEFI mode by default when creating VMs?** I have no idea what the average amount of storage self-hosting NC admins usually allocate is. So I don't know how many people this may potentially help. And using UEFI doesn't seem to have any downsides, but I'd rather be 100% certain. Maybe instead, this info should just go in the "extra tips" section. But it's one of those things you don't know you need, until you really need it. For me, I had to migrate my large instance to a new VM with UEFI/GPT and now it's working great. Sorry this point is so long but I feel it deserves the most consideration.
2. I addressed the TALK port stuff in your review comments and will add this. But as far as actually testing Talk with my instances I haven't found the time lately for getting that working. I haven't looked into what is needed to bridge the VMs yet and want to learn more about this first. The company I'm running these instances in will benefit from me enabling Talk though, so I do hope to get that working over the next few weeks/months, and will gladly write the docs for that part too unless someone beats me to it. **Do you want this PR to remain open until Talk is fully finished being documented?**
3. I used long, chained commands for convenience and readability of the guide as a whole, especially since the reader is expected to repeat steps and the commands are explained in better detail on the main README. But I know most people prefer the multiline commands for safety and better understanding. **Should I switch to multiline commands?** Also any security concerns for the "dl.cloudsmith.io" URL?
4. **Does Debian still seem good as the example distro** or is there perhaps a more suitable one?
5. **Is it okay to still recommend enabling SSH in the installer?** The alternative would be just suggesting `virsh console` and leaving SSH up to the reader to decide to install later. (I've been using virsh console lately to manage mine, but sometimes you get bad graphical "tearing" with long commands, also lacks text color).


Other small changes I want to do, just making sure they sound good as well:
1. Replace all uses of term "physical host machine" with "host physical machine" - To follow what [Red Hat docs use](https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux/7/html-single/virtualization_deployment_and_administration_guide/index), especially since we're using virsh
2. Small mention (somewhere around the top) that all commands are expected to be run as root
3. Include, after VM setup instructions, that `Ctrl + ]` lets reader detach from console session once they're done configuring a VM

szaimen:
> 1. **Should the guide use UEFI mode by default when creating VMs?**

I guess this makes sense, yes



> 2\. Do you want this PR to remain open until Talk is fully finished being documented?

No, feel free to create a follow-up PR

> 3\. **Should I switch to multiline commands?** 

Yes, this would be great!

> Also any security concerns for the "dl.cloudsmith.io" URL?

The cloudsmith.io is from the official caddy docs so should be fine



> 4\. Does Debian still seem good as the example distro

I would usually recommend Ubuntu Server LTS

> 5\. Is it okay to still recommend enabling SSH in the installer?

Yes, I'd say so

> * Replace all uses of term "physical host machine" with "host physical machine" - To follow what [Red Hat docs use](https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux/7/html-single/virtualization_deployment_and_administration_guide/index), especially since we're using virsh
> * Small mention (somewhere around the top) that all commands are expected to be run as root
> * Include, after VM setup instructions, that `Ctrl + ]` lets reader detach from console session once they're done configuring a VM

Yes, sounds good to me 👍

szaimen:
Hi @BraidenPsiuk do you have any update here? :)

BraidenPsiuk:
> Hi @BraidenPsiuk do you have any update here? :)

Yes! Sorry for the long delay. Was finishing up a data structures/algorithms course that took a lot of my time :) Anyway, [Here's a preview](https://github.com/InnerView-Ultrasound/aio-setup/blob/main/NEW.md) of my changes which I'll still be rapidly updating tonight. I'll have it finished and those changes **pushed here for review within the next 8 hours**.

I added specific instructions for Ubuntu Server and I'm testing that on hardware to make sure everything works as expected.

multiple-instances.md
BraidenPsiuk: Added in 3f80f7be6d5dfa663a1f98cfdbece64033726560

multiple-instances.md
BraidenPsiuk: Added in 3f80f7be6d5dfa663a1f98cfdbece64033726560

BraidenPsiuk:
Hi again, just letting you know that I've submitted the updated version of the guide. Any/all feedback welcome!

multiple-instances.md
```diff
@@ -3,17 +3,223 @@ It is possible to run multiple instances of AIO on one server.
...
+   --restart always \
+   --publish 8080:8080 \
+   --env APACHE_PORT=11000 \
+   --env APACHE_IP_BINDING=0.0.0.0 \
+   --env TALK_PORT=3478 \
+   --volume nextcloud_aio_mastercontainer:/mnt/docker-aio-config \
+   --volume /var/run/docker.sock:/var/run/docker.sock:ro \
+   nextcloud/all-in-one:latest
+   ```
+   The last command may take a few minutes. When it's finished, you should see a success message, saying "Initial startup of Nextcloud All-in-One complete!". Now exit the console session with `Ctrl + ]`. This concludes the setup for this particular VM.
```
szaimen, line 154:
```suggestion
   The last command may take a few minutes. When it's finished, you should see a success message, saying "Initial startup of Nextcloud All-in-One complete!". Now exit the console session with `Ctrl + [c]`. This concludes the setup for this particular VM.

```

szaimen (approved):
This looks great, thank you very much! :)

szaimen:
btw, @BraidenPsiuk I just invited you to the repo :)

multiple-instances.md
BraidenPsiuk: Hi, sorry I didn't see this until now- but that wasn't a typo, the escape sequence really is `ctrl + closing square bracket`. When you use the `virsh console` command to connect to an existing VM, it prints this helpful message:
```shell
# virsh console --domain my-virtual-machine
Connected to domain 'my-virtual-machine'
Escape character is ^] (Ctrl + ])
```
But when you first create the VM using the `virt-install` command, you never get to see this message telling you how to exit the console. So that's why I felt it's worth mentioning.

Also, it's kinda buried but there's a brief mention of this in the [virsh manpage](https://gitlab.com/libvirt/libvirt/-/blob/master/docs/manpages/virsh.rst?plain=1#L89), where they at least mention why they chose such a weird sequence :)

I'm still hoping to submit another PR in a month or so when I have more free time- to include proper config for getting Talk working with the VMs (and hopefully auto borgbackups via the AIO interface too). So if you want I can just sneak this in along with that PR.

szaimen: > I'm still hoping to submit another PR in a month or so when I have more free time- to include proper config for getting Talk working with the VMs (and hopefully auto borgbackups via the AIO interface too). So if you want I can just sneak this in along with that PR.

Yes, would be great. Thanks! :)

PR #4275, opened 2024-02-25
Update reverse-proxy.md

WaaromZoMoeilijk:
Add instructions how to use composerize.com to easily convert docker run commands to docker-compose files


Zoey2936 (changes requested):
we already provide a compose file, also composerize is very limited while converting

WaaromZoMoeilijk:
> we already provide a compose file, also composerize is very limited while converting

Not a very useful reply, elaborate why limited or change docs.

PR #4285, opened 2024-02-26
Add a clarifying comment to nginx.conf of reverse-proxy.md

apodl1:
In the example nginx.conf, if the user is running nextcloud on a subdomain of an already ssl secured domain, the <your-nc-domain> -placeholders should be replaced with eg. "sub.example.com" in server_name and just "example.com" in ssl-cert related lines. I added a comment clarifying this into the provided .conf

### DIFF
--- a/reverse-proxy.md
+++ b/reverse-proxy.md
@@ -357,6 +357,7 @@ server {
         proxy_set_header Connection $connection_upgrade;
     }
 
+    #if running nginx on a subdomain (eg. nextcloud.example.com) of a domain that already has an wildcard ssl certificate from certbot on this machine, the <your-nc-domain> in the below lines should be replaced with just the domain (eg. example.com), not the subdomain. In this case the subdomain should already be secured without additional actions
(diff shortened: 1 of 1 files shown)


Zoey2936:
should maybe be alos added to apache for example

reverse-proxy.md
```diff
@@ -357,6 +357,7 @@ server {
         proxy_set_header Connection $connection_upgrade;
     }
 
+    #if running nginx on a subdomain (eg. nextcloud.example.com) of a domain that already has an wildcard ssl certificate from certbot on this machine, the <your-nc-domain> in the below lines should be replaced with just the domain (eg. example.com), not the subdomain. In this case the subdomain should already be secured without additional actions
```
szaimen, line 360:
```suggestion
    # if running nginx on a subdomain (eg. nextcloud.example.com) of a domain that already has an wildcard ssl certificate from certbot on this machine, the <your-nc-domain> in the below lines should be replaced with just the domain (eg. example.com), not the subdomain. In this case the subdomain should already be secured without additional actions
```

szaimen (approved):

PR #4293, opened 2024-02-29
Fix and beautify link to Docker --init

rugk:
* The content has moved and is apparently now there (old link did not work):
* I use a link text instead of a plain link.


szaimen (approved):
LGTM, thanks!

PR #4363, opened 2024-03-11
Add private mirror support for OCI references #4362

elysweyr:
Implements #4362

### DIFF
--- a/nextcloud-aio-helm-chart/templates/nextcloud-aio-apache-deployment.yaml
+++ b/nextcloud-aio-helm-chart/templates/nextcloud-aio-apache-deployment.yaml
@@ -26,7 +26,11 @@ spec:
     spec:
       initContainers:
         - name: init-volumes
+          {{- if or .Values.IMAGE_MIRROR_PREFIX .Values.ALPINE_IMAGE_ORG }}
           image: "{{ .Values.IMAGE_MIRROR_PREFIX }}{{ .Values.ALPINE_IMAGE_ORG}}/alpine"
+          {{- else }}
+          image: alpine
+          {{- end }}
           command:
             - chmod
             - "777"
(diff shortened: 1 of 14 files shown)


nextcloud-aio-helm-chart/templates/nextcloud-aio-apache-deployment.yaml
```diff
@@ -50,7 +50,7 @@ spec:
             - name: COLLABORA_HOST
               value: nextcloud-aio-collabora
             - name: NC_DOMAIN
-              value: "{{ .Values.NC_DOMAIN }}"
+              value: "{{ .Values.NC_DOMAIN }}"FF
```
Zoey2936, line 53: is there a reason for this "FF"?

Zoey2936 (changes requested):
I think you need to edit this file: https://github.com/nextcloud/all-in-one/blob/main/nextcloud-aio-helm-chart/update-helm.sh because the helm-chart is generated using this script your changes will be removed again with the next run of this script

elysweyr:
> I think you need to edit this file: https://github.com/nextcloud/all-in-one/blob/main/nextcloud-aio-helm-chart/update-helm.sh because the helm-chart is generated using this script your changes will be removed again with the next run of this script

@Zoey2936 
Updated `update-helm.sh` and refined my control flow in case `ALPINE_IMAGE_ORG` and `IMAGE_MIRROR_PREFIX` are unset which would have lead to `/alpine` but if fixed now.

nextcloud-aio-helm-chart/templates/nextcloud-aio-apache-deployment.yaml
elysweyr: My bad

szaimen:
Hi, have you tested if this actually works? Because I doubt it. See https://github.com/nextcloud/all-in-one/discussions/4362#discussioncomment-8773102

szaimen:
Hi @elysweyr I just gave you write access to the repo and pushed your changes to https://github.com/nextcloud/all-in-one/pull/4387 so that all workflows work. Lets continue there.

PR #4386, opened 2024-03-14
nextcloud - update apps_paths to include absolute path

szaimen:
For https://github.com/nextcloud/server/issues/44157


come-nc (approved):
We are reverting upstream change: https://github.com/nextcloud/updater/pull/540
I still find it cleaner to not rely on OC class in config file.

PR #4443, opened 2024-03-25
docs(readme): update domain change procedure

### LINKED ISSUE #4335: Documented method for changing domain is incomplete
### Steps to reproduce
1. Install Nextcloud AIO container on a domain
2. Follow the [documented procedure](https://github.com/nextcloud/all-in-one?tab=readme-ov-file#how-to-change-the-domain) to change the domain
3. Try to use some apps

### Expected behavior 

All apps should use the new domain for any scenario

### Actual behavior 

Some apps (for example Talk) have more configuration to change, and use the old domain.

### Host OS 

Ubuntu server 23.10

#### Nextcloud AIO version 

28.0.3

#### Current channel 

Latest

#### Other valuable info 

Dumping the postgres database and grepping it returned ~12 results. Most are translations, some are from Talk, and a few I didn't discern where they're from

xyhhx:
Resolves #4335


szaimen:
Hi, thanks for you PR! However I fear the current state is unacceptable as we should not recommend to tinker with the database directly. Instead one should visit each setting and adjust the domain there.

szaimen:
See https://github.com/nextcloud/all-in-one/issues/4335#issuecomment-2020492022

PR #4495, opened 2024-04-04
nextcloud - adjust TALK_HOST logic

szaimen:


szaimen:
Merging for now as we still can improve later

PR #4513, opened 2024-04-07
feat: enable HEIC+TIFF without Imaginary

pulsejet:
Adds the required packages by default. Actually tiff is already present (at least right now) but good to be on the safe side.

1. Imaginary is not necessarily better. Actually slower for me.
2. This fails in a strange way: you can add the HEIC provider and it silently doesn't generate any previews.
3. There's no down side that I can think of

I've not tested the build but seems straightforward? Adding the `imagemagick-heic` package to a running container does fix the issue.

https://github.com/pulsejet/memories/issues/1120

### DIFF
--- a/Containers/nextcloud/Dockerfile
+++ b/Containers/nextcloud/Dockerfile
@@ -44,6 +44,8 @@ RUN set -ex; \
         icu-dev \
         imagemagick-dev \
         imagemagick-svg \
+        imagemagick-heic \
+        imagemagick-tiff \
         libevent-dev \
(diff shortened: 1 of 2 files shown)


szaimen (approved):
LGTM, thanks!

pulsejet:
Thanks for the quick turnaround!

szaimen:
This is now released with v8.2.0 Beta. Testing and feedback is welcome! See https://github.com/nextcloud/all-in-one#how-to-switch-the-channel

PR #4532, opened 2024-04-11
nextcloud - add logic for BRUTEFORCE_PROTECTION

szaimen:
For https://github.com/nextcloud/all-in-one/discussions/4525


st3iny (approved):
Not ideal but sometimes ...

PR #4555, opened 2024-04-17
Update readme.md - info about new web admin interface

kri164:
Add info about new web admin interface.


docjyJ:
Is it better to do a single PR with #4556?

docjyJ:
Merge into : https://github.com/nextcloud/all-in-one/pull/4556

PR #4560, opened 2024-04-17
trust IP of docker network gateway

### LINKED ISSUE #4528: Docker network gateway needs to be a trusted proxy in the config.php
### Steps to reproduce
1. Use AIO in reverse proxy mode (with NPMplus)
2. trigger failed login
3. see docker network gateway as IP

How to fix:
- add IP of docker network gateway to trusted_proxies in the config.php (should maybe be added to the entrypoint.sh/start.sh of the nextcloud container)

### Expected behavior 
log real ip

### Actual behavior 
logs IP of docker network gateway

### Host OS 
Debian 12.5

#### Nextcloud AIO version 
v8.1.0

#### Current channel 
beta

#### Other valuable info 
without IP of the docker gateway and then with it added to the trusted_proxies array in the config.php file
grafik

Zoey2936:
close #4528

### DIFF
--- a/Containers/nextcloud/entrypoint.sh
+++ b/Containers/nextcloud/entrypoint.sh
@@ -537,8 +537,10 @@ fi
 chmod 775 -R /var/www/html/custom_apps/notify_push/bin/
 php /var/www/html/occ config:system:set trusted_proxies 0 --value="127.0.0.1"
 php /var/www/html/occ config:system:set trusted_proxies 1 --value="::1"
+# IP of docker network gateway
+php /var/www/html/occ config:system:set trusted_proxies 2 --value="$(ip r | grep default | grep -o "[0-9]\+\.[0-9]\+\.[0-9]\+\.1")"
 if [ -n "$ADDITIONAL_TRUSTED_PROXY" ]; then
-    php /var/www/html/occ config:system:set trusted_proxies 2 --value="$ADDITIONAL_TRUSTED_PROXY"
+    php /var/www/html/occ config:system:set trusted_proxies 3 --value="$ADDITIONAL_TRUSTED_PROXY"
 fi
 if [ -n "$ADDITIONAL_TRUSTED_DOMAIN" ]; then
     php /var/www/html/occ config:system:set trusted_domains 2 --value="$ADDITIONAL_TRUSTED_DOMAIN"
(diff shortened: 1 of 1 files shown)


Containers/nextcloud/entrypoint.sh
```diff
@@ -537,8 +537,10 @@ fi
 chmod 775 -R /var/www/html/custom_apps/notify_push/bin/
 php /var/www/html/occ config:system:set trusted_proxies 0 --value="127.0.0.1"
 php /var/www/html/occ config:system:set trusted_proxies 1 --value="::1"
+# IP of docker network gateway
+php /var/www/html/occ config:system:set trusted_proxies 2 --value="$(ip r | grep default | grep -o "[0-9]\+\.[0-9]\+\.[0-9]\+\.1")"
```
szaimen, line 541: I fear this will overwrite the changes some people might have done already: https://github.com/nextcloud/all-in-one/discussions/2045
So we should probably choose another array key?

Containers/nextcloud/entrypoint.sh
```diff
@@ -537,8 +537,10 @@ fi
...
 if [ -n "$ADDITIONAL_TRUSTED_PROXY" ]; then
-    php /var/www/html/occ config:system:set trusted_proxies 2 --value="$ADDITIONAL_TRUSTED_PROXY"
+    php /var/www/html/occ config:system:set trusted_proxies 3 --value="$ADDITIONAL_TRUSTED_PROXY"
```
szaimen, line 543: same here?

Containers/nextcloud/entrypoint.sh
szaimen: Also should we maybe use the same logic like here? https://github.com/nextcloud/all-in-one/blob/17b35a45db08951e7f86e09a971563b407842cc9/Containers/apache/start.sh#L20-L24

Containers/nextcloud/entrypoint.sh
Zoey2936: What do you mean?

Containers/nextcloud/entrypoint.sh
Zoey2936: > Also should we maybe use the same logic like here?
> 
> https://github.com/nextcloud/all-in-one/blob/17b35a45db08951e7f86e09a971563b407842cc9/Containers/apache/start.sh#L20-L24

Seems to also work without

Containers/nextcloud/entrypoint.sh
szaimen: some people might have set the 3 key already...

szaimen:
This is now released with v8.2.1 Beta. Testing and feedback is welcome! See https://github.com/nextcloud/all-in-one#how-to-switch-the-channel

PR #4587, opened 2024-04-26
adjust the compose file to use `:` instead of `=`

szaimen:


Zoey2936 (approved):
Is there a reason for this?

szaimen:
> Is there a reason for this?

I think if we use the colon, quotation marks will be eliminated if they are used.

PR #4589, opened 2024-04-26
compose file: Correct the mix of two syntaxes in environment variables

Zakmaf:
Correcting the mix of two syntaxes in environment variables


szaimen (approved):
Thanks!

PR #4593, opened 2024-04-26
nginx config should also work for freenginx and openresty

Zoey2936:
freenginx is a fork of nginx
and openresty is nginx with some patches + some modules and lua support

### DIFF
--- a/reverse-proxy.md
+++ b/reverse-proxy.md
@@ -301,7 +301,7 @@ backend Nextcloud
 
 </details>
 
-### Nginx
+### freenginx/nginx/openresty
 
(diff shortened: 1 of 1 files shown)


reverse-proxy.md
```diff
@@ -301,7 +301,7 @@ backend Nextcloud
 
 </details>
 
-### Nginx
+### freenginx/nginx/openresty
```
szaimen, line 304: Maybe?
```suggestion
### Nginx, Freenginx, Openresty
```

szaimen (approved):

PR #4660, opened 2024-05-16
Update wording in aio-config.twig

ncguk:
Small improvements to text. No code changes.

### DIFF
--- a/php/templates/includes/aio-config.twig
+++ b/php/templates/includes/aio-config.twig
@@ -1,19 +1,19 @@
 <details>
     <summary>Click here to view the current AIO config and documentation links</summary><br />
     {% if was_start_button_clicked == true %}
-        Nextclouds config.php file is stored in the nextcloud_aio_nextcloud Docker volume and can be edited by following the <a href="https://github.com/nextcloud/all-in-one#how-to-edit-nextclouds-configphp-file-with-a-texteditor">config.php documentation</a>.<br><br>
+        Nextcloud's config.php file is stored in the nextcloud_aio_nextcloud Docker volume and can be edited by following the <a href="https://github.com/nextcloud/all-in-one#how-to-edit-nextclouds-configphp-file-with-a-texteditor">config.php documentation</a>.<br><br>
         You can run Nextcloud's usual occ commands by following the <a href="https://github.com/nextcloud/all-in-one#how-to-run-occ-commands">occ documentation</a></b>.<br><br>
     {% endif %}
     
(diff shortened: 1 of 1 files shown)


php/templates/includes/aio-config.twig
```diff
@@ -1,19 +1,19 @@
...
     <summary>Click here to view the current AIO config and documentation links</summary><br />
     {% if was_start_button_clicked == true %}
-        Nextclouds config.php file is stored in the nextcloud_aio_nextcloud Docker volume and can be edited by following the <a href="https://github.com/nextcloud/all-in-one#how-to-edit-nextclouds-configphp-file-with-a-texteditor">config.php documentation</a>.<br><br>
+        Nextcloud's config.php file is stored in the nextcloud_aio_nextcloud Docker volume and can be edited by following the <a href="https://github.com/nextcloud/all-in-one#how-to-edit-nextclouds-configphp-file-with-a-texteditor">config.php documentation</a>.<br><br>
         You can run Nextcloud's usual occ commands by following the <a href="https://github.com/nextcloud/all-in-one#how-to-run-occ-commands">occ documentation</a></b>.<br><br>
     {% endif %}
     
     {% if nextcloud_datadir starts with '/' %}
-        Nextcloud's datadir is getting stored in the {{ nextcloud_datadir }} directory.
+        Nextcloud's datadir is stored in the {{ nextcloud_datadir }} directory.
```
szaimen, line 9: actually, during the initial installation, the data is not stored yet. Only after finishing it. That is why this kind of wording was chosen. What would you suggest to improve this?

ncguk: Good point. I would suggest then that the wording be "Nextcloud's datadir will be stored in the {{ nextcloud_datadir }} directory."

szaimen: hm... I fear this doesnt work after the instance is installed then? (this template is used in both cases...)

ncguk: Hmm. I see what you mean.

"Nextcloud backups will be saved to {{ nextcloud_datadir }}." works for both a first back up and subsequent backups that have yet to be made. You could also have:

"All Nextcloud backups will be saved to {{ nextcloud_datadir }}."
"Your Nextcloud backups will be saved to {{ nextcloud_datadir }}."
"All of your Nextcloud backups will be saved to {{ nextcloud_datadir }}."
"Once you have made your first backup it, and all subsequent backups, will be stored in {{ nextcloud_datadir }}."

PR #4671, opened 2024-05-18
Add note to get correct source IP when using rootless docker

### LINKED ISSUE #4621: Containers only see IP of docker gateway when using nextcloud in docker rootless due to rootless "builtin" default port driver
### Steps to reproduce
1. Setup nextcloud aio with docker rootless as documented in [docker-rootless.md](https://github.com/nextcloud/all-in-one/blob/main/docker-rootless.md). Don't use an external reverse proxy.
2. Try logging in with an invalid password.
3. The IP logged in the admin logs is the IP of the docker network gateway, not the actual remote IP of the client.

### Expected behavior 
Correct IP logged.

### Actual behavior 
IP of the docker network gateway logged.

### Host OS 
linux

#### Nextcloud AIO version 
Nextcloud AIO v8.2.1

#### Current channel 
latest

#### Root cause
When using rootless docker defaults to using the "builtin" port driver.
See https://docs.docker.com/engine/security/rootless/#networking-errors.
But that port driver doesn't allow forwarding of the remote IP, instead containers see connections as coming from the IP of the docker network gateway (e.g. 172.19.0.1).
This means that nextclouds IP-based security measures don't work correctly and it is e.g. possible to do a denial of service attack against the nextcloud instance by spamming invalid logins and getting the IP of the gateway throtteled or even blocked.

### Fix

In order to fix this one can also use `slirp4netns` as a port driver.
* Give `slirp4netns` access to priviliged ports by adding `net.ipv4.ip_unprivileged_port_start=0` in `/etc/sysctl.conf`. (Only adding via netcap doesn't work, https://github.com/rootless-containers/slirp4netns/issues/251#issuecomment-761415404 says this is because rootlesskit drops the privs again before they could be used).
* Create `~/.config/systemd/user/docker.service.d/override.conf`
  with content
  ```
  [Service]
  Environment="DOCKERD_ROOTLESS_ROOTLESSKIT_PORT_DRIVER=slirp4netns"
  ```
* And then restart the daemon:
  ```
  $ systemctl --user daemon-reload 
  $ systemctl --user restart docker
  ```

Note that this will likely decrease performance.
See also: Following https://rootlesscontaine.rs/getting-started/docker/#changing-the-port-forwarder

### Update docs

I propose to updating the documentation for nextcloud in docker rootless.
Mention that if not using a non-dockerized reverse proxy the logged remote IPs will be that of the docker gateway and how to fix that.

ccaccb:
Fixes #4621


szaimen:
Hi, thanks for the PR! :)

Would you mind fixing DCO? We need this so that we can merge your commit. 

See 
image

szaimen:
@Kirk1984 does this change look good to you? :)

Kirk1984:
Sorry, i don't know. I have to confess that i ditched rootless a year+ ago for the time being. Some of my selfhosted services hat issues with it and from time to time the docker network died (around the 25+ container Mark). Never found the issue. I isolated my stuff in Proxmox VMs now.

szaimen:
> Sorry, i don't know. I have to confess that i ditched rootless a year+ ago for the time being. Some of my selfhosted services hat issues with it and from time to time the docker network died (around the 25+ container Mark). Never found the issue. I isolated my stuff in Proxmox VMs now.

I see. Thanks for the feedback!

szaimen (approved):
Thank you!

PR #4715, opened 2024-05-25
Add documentation not to use @ or : in passwords

### LINKED ISSUE #4567: Add information to not use @ or : in password (helm values.yaml)

I had multiple issues with "unique and good" passwords in the values.yaml when these contain special characters as @ or : as they are interpreted in postgresql URI.

Add a information in values.yaml above all the passwords to not use such symbols as @ and : in secrets.

- Escape those characters before building the PostgreSQL or redis connection URIs
- Provide a good error notice. Something better than "DATABASE_URL environment variable malformed" when I don't have to provide any DATABASE_URL env variable in values.yaml

yannik-b:
Fixes #4567


szaimen (approved):
LGTM, thanks!

PR #4737, opened 2024-05-30
Docs: Update manual-upgrade.md with alternative method

Doozy134:
Include an additional manual upgrade method for when documented way does not work (such was the case for me).

Amend/rephrase parts of the already documented method for clarity.


szaimen (approved):
Thanks!

PR #4742, opened 2024-05-31
facerecognition: change to FACE_MODEL=3 and adjust memory values

szaimen:
For https://github.com/nextcloud/all-in-one/discussions/4721#discussioncomment-9618685

### DIFF
--- a/community-containers/facerecognition/facerecognition.json
+++ b/community-containers/facerecognition/facerecognition.json
@@ -14,7 +14,7 @@
                 "FACE_MODEL=3"
             ],
             "aio_variables": [
-                "nextcloud_memory_limit=4096M"
+                "nextcloud_memory_limit=2048M"
             ],
             "nextcloud_exec_commands": [
                 "php /var/www/html/occ app:install facerecognition",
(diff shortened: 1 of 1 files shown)


szaimen:
@matiasdelellis can we also lower the limit in https://github.com/nextcloud/all-in-one/blob/b1d24962c237d8bd3cc9ece9cb20c9860ea41c54/community-containers/facerecognition/facerecognition.json#L16 and https://github.com/nextcloud/all-in-one/blob/b1d24962c237d8bd3cc9ece9cb20c9860ea41c54/community-containers/facerecognition/facerecognition.json#L24 then? Or does it need to stay as high due to https://github.com/nextcloud/all-in-one/blob/b1d24962c237d8bd3cc9ece9cb20c9860ea41c54/community-containers/facerecognition/facerecognition.json#L25?

matiasdelellis:
>  "nextcloud_memory_limit=4096M"

The absolute minimum is 512 MB that you have to set to allow it to work. but if the user has a lot of images (50 thousand?) they will probably need more memory to perform the face clustering step, therefore we must leave a reasonable value to start.

I will try to evaluate but maybe 2GB could be something very reasonable value. 

> "php /var/www/html/occ face:setup -M 4G", 

1G, it is a mere formality but we must assign it.

> "php /var/www/html/occ config:app:set facerecognition analysis_image_area --value 4320000",  

It's a lot, but it will work without problems.

matiasdelellis:
p.s: You can also increase the memory only for the commands and leave the default configuration.

> php -d  memory_limit=2048M

szaimen:
all right, adjusted :)

matiasdelellis (approved):

szaimen:
This is now released with v9.0.0 Beta. Testing and feedback is welcome! See https://github.com/nextcloud/all-in-one#how-to-switch-the-channel

PR #4755, opened 2024-06-03
revert "set NEXTCLOUD_MOUNT to rshared #2601"

szaimen:
as it causes too many problems.

See 
- https://github.com/nextcloud/all-in-one/discussions/4744
- https://github.com/nextcloud/all-in-one/discussions/4595
- https://github.com/nextcloud/all-in-one/discussions/3136
- and more


chatziko:
Hi, I just discovered that this was reverted in latest AIO, which breaks installs that rely on `rshared`. Is there any workaround to manually enable it?

PR #4796, opened 2024-06-08
Update reverse-proxy.md

foegra:
after Nextcloud 29 - new configuration has to be used for Caddyfile

### DIFF
--- a/reverse-proxy.md
+++ b/reverse-proxy.md
@@ -140,6 +140,10 @@ Add this to your Caddyfile:
 
 ```
 https://<your-nc-domain>:443 {
+    redir /.well-known/webfinger /index.php/.well-known/webfinger 301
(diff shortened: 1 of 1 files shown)


Zoey2936 (commented):
should be added to the AIO caddyfile, not the reverse proxy guide

szaimen:
> should be added to the AIO caddyfile, not the reverse proxy guide

Already done via https://github.com/nextcloud/all-in-one/pull/4774 and released with v9.0.0 Beta

foegra:
> should be added to the AIO caddyfile, not the reverse proxy guide

Where is that located?

foegra:
> should be added to the AIO caddyfile, not the reverse proxy guide

Big pardon - but why Caddyfile is included in the AIO installation?

PR #4800, opened 2024-06-09
Update reverse-proxy.md

ArmanTheParman:
exclude mac for network=host

### DIFF
--- a/reverse-proxy.md
+++ b/reverse-proxy.md
@@ -39,7 +39,7 @@ In order to run Nextcloud behind a web server or reverse proxy (like Apache, Ngi
 
     For this setup, you can use as target `host.docker.internal:$APACHE_PORT` instead of `localhost:$APACHE_PORT`. **⚠️ Important:** In order to make this work on Docker for Linux, you need to add `--add-host=host.docker.internal:host-gateway` to the docker run command of your reverse proxy container or `extra_hosts: ["host.docker.internal:host-gateway"]` in docker compose (it works on Docker Desktop by default).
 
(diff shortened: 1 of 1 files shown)


reverse-proxy.md
```diff
@@ -39,7 +39,7 @@ In order to run Nextcloud behind a web server or reverse proxy (like Apache, Ngi
...
 
-    Another option and actually the recommended way in this case is to use `--network host` option (or `network_mode: host` for docker-compose) as setting for the reverse proxy container to connect it to the host network. If you are using a firewall on the server, you need to open ports 80 and 443 for the reverse proxy manually. By doing so, the default sample configurations that point at `localhost:$APACHE_PORT` should work without having to modify them.
+    In this case, another option, and actually the recommended way, is to use `--network host` option (or `network_mode: host` for docker-compose) as setting for the reverse proxy container to connect it to the host network. Do note that this method is not available on MacOS. If you are using a firewall on the server, you need to open ports 80 and 443 for the reverse proxy manually. By doing so, the default sample configurations that point at `localhost:$APACHE_PORT` should work without having to modify them.
```
szaimen, line 42: maybe we should say that this is not supported on docker desktop in general?

szaimen (commented):
Thanks, just one comment

reverse-proxy.md
ArmanTheParman: your call, but why? it's only the MacOS version that doesn't support --nework=host.

very annoying, btw, has caused me much headache making my own software compatible for both Mac and Linux

szaimen: as far as I know it also does not work on windows and in general on all platforms that use docker deesktop...

Zoey2936: maybe see https://docs.docker.com/network/drivers/host/#docker-desktop

ArmanTheParman:
OK, I learned something new

szaimen:
Can be closed iirc

PR #4807, opened 2024-06-10
nextcloud: improve getting values for APPSTORAGE

szaimen:
Fix https://github.com/nextcloud/all-in-one/issues/4806

### DIFF
--- a/Containers/nextcloud/entrypoint.sh
+++ b/Containers/nextcloud/entrypoint.sh
@@ -162,8 +162,13 @@ if ! [ -f "$NEXTCLOUD_DATA_DIR/skip.update" ]; then
                 declare -Ag APPSTORAGE
                 echo "Disabling apps before the update in order to make the update procedure more safe. This can take a while..."
                 for app in "${NC_APPS_ARRAY[@]}"; do
-                    APPSTORAGE[$app]=$(php /var/www/html/occ config:app:get "$app" enabled)
-                    php /var/www/html/occ app:disable "$app"
+                    if php /var/www/html/occ config:app:get "$app" enabled >/dev/null; then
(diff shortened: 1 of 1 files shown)


Containers/nextcloud/entrypoint.sh
```diff
@@ -162,8 +162,13 @@ if ! [ -f "$NEXTCLOUD_DATA_DIR/skip.update" ]; then
...
                 for app in "${NC_APPS_ARRAY[@]}"; do
-                    APPSTORAGE[$app]=$(php /var/www/html/occ config:app:get "$app" enabled)
-                    php /var/www/html/occ app:disable "$app"
+                    if php /var/www/html/occ config:app:get "$app" enabled >/dev/null; then
+                        APPSTORAGE[$app]="$(php /var/www/html/occ config:app:get "$app" enabled)"
+                        php /var/www/html/occ app:disable "$app"
+                    else
+                        APPSTORAGE[$app]=""
+                        echo "Not disabling $app because the occ command to get the enabled state was failing."
+                    fi
```
joshtrichards, line 171: To eliminate running twice (and also make the redirect unnecessary) maybe:

```suggestion
                    if APPSTORAGE[$app]="$(php /var/www/html/occ config:app:get "$app" enabled)"; then
                        php /var/www/html/occ app:disable "$app"
                    else
                        APPSTORAGE[$app]=""
                        echo "Not disabling $app because the occ command to get the enabled state was failing."
                    fi
```

joshtrichards (approved):

Containers/nextcloud/entrypoint.sh
szaimen: Good point. Honestly, I personally dislike this syntax but of course this is more performant.

szaimen:
Thanks for the review Josh! :)

PR #4811, opened 2024-06-11
Stalwart: Change maintainer

docjyJ:
I see `%AIO_CHANNEL%`, is it working for AIO community containers ?

### DIFF
--- a/community-containers/stalwart/readme.md
+++ b/community-containers/stalwart/readme.md
@@ -4,7 +4,7 @@
 > The stability of Stalwart services is not guaranteed.
 > Do not use this feature as a main mail server without a redundancy system and without knowledge.
 > 
-> To learn or use as a secondary server enjoy it and please report bugs at [marcoambrosini/aio-stalwart](https://github.com/marcoambrosini/aio-stalwart/issues).
+> To learn or use as a secondary server enjoy it and please report bugs at [docjyj/aio-stalwart](https://github.com/docjyj/aio-stalwart/issues).
 
 ## Stalwart mail server
 This container bundles stalwart mail server and auto-configures it for you.
@@ -22,7 +22,7 @@ This container bundles stalwart mail server and auto-configures it for you.
 - See https://github.com/nextcloud/all-in-one/tree/main/community-containers#community-containers how to add it to the AIO stack
 
 ### Repository
-https://github.com/marcoambrosini/aio-stalwart
+https://github.com/docjyj/aio-stalwart
 
 ### Maintainer
-https://github.com/marcoambrosini
+https://github.com/docjyj
--- a/community-containers/stalwart/stalwart.json
+++ b/community-containers/stalwart/stalwart.json
@@ -4,7 +4,7 @@
             "container_name": "nextcloud-aio-stalwart",
             "display_name": "Stalwart",
             "documentation": "https://github.com/nextcloud/all-in-one/tree/main/community-containers/stalwart",
-            "image": "marcoambrosini/aio-stalwart",
+            "image": "ghcr.io/docjyj/aio-stalwart",
             "image_tag": "%AIO_CHANNEL%",
             "internal_port": "10003",
             "restart": "unless-stopped",
--- a/community-containers/stalwart/stalwart.json
+++ b/community-containers/stalwart/stalwart.json
@@ -4,7 +4,7 @@
             "container_name": "nextcloud-aio-stalwart",
             "display_name": "Stalwart",
             "documentation": "https://github.com/nextcloud/all-in-one/tree/main/community-containers/stalwart",
-            "image": "ghcr.io/docjyj/aio-stalwart",
+            "image": "docjyj/aio-stalwart",
             "image_tag": "%AIO_CHANNEL%",
             "internal_port": "10003",
             "restart": "unless-stopped",
--- a/community-containers/stalwart/stalwart.json
+++ b/community-containers/stalwart/stalwart.json
@@ -5,7 +5,7 @@
             "display_name": "Stalwart",
             "documentation": "https://github.com/nextcloud/all-in-one/tree/main/community-containers/stalwart",
             "image": "marcoambrosini/aio-stalwart",
-            "image_tag": "v2",
+            "image_tag": "%AIO_CHANNEL%",
             "internal_port": "10003",
             "restart": "unless-stopped",
             "ports": [


Zoey2936 (changes requested):
I think this would not work

szaimen:
> I think this would not work

Actually it would but it requires all "normal" aio tags also to be existing for https://hub.docker.com/r/marcoambrosini/aio-stalwart/tags. So at least `lastest`, `beta` and `dev`. Also you will sooner or later definitely run into this: https://github.com/nextcloud/all-in-one/discussions/4784#discussioncomment-9712152

docjyJ:
> Actually it would but it requires all "normal" aio tags also to be existing for https://hub.docker.com/r/marcoambrosini/aio-stalwart/tags. So at least `latest`, `beta` and `dev`. So you will at least need to mantain all these tags in parallel if you want to do so.

It's not necessary to maintain 3 tags. All 3 tags can point to the same container.

When upgrading, simply point the beta to the new container, and all beta users will receive the update without having to wait for an AIO beta.

And to release to all users, there's no need to be rushed by AIO beta deadlines.



> Also you will sooner or later definitely run into this: https://github.com/nextcloud/all-in-one/discussions/4784#discussioncomment-9712152

For changes requiring json modification, you'll always have to go through JSON modification...

It would be really nice to have a solution for free version testing.

szaimen:
> > Actually it would but it requires all "normal" aio tags also to be existing for https://hub.docker.com/r/marcoambrosini/aio-stalwart/tags. So at least `latest`, `beta` and `dev`. So you will at least need to mantain all these tags in parallel if you want to do so.
> 
> It's not necessary to maintain 3 tags. All 3 tags can point to the same container.
> 
> When upgrading, simply point the beta to the new container, and all beta users will receive the update without having to wait for an AIO beta.
> 
> And to release to all users, there's no need to be rushed by AIO beta deadlines.

I am fine with this change honestly but it gets more complicated for its maintainer imho...

> > Also you will sooner or later definitely run into this: [#4784 (reply in thread)](https://github.com/nextcloud/all-in-one/discussions/4784#discussioncomment-9712152)
> 
> For changes requiring json modification, you'll always have to go through JSON modification...
> 
> It would be really nice to have a solution for free version testing.

So what would be your suggestion?

docjyJ:
`V2.%AIO_CHANNEL%` should work.
In the case of complex changes (port, env...) adding a `VX` prefix would ensure that the correct volumes are mounted, etc.

I don't think it will be any more difficult.

With Github Repository and Github Actions, everything can be automated.

szaimen:
> `V2.%AIO_CHANNEL%` should work.

I fear this is going to break. See https://github.com/search?q=repo%3Anextcloud%2Fall-in-one%20AIO_CHANNEL&type=code. So in that case we would need to update the logic internally in order to allow for something like that.

docjyJ:
Exemple of workflow :
```yaml
name: Create and publish a Docker image

on:
  push:
    branches: 
      - 'latest'
      - 'beta'
      - 'dev'

env:
  aio_tag: v1-{{branch}}



jobs:
  build-and-push-image:
    name: Build and Push Docker image to GitHub Container Registry
    runs-on: ubuntu-latest

    permissions:
      contents: read
      packages: write
      attestations: write
      id-token: write

    steps:
      - name: Checkout repository
        uses: actions/checkout@v4

      - name: Extract metadata (tags, labels) for Docker
        id: meta -V1
        uses: docker/metadata-action@v5
        with:
          images: ghcr.io/${{ github.repository }}
          tags: ${{ aio_tag }}

      - name: Log in to the GitHub Container Registry
        uses: docker/login-action@v3
        with:
          registry: ghcr.io
          username: ${{ github.actor }}
          password: ${{ secrets.GITHUB_TOKEN }}

      - name: Extract metadata (tags, labels) for Docker
        id: meta
        uses: docker/metadata-action@v5
        with:

      - name: Build and push Docker images
        id: push
        uses: docker/build-push-action@v5
        with:
          context: .
          push: true
          tags: ${{ steps.meta.outputs.tags }}
          labels: ${{ steps.meta.outputs.labels }}
      
      - name: Generate artifact attestation
        uses: actions/attest-build-provenance@v1
        with:
          subject-name: ghcr.io/${{ github.repository }}
          subject-digest: ${{ steps.push.outputs.digest }}
          push-to-registry: true
```

Push to dev automatic update dev docker image.
Then merge dev into beta and beta into latest.

When change volume port and other breaking :
- Change, rename tag `v1-{{branch}}` to `v2-{{branch}}`.
- Propage dev - beta - latest.
- Update `v1-

It's can't be easier

docjyJ:
> > `V2.%AIO_CHANNEL%` should work.
> 
> I fear this is going to break. See https://github.com/search?q=repo%3Anextcloud%2Fall-in-one%20AIO_CHANNEL&type=code. So in that case we would need to update the logic internally in order to allow for something like that.

Yes I think it's doesn't work...

szaimen:
> > > `V2.%AIO_CHANNEL%` should work.
> > 
> > 
> > I fear this is going to break. See https://github.com/search?q=repo%3Anextcloud%2Fall-in-one%20AIO_CHANNEL&type=code. So in that case we would need to update the logic internally in order to allow for something like that.
> 
> Yes I think it's doesn't work...

so this would need to be adjusted first

docjyJ:
> > > > `V2.%AIO_CHANNEL%` should work.
> > > 
> > > 
> > > I fear this is going to break. See https://github.com/search?q=repo%3Anextcloud%2Fall-in-one%20AIO_CHANNEL&type=code. So in that case we would need to update the logic internally in order to allow for something like that.
> > 
> > Yes I think it's doesn't work...
> 
> so this would need to be adjusted first

If you think it's a good idea, I'll try to implement it.

szaimen:
> > > > > `V2.%AIO_CHANNEL%` should work.
> > > > 
> > > > 
> > > > I fear this is going to break. See https://github.com/search?q=repo%3Anextcloud%2Fall-in-one%20AIO_CHANNEL&type=code. So in that case we would need to update the logic internally in order to allow for something like that.
> > > 
> > > 
> > > Yes I think it's doesn't work...
> > 
> > 
> > so this would need to be adjusted first
> 
> If you think it's a good idea, I'll try to implement it.

I don't think it is necessary a good idea but I get the need from your side for such a feature. That is why I would be fine if you go ahead implementing this.

docjyJ:
image

I use all aio branch.

docjyJ:
I prefer to keep things as simple as possible, to avoid introducing bugs.

I will see about publishing on Docker Hub.

And I used the 3 channels, the Stalwart mail server is starting to become serious. Port and volume should not change.

For environment variables I can simply duplicate them (eg: if the MY_ENV variable has a new schema, I create MY_ENV_2 to ensure compatibility, and I possibly correct the rename in the script or config).

I think a lot of people would like a mail server associated with AIO. I would really like to provide stability. (And I would like to migrate my own config to integrate it into the backup)

szaimen (approved):
LGTM. Will be merged after the next stable release is out and then release with the next beta release :)

marcoambrosini (approved):

szaimen:

docjyJ:
Tested locally.
Admin login and WebUI work !

PR #4819, opened 2024-06-12
Allow GitHub Container Registry

docjyJ:


Zoey2936:
if I remember correctly, ghcr is very slow

docjyJ:
> if I remember correctly, ghcr is very slow

I don't know, but is this a problem?

I find that GitHub is more practical (no secret, only one account, integration with sources).

php/containers-schema.json
```diff
@@ -15,7 +15,7 @@
 					"image": {
 						"type": "string",
 						"minLength": 1,
-						"pattern": "^[a-z0-9/-]+$"
+						"pattern": "^(ghcr[.]io[/])?[a-z0-9-]+[/][a-z0-9-]+$"
```
szaimen, line 18: The reason that we dont allow ghcr.io is that we could not make the container update check work with ghcr.io https://github.com/nextcloud/all-in-one/blob/main/php/src/Docker/DockerHubManager.php

szaimen: If someone finds a solution then we can allow it :)

PR #4856, opened 2024-06-20
database-restore: only get the first match

szaimen:
Fix https://github.com/nextcloud/all-in-one/discussions/4849#discussioncomment-9828253


ralfrupf1976:
Thank you - do I understand right, I just have to wait for the fix - it will be posted here, right?

szaimen:
> Thank you - do I understand right, I just have to wait for the fix - it will be posted here, right?

I'll inform you once it is available in the beta channel. Probably next tuesday...

szaimen:

ralfrupf1976:
I just updated to version 9.1.0 and everything works fine again! Thank you

PR #4857, opened 2024-06-20
update alpine to v3.20.1

Zoey2936:

PR #4860, opened 2024-06-21
docker-socket-proxy: update haproxy to v3

szaimen:
See https://github.com/nextcloud/all-in-one/pull/4857#discussion_r1648752593


Zoey2936:
For me it still worked without changing my config file, but should be tested

szaimen:

PR #4912, opened 2024-06-29
Implement internal mode

docjyJ:
It allows you not to expose AIO servers when you use reverse proxy attach in the docker aio network.

It is useful for filtered all AIO services access (with Forward Auth and Community Container).
And it prevents the proxy from bypass.

Sample compose.yaml:
```yaml
services:
    init: true
    restart: always
    container_name: nextcloud-aio-mastercontainer
    volumes:
      - nextcloud_aio_mastercontainer:/mnt/docker-aio-config
      - /var/run/docker.sock:/var/run/docker.sock:ro
    ports:
      - 127.0.0.1:8080:8080
    environment:
      - APACHE_IP_BINDING=@INTERNAL
    networks:
      - nextcloud-aio

  caddy:
    image: caddy:alpine
    restart: always
    container_name: caddy
    volumes:
      - ./Caddyfile:/etc/caddy/Caddyfile
    ports:
      - 80:80
      - 433:433
    networks:
      - nextcloud-aio

volumes:
   name: nextcloud_aio_mastercontainer

networks:
  nextcloud-aio:
    name: nextcloud-aio
```

I do not think that services exposed on the host is a real issue, but in terms of security and isolation it is an appropriate improvement.

### DIFF
--- a/reverse-proxy.md
+++ b/reverse-proxy.md
@@ -8,6 +8,8 @@ A [reverse proxy](https://en.wikipedia.org/wiki/Reverse_proxy) is basically a we
 
 In order to run Nextcloud behind a web server or reverse proxy (like Apache, Nginx, Cloudflare Tunnel and else), you need to specify the port that AIO's Apache container shall use, add a specific config to your web server or reverse proxy and modify the startup command a bit. All examples below will use port `11000` as example `APACHE_PORT` which will be exposed on the host to receive unencrypted HTTP traffic from the reverse proxy. **Advice:** If you need https between Nextcloud and the reverse proxy because it is running on a different server in the same network, simply add another reverse proxy to the chain that runs on the same server like AIO and takes care of https proxying (most likely via self-signed cert). Another option is to create a VPN between the server that runs AIO and the server that runs the reverse proxy which takes care of encrypting the connection.
 
+If you are using a reverse proxy attached to the `nextcloud-aio` virtual network (like the [caddy community container](https://github.com/nextcloud/all-in-one/tree/main/community-containers/caddy)), you can set the `APACHE_IP_BINDING` to `@INTERNAL` to disable the exposure of the Apache container to the host network.
+
 **Attention:** The process to run Nextcloud behind a reverse proxy consists of at least steps 1, 2 and 4:
 1. **Configure the reverse proxy! See [point 1](#1-configure-the-reverse-proxy)**
 1. **Use this startup command! See [point 2](#2-use-this-startup-command)**
--- a/php/src/Docker/DockerActionManager.php
+++ b/php/src/Docker/DockerActionManager.php
@@ -137,7 +137,7 @@ class DockerActionManager
         } elseif($internalPort === '%TALK_PORT%') {
             $internalPort = $this->configurationManager->GetTalkPort();
         }
-        
+
         if ($internalPort !== "" && $internalPort !== 'host') {
             $connection = @fsockopen($containerName, (int)$internalPort, $errno, $errstr, 0.2);
             if ($connection) {
@@ -438,7 +438,7 @@ class DockerActionManager
         $requestBody['HostConfig']['RestartPolicy']['Name'] = $container->GetRestartPolicy();
 
         $requestBody['HostConfig']['ReadonlyRootfs'] = $container->GetReadOnlySetting();
- 
+
         $exposedPorts = [];
         if ($container->GetInternalPort() !== 'host') {
             foreach($container->GetPorts()->GetPorts() as $value) {
@@ -478,6 +478,10 @@ class DockerActionManager
                 $ipBinding = $value->ipBinding;
                 if ($ipBinding === '%APACHE_IP_BINDING%') {
                     $ipBinding = $this->configurationManager->GetApacheIPBinding();
+                    // Do not expose if AIO is in internal network mode
+                    if ($ipBinding === '@INTERNAL') {
+                        continue;
+                    }
                 }
                 $portWithProtocol = $port . '/' . $protocol;
                 $requestBody['HostConfig']['PortBindings'][$portWithProtocol] = [
@@ -708,7 +712,7 @@ class DockerActionManager
             if (!isset($imageOutput['RepoDigests'])) {
                 error_log('RepoDigests is not set of container ' . $containerName);
                 return null;
-            } 
+            }
 
             if (!is_array($imageOutput['RepoDigests'])) {
                 error_log('RepoDigests of ' . $containerName . ' is not an array which is not allowed!');


docjyJ:
> Also I think this is currently going to break here:
> 
> https://github.com/nextcloud/all-in-one/blob/3c8cb2bdb2b3d80190d729d3b1a73cff980bd42e/Containers/mastercontainer/start.sh#L182-L186

Fix

docjyJ:
According :
- https://github.com/nextcloud/all-in-one/blob/52cb90a80294588bcc7c78518084e073fbcbbb89/php/templates/containers.twig#L121
- https://github.com/nextcloud/all-in-one/blob/52cb90a80294588bcc7c78518084e073fbcbbb89/reverse-proxy.md#L738

If reverse proxy points correctly to `http://nextcloud-aio-apache:$APACHE_PORT/` from `https://exemple.com:443/`, it should work.

But, Caddy Community Container is set up after validation, so it will not work. Reverse proxy must be configured before validation.

szaimen:
> If reverse proxy points correctly to `http://nextcloud-aio-apache:$APACHE_PORT/` from `https://exemple.com:443/`, it should work.

No, it will not because it would need to point at `http://nextcloud-aio-domaincheck:$APACHE_PORT/` during the dlmain validation. However this makes the domain validation senseless because the container gets replaced after the validation by the apache container...




> But, Caddy Community Container is set up after validation, so it will not work. Reverse proxy must be configured before validation.

Yes, the caddy community works differently...

docjyJ:
What is the purpose of the domaine validations ?

matteoipri:
Hi all,
I'll add my experience to this thread.

On my Arch Linux server, I am running several docker compose stacks and one of them is NC AiO and another one is for HAProxy to serve as reverse proxy in front of all web services, including TLS termination.
Every stack has its own docker network and then I have an additional network created outside of docker compose to connect all frontends with HAProxy.
My HAProxy stack is also connected to the NC AiO docker network.

My HAProxy config has the following lines in the HTTPS frontend
```
  acl host_nc hdr_beg(host) -i nc
  use_backend nextcloud if host_nc
```
and the following backend
```
backend nextcloud
  server nextcloud-aio-apache nextcloud-aio-apache:11000 check init-addr none resolvers dockerdns
  server nextcloud-aio-domaincheck nextcloud-aio-domaincheck:11000 check init-addr none resolvers dockerdns
```

With this setup, I got AiO installed, including the domain verification, using HAProxy as a reverse proxy running on the same host in a different container.


I would say that things work, the only issue that I still encounter is the login rate limit, which I do not know if it is legitimate (I am under brute force attack) or if it is due to my setup.
```
Your remote address was identified as "fd12:3456:789a:2::a" and is brute-force throttled at the moment  
slowing down the performance of various requests. If the remote address is not your address this can be  
an indication that a proxy is not configured correctly. For more details see the documentation ↗.
```

Can my setup inspire a solution for this topic?
How can I fix the login throttling?

Thanks,
Matteo

docjyJ:
> Can my setup inspire a solution for this topic?
> How can I fix the login throttling?

I dont know how resolve this issue.

Disable https://apps.nextcloud.com/apps/bruteforcesettings fix the warning message.


I don't know if you've read it yet, but it might help:  
https://docs.nextcloud.com/server/latest/admin_manual/configuration_server/reverse_proxy_configuration.html

Static IP for docker containers : https://www.baeldung.com/ops/docker-assign-static-ip-container

docjyJ:
Sample (Not tested):
```yaml
services:
    init: true
    restart: always
    container_name: nextcloud-aio-mastercontainer
    volumes:
      - nextcloud_aio_mastercontainer:/mnt/docker-aio-config
      - /var/run/docker.sock:/var/run/docker.sock:ro
    ports:
      - 127.0.0.1:8080:8080
    environment:
      - APACHE_IP_BINDING=@INTERNAL
      - APACHE_PORT=80
    networks:
      - nextcloud-aio

  caddy:
    image: caddy:alpine
    restart: always
    container_name: my-caddy
    volumes:
      - ./Caddyfile:/etc/caddy/Caddyfile
    ports:
      - 80:80
      - 433:433
    networks:
      nextcloud-aio:
        ipv4_address: 10.9.8.7

volumes:
   name: nextcloud_aio_mastercontainer

networks:
  nextcloud-aio:
    name: nextcloud-aio
    driver: bridge
    ipam:
      config:
        - subnet: 10.9.8.0/24
          gateway: 10.9.8.1
```
```caddyfile
https://mycloud.com:443 {
 redir /.well-known/carddav /remote.php/dav/ 301
 redir /.well-known/caldav /remote.php/dav/ 301
  reverse_proxy {
    to http://nextcloud-aio-apache:80 http://nextcloud-aio-domaincheck:80
    lb_policy first
    health_uri /
    health_port 80
    health_interval 60s
  }
}
```
```php
$CONFIG = array (
  'trusted_proxies'   => ['10.9.8.7'],
  'overwritehost'     => 'my-caddy',
  'overwriteprotocol' => 'http'
);
```

ManOki:
@matteoipri I have a similar setup (swag instead of HAProxy) including an additional network just for frontends.

client -1-> swag -2-> caddy -3-> nextcloud

1 LAN/WAN
2 swag-network
3 nextcloud-aio-network

While caddy is automatically marked as a trusted proxy, this ain't the case for your HAProxy.
https://github.com/nextcloud/all-in-one/blob/main/Containers/nextcloud/entrypoint.sh#L563C73-L563C97
You have to setup ADDITIONAL_TRUSTED_PROXY environment variable for your nextcloud-aio-nextcloud container.

reverse-proxy.md
```diff
@@ -8,6 +8,8 @@ A [reverse proxy](https://en.wikipedia.org/wiki/Reverse_proxy) is basically a we
...
 
+If you are using a reverse proxy attached to the `nextcloud-aio` virtual network (like the [caddy community container](https://github.com/nextcloud/all-in-one/tree/main/community-containers/caddy)), you can set the `APACHE_IP_BINDING` to `@INTERNAL` to disable the exposure of the Apache container to the host network.
```
szaimen, line 11: this lacks the note to skip the domain validation in that case. Also I would probably rather add it somewhere further down...

docjyJ: Edited, is it better ?

szaimen:
So my problem with this in general is that the documentation is already very complicated because it offers many options. Adding yet another method will not make it easier to understand...

We can add this but lets not document this for now.

szaimen (approved):
LGTM. Thanks a lot @docjyJ!

reverse-proxy.md
szaimen: See https://github.com/nextcloud/all-in-one/pull/4912#issuecomment-2222790801

szaimen:
This is now released with v9.2.0 Beta. Testing and feedback is welcome! See https://github.com/nextcloud/all-in-one#how-to-switch-the-channel

docjyJ:
> This is now released with v9.2.0 Beta. Testing and feedback is welcome! See https://github.com/nextcloud/all-in-one#how-to-switch-the-channel

Work without domain validation

PR #4957, opened 2024-07-09
Update reverse-proxy.md

LinuxSpielKind:
Users running AIO on Ubuntu Jammy may face blocked web socket traffic trough Apache, since rewrites with "?" are deemed unsafe. A error like this can be found in the error logs: "[rewrite:error] [pid396674] [client 192.168.xxx.yyy:57444] AH: Unsafe URL with %3f URL rewritten without UnsafeAllow3F" The rewrite rule UnsafeAllow3F is fixing this issue. Infos on that flag can be found here: https://httpd.apache.org/docs/current/rewrite/flags.html#flag_unsafe_allow_3f


szaimen (approved):
Thanks!

Issue #4918, opened 2024-07-01
Cannot upload files larger than 100 MB via share link because of ClamAV


saytamkenorh:
### Steps to reproduce
1. Verify that logged in user can upload files larger than 100 MB through the Files app via the New button
2. Create a folder with a "File Drop" share link or a share link with modify rights
3. Try to upload a file larger than 100 MB using the link (it does not matter if the link is opened in the same browser window or in an anonymous window where a Nextcloud user is not logged in)

### Expected behavior 
File uploads without any problems.

### Actual behavior 
File does not upload. The user sees a message that an unknown error occurred.

**An error also appears in the Event Log:**

`BadRequest
Expected filesize of 325033371 bytes but read (from Nextcloud client) and wrote (to Nextcloud storage) 134111232 bytes. Could either be a network problem on the sending side or a problem writing to the storage on the server side.`

When I changed the files_antivirus av_stream_max_length setting via occ to the equivalent of 300 MB, the upload works fine but another error appears in the event log:
`Failed to write a chunk. Check if Stream Length matches StreamMaxLength in anti virus daemon settings.`
I should probably also change the settings in the /etc/clamav/clamd.conf file but this would require modifying the ClamAV container. **Nevertheless, it is certain that the antivirus somehow blocks the file upload.**

### Host OS 
Debian 12.5

#### Nextcloud AIO version 
v8.3.0

#### Current channel 
latest

#### Other valuable info 
Running behind a nginx reverse proxy

szaimen:
Hi, thanks for you report! I'll probably have a look next week.

szaimen:
Looks like I can reproduce this

saytamkenorh:
@szaimen Thank you for the fix!

szaimen:
This is now released with v9.2.0 Beta. Testing and feedback is welcome! See https://github.com/nextcloud/all-in-one#how-to-switch-the-channel

saytamkenorh:
@szaimen So I have installed the beta on my homelab server and I can confirm that with ClamAV enabled uploading files via a share link works now seamlessly out of the box. I was able to upload a ~1GB and a ~4GB file.

However, I found a warning in the ClamAV log:
`LibClamAV Warning: Max file-size was set to 10737418240 bytes. Unfortunately, scanning files greater than 2147483647 bytes (2 GiB - 1) is not supported.`
[...]

PR #4960, opened 2024-07-09
clamav - adjust max filesize conditionally

szaimen:
Fix https://github.com/nextcloud/all-in-one/issues/4918

### DIFF
--- a/Containers/clamav/start.sh
+++ b/Containers/clamav/start.sh
@@ -1,0 +1,11 @@
+#!/bin/bash
+
+# Adjust settings
+cat /etc/clamav/clamd.conf > /tmp/clamd.conf
+CLAMAV_FILE="$(sed "s|10G|$MAX_SIZE|" /clamav.conf)"
+echo "$CLAMAV_FILE" >> /tmp/clamd.conf
+
+# Call initial init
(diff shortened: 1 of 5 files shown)


Containers/clamav/start.sh
```diff
@@ -0,0 +1,11 @@
+#!/bin/bash
+
+if ! grep -q "AIO settings" /etc/clamav/clamd.conf; then
+    CADDYFILE="$(sed "s|10G|$UPLOAD_LIMIT|" /clamav.conf)"
+    echo "$CADDYFILE" >> /etc/clamav/clamd.conf
+fi
+
+# Call initial init
+/init
```
Zoey2936, line 9: Why not exec /init?

Containers/clamav/start.sh
```diff
@@ -0,0 +1,11 @@
+#!/bin/bash
+
+if ! grep -q "AIO settings" /etc/clamav/clamd.conf; then
+    CLAMAV_FILE="$(sed "s|10G|$MAX_SIZE|" /clamav.conf)"
+    echo "$CLAMAV_FILE" >> /etc/clamav/clamd.conf
```
szaimen, line 5: now the problem is that this file is read-only

PR #4996, opened 2024-07-17
update-helm: adjust workflow for new tags syntax

szaimen:
Needs https://github.com/nextcloud-releases/all-in-one/pull/48


st3iny (approved):
Looks good

PR #5062, opened 2024-07-29
docker-socket-proxy: move installation of app_api to containers.json …

szaimen:
…and nextcloud_exec_commands

### DIFF
--- a/Containers/nextcloud/entrypoint.sh
+++ b/Containers/nextcloud/entrypoint.sh
@@ -815,22 +815,5 @@ else
     fi
 fi
 
-# Docker socket proxy
-if version_greater "$installed_version" "27.1.2.0"; then
-    if [ "$DOCKER_SOCKET_PROXY_ENABLED" = 'yes' ]; then
-        if ! [ -d "/var/www/html/custom_apps/app_api" ]; then
-            php /var/www/html/occ app:install app_api
-        elif [ "$(php /var/www/html/occ config:app:get app_api enabled)" != "yes" ]; then
-            php /var/www/html/occ app:enable app_api
-        elif [ "$SKIP_UPDATE" != 1 ]; then
-            php /var/www/html/occ app:update app_api
-        fi
-    else
-        if [ "$REMOVE_DISABLED_APPS" = yes ] && [ -d "/var/www/html/custom_apps/app_api" ]; then
-            php /var/www/html/occ app:remove app_api
-        fi
-    fi
-fi
-
 # Remove the update skip file always
 rm -f "$NEXTCLOUD_DATA_DIR"/skip.update
--- a/php/containers.json
+++ b/php/containers.json
@@ -745,6 +745,11 @@
       ],
       "cap_drop": [
         "NET_RAW"
+      ],
+      "nextcloud_exec_commands": [
+        "echo 'Activating App API...'",
+        "php /var/www/html/occ app:install app_api",
+        "php /var/www/html/occ app:enable app_api"
       ]
     }
   ]


Containers/nextcloud/entrypoint.sh
```diff
@@ -815,22 +815,5 @@ else
...
-        if ! [ -d "/var/www/html/custom_apps/app_api" ]; then
-            php /var/www/html/occ app:install app_api
-        elif [ "$(php /var/www/html/occ config:app:get app_api enabled)" != "yes" ]; then
-            php /var/www/html/occ app:enable app_api
-        elif [ "$SKIP_UPDATE" != 1 ]; then
-            php /var/www/html/occ app:update app_api
-        fi
-    else
-        if [ "$REMOVE_DISABLED_APPS" = yes ] && [ -d "/var/www/html/custom_apps/app_api" ]; then
-            php /var/www/html/occ app:remove app_api
```
docjyJ, line 830: I don't think this is a problem, but the update and delete functionality is not offered by nextcloud_exec_commands.

docjyJ: With the addition of this option, it would be possible to manage the required applications more precisely.
`containers-schema.json`
```json
"nextcloud_applications": {
  "type": "array",
  "items": {
    "type": "string",
    "pattern": "^\\w+$"
  }
},
```
If you think it might be good, I can add installation and activation with this option.

szaimen: > ... and delete functionality is not offered by nextcloud_exec_commands.

ah good point. that is why I didnt use it for that in the past. Thanks for reminding me :)

> If you think it might be good, I can add installation and activation with this option.

I think this is a big overengineered for now but thanks for the idea :)

szaimen:
btw @docjyJ if you are looking for things to do or improve regarding the CCs, I have this list of things I think would be useful: https://github.com/nextcloud/all-in-one/discussions/3889

docjyJ:
Thanks!

I'm working on Stalwart and the 9.0 update. I just deployed my AIO instance on the RK3588, and I'll be migrating to Stalwart soon.

Maybe add a way to edit the default configuration (eg: use your own certificate, use another subdomain...).

It would be nice to have a perfect solution for beginners but powerful for advanced users (at their own risk).

docjyJ:
Also do you have some time to take a look at the aio-stalwart PRs?

It's good to have feedback, especially from someone with experience.

szaimen:
> I'm working on Stalwart and the 9.0 update. I just deployed my AIO instance on the RK3588, and I'll be migrating to Stalwart soon.

all right, cool

> Maybe add a way to edit the default configuration (eg: use your own certificate, use another subdomain...).

This could be done with a CC like the caddy container but will not be added for the AIOs default caddy one, mostly due to maintainability concerns.

> It would be nice to have a perfect solution for beginners but powerful for advanced users (at their own risk).

This is fine with CCs but not for the default stack. Since adding options will also compllicate things for beginners.

> Also do you have some time to take a look at the aio-stalwart PRs?
> 
> It's good to have feedback, especially from someone with experience.

What exactly do you want me to look into there?

docjyJ:
> This is fine with CCs but not for the default stack. Since adding options will also compllicate things for beginners.

I was thinking about a file to rewrite default rules. Especially to use a custom proxy (instead of Caddy CC) or use a custom domain. This is the only thing that the container manages with admin IDs and logs.

> What exactly do you want me to look into there?

Just to review these features. By asking for review via github feature. If you agree.

szaimen:
> I was thinking about a file to rewrite default rules. Especially to use a custom proxy (instead of Caddy CC) or use a custom domain. This is the only thing that the container manages with admin IDs and logs.

Hm... can you describe your use case a bit more in detail?

> Just to review these features. By asking for review via github feature. If you agree.

I can probably review changes on github that you do but I cannot test updates for stalwart unfortunately since my test instance does not allow to test stalwart because I am unfortunately not in control over the dns confguration that the server uses...

docjyJ:
Here example of my idea : https://github.com/docjyJ/aio-stalwart/blob/dev/manager/start.sh

Users can decide to disable some logics if they prefer to manage by themselves.

Some use cases I see:
- Use own certificates.
- Disable admin account or change passwords  <==== important
- Use POP without IMAP
- Use an alternative directory (ldap for example) <==== important
- Change database format

In the future, it could allow creating a Nextcloud / Stalwart user synchronization by simply activating a line in a file.

szaimen:
I see, looks interesting! Do you have an example how a user would adjust these settings? Going into the container and then adjusting the settings in a file for example?

docjyJ:
It would be in a file `/opt/stalwart-mail/config/nc-aio-config.toml`:

```toml
skip-secure-check = false
manage.binding.mail = true
manage.binding.caddy-web = true
manage.storage.data = true
manage.storage.directory = true
manage.log.file = true
manage.log.stdout = true
manage.admin = true
manage.certificate = true
```

szaimen:
Looks cool! I mean if you want to add and support this, in your stalwart container feel free to! :)

szaimen:
Anotjer option would be to add an env file that gets sourced inside start.sh into the volume. Then you dont need a custom logic for getting and setting env. Wdyt about this idea?

docjyJ:
Yes, maybe better

PR #5078, opened 2024-08-01
daily-backup: fix issue with APACHE_PORT

szaimen:
Regressions from https://github.com/nextcloud/all-in-one/pull/4912

Fix https://help.nextcloud.com/t/aio-daily-backup-does-not-start-finish/199774

### DIFF
--- a/Containers/mastercontainer/daily-backup.sh
+++ b/Containers/mastercontainer/daily-backup.sh
@@ -16,11 +16,15 @@ fi
 sudo -u www-data touch "/mnt/docker-aio-config/data/daily_backup_running"
 
 # Check if apache is running/stopped, watchtower is stopped and backupcontainer is stopped
-APACHE_PORT="$(docker inspect nextcloud-aio-apache --format "{{.HostConfig.PortBindings}}" | grep -o '[0-9]\+' | head -1)"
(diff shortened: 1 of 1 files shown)


Containers/mastercontainer/daily-backup.sh
```diff
@@ -16,11 +16,15 @@ fi
...
 
 # Check if apache is running/stopped, watchtower is stopped and backupcontainer is stopped
-APACHE_PORT="$(docker inspect nextcloud-aio-apache --format "{{.HostConfig.PortBindings}}" | grep -o '[0-9]\+' | head -1)"
-while docker ps --format "{{.Names}}" | grep -q "^nextcloud-aio-apache$" && ! nc -z nextcloud-aio-apache "$APACHE_PORT"; do
-    echo "Waiting for apache to become available"
-    sleep 30
-done
+APACHE_PORT="$(docker inspect nextcloud-aio-apache --format "{{.Config.Env}}" | grep -o 'APACHE_PORT=[0-9]\+' | grep -o '[0-9]\+' | head -1)"
+if [ -z "$APACHE_PORT" ]; then
+    echo "APACHE_PORT is not set which is not expected..."
```
devnoname120, line 21: IMO this is confusing because the error claims that `APACHE_PORT` is not defined when it could actually just be that the format is incorrect.

PR #5082, opened 2024-08-02
Update for stalwart 0.9

docjyJ:
See https://github.com/stalwartlabs/mail-server/releases/tag/v0.9.0


szaimen:
Hi @docjyJ thanks but I am wondering if it would make more sense to move this upgrade documentation to somewhere inside your own repo so that you dont need to update the upgrade docs here all the time? Does this make sense?

docjyJ:
Yes

PR #5089, opened 2024-08-03
readme: fix mistakes in Cloudflare section

devnoname120:
- Cloudflare Tunnel requires the proxy option to be enabled. The documentation made it sound like you had the option to either disable Cloudflare Proxy **or** Cloudflare Tunnel to fix it. But this is an illusory choice because disabling the former also disables the latter.
- Cloudflare Proxy and Cloudflare Tunnel both cause the traffic to be mitm'ed by Cloudflare. It's an important privacy concern that may undermine altogether the point of even self-hosting Nextcloud 
- TLS termination at Cloudflare doesn't degrade performances a lot in my experience. I have stellar performances with Cloudflare Tunnel both in terms of latency and speed. I rephrased the remark to tone it down a bit but we may want to consider removing it altogether unless/until we have benchmarks proving it.


szaimen (approved):
Fine by me, thanks!

PR #5092, opened 2024-08-04
Added reverse proxy configuration example for IIS with ARR & Url Rewrite

sockless-coding:
Information on how to configure IIS with ARR and URL Rewrite as a reverse proxy.

### DIFF
--- a/reverse-proxy.md
+++ b/reverse-proxy.md
@@ -639,6 +639,63 @@ The examples below define the dynamic configuration in YAML files. If you rather
 
 </details>
 
+### IIS with ARR and URL Rewrite
+
+<details>
+
+<summary>click here to expand</summary>
+
+**Disclaimer:** It might be possible that the config below is not working 100% correctly, yet. Improvements to it are very welcome!
+
+#### Prerequisites
+1. **Windows Server** with IIS installed.
+2. [**Application Request Routing (ARR)**](https://www.iis.net/downloads/microsoft/application-request-routing) and [**URL Rewrite**](https://www.iis.net/downloads/microsoft/url-rewrite) modules installed.
+3. [**WebSocket Protocol**](https://learn.microsoft.com/en-us/iis/configuration/system.webserver/websocket) feature enabled.
+
+
+In the web.config example below, a Server farm named `nc-server-farm` has been created an is pointing to the Nextcloud server:
+
+```xml
+<?xml version="1.0" encoding="UTF-8"?>
+<configuration>
+  <system.web>
+    <!-- Allow all urls -->
+    <httpRuntime requestValidationMode="2.0" requestPathInvalidCharacters="" />
+  </system.web>
+  <system.webServer>
+    <rewrite>
+      <rules>
+        <!-- Force https -->
+        <rule name="Https" stopProcessing="true">
+          <match url="(.*)" />
+          <conditions>
+            <add input="{HTTPS}" pattern="^OFF$" />
+          </conditions>
+          <action type="Redirect" url="https://{HTTP_HOST}/{REQUEST_URI}" appendQueryString="false" />
+        </rule>
+        <!-- Redirect to internal nextcloud server -->
+        <rule name="To nextcloud" stopProcessing="true">
+          <match url="(.*)" />
+          <conditions>
+            <add input="{HTTPS}" pattern="^ON$" />
+          </conditions>
+          <action type="Rewrite" url="http://nc-server-farm:11000/{UNENCODED_URL}" appendQueryString="false" />
+        </rule>
+      </rules>
+    </rewrite>
+    <security>
+      <!-- Increase upload limit to 2GiB -->
+      <requestFiltering allowDoubleEscaping="true">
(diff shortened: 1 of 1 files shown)


reverse-proxy.md
```diff
@@ -639,6 +639,63 @@ The examples below define the dynamic configuration in YAML files. If you rather
...
+          <conditions>
+            <add input="{HTTPS}" pattern="^OFF$" />
+          </conditions>
+          <action type="Redirect" url="https://{HTTP_HOST}/{REQUEST_URI}" appendQueryString="false" />
+        </rule>
+        <!-- Redirect to internal nextcloud server -->
+        <rule name="To nextcloud" stopProcessing="true">
+          <match url="(.*)" />
+          <conditions>
+            <add input="{HTTPS}" pattern="^ON$" />
```
szaimen, line 680: so all queries will be forwarded to nextcloud not only if your given domain is queried?

reverse-proxy.md
```diff
@@ -639,6 +639,63 @@ The examples below define the dynamic configuration in YAML files. If you rather
...
+          </conditions>
+          <action type="Rewrite" url="http://nc-server-farm:11000/{UNENCODED_URL}" appendQueryString="false" />
+        </rule>
+      </rules>
+    </rewrite>
+    <security>
+      <!-- Increase upload limit to 2GiB -->
+      <requestFiltering allowDoubleEscaping="true">
+        <requestLimits maxAllowedContentLength="2147483648" />
```
szaimen, line 689: is it possible to set this to 0 so unlimited? all other rp docs do this...

szaimen (commented):
Thanks, a few remarks:

1. Where is the certificate added? I don't see it in the config...

reverse-proxy.md
sockless-coding: What hosts and certificates are used is handled elsewhere, this config is located on the site level.
I was thinking about adding a whole guide on it, but I thought if you are planing on using IIS as a reverse proxy you probably have some basic understanding of it.
FYI: https://learn.microsoft.com/en-us/iis/manage/configuring-security/how-to-set-up-ssl-on-iis
https://medium.com/@Evelyn.Taylor/how-to-setup-hostname-to-website-in-local-iis-ac3fee333526

The main point of adding this is that the normal way of creating rewrite action would look like this:
`<action type="Rewrite" url="http://nc-server-farm:11000/{R:1}" appendQueryString="false" />`
But that breaks Collabora due to how the /cool/ request is formatted, so `{UNENCODED_URL}` is required instead of the `{R:1}`

I created a post about on the forum
https://help.nextcloud.com/t/document-loading-failed-and-bad-uri-syntax-logged-in-collabora-container/200067

reverse-proxy.md
sockless-coding: It can be increased to 4GB but to turn it off it has to be done on the server level, and that is not recommended (it handles more things).

reverse-proxy.md
```diff
@@ -639,6 +639,63 @@ The examples below define the dynamic configuration in YAML files. If you rather
...
+      </requestFiltering>
+    </security>
+  </system.webServer>
+</configuration>
+
+```
+
```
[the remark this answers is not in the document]
sockless-coding: Done

reverse-proxy.md
szaimen: > I was thinking about adding a whole guide on it 

I guess that would be nice! Feel free to create it here: https://github.com/nextcloud/all-in-one/discussions/new?category=wiki

> What hosts and certificates are used is handled elsewhere, this config is located on the site level.
I was thinking about adding a whole guide on it, but I thought if you are planing on using IIS as a reverse proxy you probably have some basic understanding of it.

I see. However is the above config then complete enough to actually make it work?

reverse-proxy.md
szaimen: Hm... I guess there should be a link to how to turn it off as otherwise uploads bigger than 4GB via public share link will always fail which is not expected...

reverse-proxy.md
sockless-coding: I made the instructions a little clearer, the provided example web.config is the one I'm using, and it's working fine.

reverse-proxy.md
sockless-coding: I added a "please note" describing the limitations of using IIS as the reverse proxy.
On a side note, wouldn't it be more efficient to perform the uploads in segments... sending large files in one single request could cause all kind of troubles if the request gets interrupted.

szaimen: > On a side note, wouldn't it be more efficient to perform the uploads in segments... sending large files in one single request could cause all kind of troubles if the request gets interrupted.

You are right but this is currently a limitation of Nextcloud: https://github.com/nextcloud/server/issues/4109

reverse-proxy.md
szaimen: LGTM

szaimen (approved):
LGTM, thanks!

PR #5112, opened 2024-08-08
Document AIO_URL

LukasPietzschmann:
`AIO_URL` is part of the Nextcloud container: https://github.com/nextcloud/all-in-one/blob/41e6d7cf6d2330e695b26ada5eca2a25634d91b9/Containers/nextcloud/Dockerfile#L12
When running the AIO interface, e.g., behind a reverse proxy, one can change this variable to the proxied URL.

### DIFF
--- a/compose.yaml
+++ b/compose.yaml
@@ -13,6 +13,7 @@ services:
       - 8080:8080
       - 8443:8443 # Can be removed when running behind a web server or reverse proxy (like Apache, Nginx, Caddy, Cloudflare Tunnel and else). See https://github.com/nextcloud/all-in-one/blob/main/reverse-proxy.md
     # environment: # Is needed when using any of the options below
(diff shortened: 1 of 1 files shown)


LukasPietzschmann:
I don't want to bother you, but I set the variable in my compose file and when I run `docker exec nextcloud-aio-nextcloud env | grep AIO_URL`, the variable is indeed set. The "Open Nextcloud AIO Interface" button in Nextcloud's admin overview also links to the URL set in `AIO_URL`.

docjyJ:
Hi!

The URL is automatically updated to the last domain name used.

If you access the aio via `admin.example.con` normally the variable is updated to that domain.
No need to use the environment variable, the master container does not read the `AIO_URL` variable.

You don't bother, Thanks! ❤️

PR #5119, opened 2024-08-11
apache: re-enable content-length header

szaimen:
- See https://github.com/nextcloud/server/issues/47017
- See https://help.nextcloud.com/t/no-indication-of-download-progress-or-eta-when-donwloading-big-files-from-nextcloud-aio/200261

### DIFF
--- a/Containers/apache/nextcloud.conf
+++ b/Containers/apache/nextcloud.conf
@@ -51,4 +51,7 @@ Listen 8000
 
     # See https://httpd.apache.org/docs/trunk/mod/core.html#traceenable
     TraceEnable Off
+
+    # See https://github.com/nextcloud/server/issues/47017#issuecomment-2267102392
+    SetEnv ap_trust_cgilike_cl
 </VirtualHost>


Containers/apache/nextcloud.conf
```diff
@@ -51,4 +51,7 @@ Listen 8000
 
     # See https://httpd.apache.org/docs/trunk/mod/core.html#traceenable
     TraceEnable Off
+
+    # See https://github.com/nextcloud/server/issues/47017#issuecomment-2267102392
+    SetEnv ap_trust_cgilike_cl
```
szaimen, line 56:
```suggestion
    SetEnvIf Request_URI "remote\.php\/dav\/files\/.*" ap_trust_cgilike_cl
```

kesselb: The content-length header is also set in our app framework (e.g. https://github.com/nextcloud/server/blob/b34edf222419df41dac091a4a7b4e41814f5a7c8/lib/public/AppFramework/Http/FileDisplayResponse.php#L50). It might still work to limit if to remote.php for now.

szaimen:
```suggestion
    SetEnvIf Request_URI "remote\.php\/.*" ap_trust_cgilike_cl
```

PR #5166, opened 2024-08-24
helm-chart: Allow users to change apache service type

yannik-b:
I am using nextcloud-aio via the Helm chart in a Kubernetes environment. As I am using Nextcloud behind a kubernetes ingress controller, there is no need for my nextcloud-aio-apache to be a LoadBalancer service. As I have no Load Balancer IP addresses free, nextcloud-aio-apache service is pending (and so is the `helm install`) until I manually change the type to `ClusterIP`.

Therefore I introduced a helm variable to change the apache service type.


docjyJ:
LGTM

szaimen:
Hi, this helm chart is not intended to run behind an ingress and we also don't have plans to make it officially supported. 
However feel free to apply this one-line change to your instance manually.

PR #5203, opened 2024-09-04
aio-interface: allow to install nc 30 directly

szaimen:


szaimen:
Todo: update link to docs

PR #5306, opened 2024-09-21
Update containers.json to fix inter-container-connectivity

qudiqudi:
correct FULLTEXTSEARCH_HOST resolution to get the fulltextsearch container connect to the right network and the nextcloud container, fixes https://github.com/nextcloud/fulltextsearch/issues/831


szaimen:
Hi, just to be clear, this is a bug in alpines curl implementation: https://gitlab.alpinelinux.org/alpine/aports/-/issues/15690

szaimen:
However, since the change seems simple enough, I am fine to include it here for the time being.

szaimen (approved):

PR #5307, opened 2024-09-21
Ench/noid/heathcheck

docjyJ:
- Add Unhealty state
- Replace class by [backed enum](https://www.php.net/manual/en/language.enumerations.backed.php)

### DIFF
--- a/php/README.md
+++ b/php/README.md
@@ -33,6 +33,7 @@ docker run \
 --rm \
 --name nextcloud-aio-mastercontainer \
 --volume nextcloud_aio_mastercontainer:/mnt/docker-aio-config \
+--volume /var/run/docker.sock:/var/run/docker.sock \
 nextcloud/all-in-one:latest
 ```
 
--- a/php/src/Container/Container.php
+++ b/php/src/Container/Container.php
@@ -2,7 +2,6 @@
 
 namespace AIO\Container;
 
-use AIO\Container\State\IContainerState;
 use AIO\Data\ConfigurationManager;
 use AIO\Docker\DockerActionManager;
 use AIO\ContainerDefinitionFetcher;
@@ -170,7 +169,7 @@ class Container {
         return $this->dockerActionManager->GetContainerRestartingState($this);
     }
 
-    public function GetUpdateState() : IContainerState {
+    public function GetUpdateState() : bool {
         return $this->dockerActionManager->GetContainerUpdateState($this);
     }
 
--- a/php/src/Container/IContainerState.php
+++ b/php/src/Container/IContainerState.php
@@ -1,0 +1,14 @@
+<?php
+
+namespace AIO\Container;
+
+enum IContainerState
+{
+    case ImageDoesNotExist;
+    case NotRestarting;
+    case Restarting;
+    case Running;
+    case Starting;
+    case Stopped;
+    case Unhealthy;
+}
--- a/php/src/Container/IContainerState.php
+++ b/php/src/Container/IContainerState.php
@@ -2,13 +2,13 @@
 
 namespace AIO\Container;
 
-enum IContainerState
+enum IContainerState: string
 {
-    case ImageDoesNotExist;
-    case NotRestarting;
-    case Restarting;
-    case Running;
-    case Starting;
-    case Stopped;
-    case Unhealthy;
+    case ImageDoesNotExist = 'image_does_not_exist';
+    case NotRestarting = 'not_restarting';
+    case Restarting = 'restarting';
+    case Running = 'running';
+    case Starting = 'starting';
+    case Stopped = 'stopped';
+    case Unhealthy = 'unhealthy';
 }
--- a/php/src/Controller/DockerController.php
+++ b/php/src/Controller/DockerController.php
@@ -2,7 +2,7 @@
 
 namespace AIO\Controller;
 
-use AIO\Container\State\RunningState;
+use AIO\Container\IContainerState;
 use AIO\ContainerDefinitionFetcher;
 use AIO\Docker\DockerActionManager;
 use Psr\Http\Message\ResponseInterface as Response;
@@ -35,7 +35,7 @@ class DockerController
 
         // Don't start if container is already running
         // This is expected to happen if a container is defined in depends_on of multiple containers
-        if ($container->GetRunningState() instanceof RunningState) {
+        if ($container->GetRunningState() === IContainerState::Running) {
             error_log('Not starting ' . $id . ' because it was already started.');
             return;
         }
@@ -48,7 +48,7 @@ class DockerController
             }
         }
 
-        // Check if docker hub is reachable in order to make sure that we do not try to pull an image if it is down 
+        // Check if docker hub is reachable in order to make sure that we do not try to pull an image if it is down
         // and try to mitigate issues that are arising due to that
         if ($pullImage) {
             if (!$this->dockerActionManager->isDockerHubReachable($container)) {
@@ -261,10 +261,10 @@ class DockerController
         $domaincheckContainer = $this->containerDefinitionFetcher->GetContainerById($id);
         $apacheContainer = $this->containerDefinitionFetcher->GetContainerById(self::TOP_CONTAINER);
         // Don't start if apache is already running
-        if ($apacheContainer->GetRunningState() instanceof RunningState) {
+        if ($apacheContainer->GetRunningState() ===IContainerState::Running) {
             return;
         // Don't start if domaincheck is already running
-        } elseif ($domaincheckContainer->GetRunningState() instanceof RunningState) {
+        } elseif ($domaincheckContainer->GetRunningState() === IContainerState::Running) {
             $domaincheckWasStarted = apcu_fetch($cacheKey);
             // Start domaincheck again when 10 minutes are over by not returning here
             if($domaincheckWasStarted !== false && is_string($domaincheckWasStarted)) {
--- a/php/src/Docker/DockerActionManager.php
+++ b/php/src/Docker/DockerActionManager.php
@@ -3,15 +3,7 @@
 namespace AIO\Docker;
 
 use AIO\Container\Container;
-use AIO\Container\State\IContainerState;
-use AIO\Container\State\ImageDoesNotExistState;
-use AIO\Container\State\StartingState;
-use AIO\Container\State\RunningState;
-use AIO\Container\State\RestartingState;
-use AIO\Container\State\NotRestartingState;
-use AIO\Container\State\VersionDifferentState;
-use AIO\Container\State\StoppedState;
-use AIO\Container\State\VersionEqualState;
+use AIO\Container\IContainerState;
 use AIO\Data\ConfigurationManager;
 use GuzzleHttp\Exception\RequestException;
 use AIO\ContainerDefinitionFetcher;
@@ -62,17 +54,21 @@ class DockerActionManager
             $response = $this->guzzleClient->get($url);
         } catch (RequestException $e) {
             if ($e->getCode() === 404) {
-                return new ImageDoesNotExistState();
+                return IContainerState::ImageDoesNotExist;
             }
             throw $e;
         }
 
         $responseBody = json_decode((string)$response->getBody(), true);
 
-        if ($responseBody['State']['Running'] === true) {
-            return new RunningState();
-        } else {
-            return new StoppedState();
+        if ($responseBody['State']['Running'] !== true) {
+            return IContainerState::Stopped;
+        }
+        else if(array_key_exists('Health', $responseBody['State']) && $responseBody['State']['Health']['Status'] !== 'healthy') {
+            return IContainerState::Unhealthy;
+        }
+        else {
+            return IContainerState::Running;
         }
     }
 
@@ -83,7 +79,7 @@ class DockerActionManager
             $response = $this->guzzleClient->get($url);
         } catch (RequestException $e) {
             if ($e->getCode() === 404) {
-                return new ImageDoesNotExistState();
+                return IContainerState::ImageDoesNotExist;
             }
             throw $e;
         }
@@ -91,13 +87,13 @@ class DockerActionManager
         $responseBody = json_decode((string)$response->getBody(), true);
 
         if ($responseBody['State']['Restarting'] === true) {
-            return new RestartingState();
+            return IContainerState::Restarting;
         } else {
-            return new NotRestartingState();
+            return IContainerState::NotRestarting;
         }
     }
 
-    public function GetContainerUpdateState(Container $container) : IContainerState
+    public function GetContainerUpdateState(Container $container) : bool
     {
         $tag = $container->GetImageTag();
         if ($tag === '%AIO_CHANNEL%') {
@@ -106,28 +102,26 @@ class DockerActionManager
 
         $runningDigests = $this->GetRepoDigestsOfContainer($container->GetIdentifier());
         if ($runningDigests === null) {
-            return new VersionDifferentState();
+            return false;
         }
         $remoteDigest = $this->dockerHubManager->GetLatestDigestOfTag($container->GetContainerName(), $tag);
         if ($remoteDigest === null) {
-            return new VersionEqualstate();
+            return true;
         }
 
         foreach($runningDigests as $runningDigest) {
             if ($runningDigest === $remoteDigest) {
-                return new VersionEqualState();
+                return true;
             }
         }
-        return new VersionDifferentState();
+        return false;
     }
 
     public function GetContainerStartingState(Container $container) : IContainerState
     {
         $runningState = $this->GetContainerRunningState($container);
-        if ($runningState instanceof StoppedState) {
-            return new StoppedState();
-        } elseif ($runningState instanceof ImageDoesNotExistState) {
-            return new ImageDoesNotExistState();
+        if ($runningState === IContainerState::Stopped || $runningState === IContainerState::ImageDoesNotExist) {
+            return $runningState;
         }
 
         $containerName = $container->GetIdentifier();
@@ -142,12 +136,12 @@ class DockerActionManager
             $connection = @fsockopen($containerName, (int)$internalPort, $errno, $errstr, 0.2);
             if ($connection) {
                 fclose($connection);
-                return new RunningState();
+                return $runningState;
             } else {
-                return new StartingState();
+                return IContainerState::Starting;
             }
         } else {
-            return new RunningState();
+            return $runningState;
         }
     }
 
@@ -641,7 +635,7 @@ class DockerActionManager
         $container = $this->containerDefinitionFetcher->GetContainerById($id);
 
         $updateAvailable = "";
-        if ($container->GetUpdateState() instanceof VersionDifferentState) {
+        if ($container->GetUpdateState() === false) {
             $updateAvailable = '1';
         }
         foreach ($container->GetDependsOn() as $dependency) {
@@ -802,7 +796,7 @@ class DockerActionManager
 
     public function sendNotification(Container $container, string $subject, string $message, string $file = '/notify.sh') : void
     {
-        if ($this->GetContainerStartingState($container) instanceof RunningState) {
+        if ($this->GetContainerStartingState($container) === IContainerState::Running) {
 
             $containerName = $container->GetIdentifier();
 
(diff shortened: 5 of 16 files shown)


php/src/Container/Container.php
```diff
@@ -170,7 +169,7 @@ public function GetRestartingState() : IContainerState {
         return $this->dockerActionManager->GetContainerRestartingState($this);
     }
 
-    public function GetUpdateState() : IContainerState {
+    public function GetUpdateState() : bool {
```
st3iny, line 172: Was this changed on purpose?

php/src/Controller/DockerController.php
```diff
@@ -261,10 +261,10 @@ public function StartDomaincheckContainer() : void
         $domaincheckContainer = $this->containerDefinitionFetcher->GetContainerById($id);
         $apacheContainer = $this->containerDefinitionFetcher->GetContainerById(self::TOP_CONTAINER);
         // Don't start if apache is already running
-        if ($apacheContainer->GetRunningState() instanceof RunningState) {
+        if ($apacheContainer->GetRunningState() ===IContainerState::Running) {
```
st3iny, line 264:
```suggestion
        if ($apacheContainer->GetRunningState() === IContainerState::Running) {
```

php/src/Docker/DockerActionManager.php
```diff
@@ -62,17 +54,21 @@ public function GetContainerRunningState(Container $container) : IContainerState
...
         $responseBody = json_decode((string)$response->getBody(), true);
 
-        if ($responseBody['State']['Running'] === true) {
-            return new RunningState();
-        } else {
-            return new StoppedState();
+        if ($responseBody['State']['Running'] !== true) {
+            return IContainerState::Stopped;
+        }
+        else if(array_key_exists('Health', $responseBody['State']) && $responseBody['State']['Health']['Status'] !== 'healthy') {
```
st3iny, line 67:
```suggestion
        } elseif (array_key_exists('Health', $responseBody['State']) && $responseBody['State']['Health']['Status'] !== 'healthy') {
```

In php `else if` is not the same as `elseif`.

php/src/Docker/DockerActionManager.php
```diff
@@ -62,17 +54,21 @@ public function GetContainerRunningState(Container $container) : IContainerState
...
+            return IContainerState::Unhealthy;
+        }
+        else {
```
st3iny, line 70:
```suggestion
        } else {
```

php/src/Docker/DockerActionManager.php
```diff
@@ -83,21 +79,21 @@ public function GetContainerRestartingState(Container $container) : IContainerSt
...
-            return new RestartingState();
+            return IContainerState::Restarting;
         } else {
-            return new NotRestartingState();
+            return IContainerState::NotRestarting;
         }
     }
 
-    public function GetContainerUpdateState(Container $container) : IContainerState
+    public function GetContainerUpdateState(Container $container) : bool
```
st3iny, line 96: I liked the old, more descriptive version better. A boolean return type could be obscure and easy to misinterpret.

Perhaps you could add another enum or add the two extra states to the existing ones?

php/src/Docker/DockerActionManager.php
```diff
@@ -641,7 +635,7 @@ private function isContainerUpdateAvailable(string $id) : string
         $container = $this->containerDefinitionFetcher->GetContainerById($id);
 
         $updateAvailable = "";
-        if ($container->GetUpdateState() instanceof VersionDifferentState) {
+        if ($container->GetUpdateState() === false) {
```
st3iny, line 638: See my other comments above.

php/README.md
```diff
@@ -33,6 +33,7 @@ docker run \
 --rm \
 --name nextcloud-aio-mastercontainer \
 --volume nextcloud_aio_mastercontainer:/mnt/docker-aio-config \
+--volume /var/run/docker.sock:/var/run/docker.sock \
```
st3iny, line 36: Is this change related?

php/src/Container/IContainerState.php
```diff
@@ -0,0 +1,14 @@
+<?php
+
+namespace AIO\Container;
+
+enum IContainerState: string
```
st3iny, line 5:
```suggestion
enum ContainerState: string
```

The `I` prefix is usually reserved for interfaces and it can be dropped in my opinion.

st3iny (commented):
Hey [Jean-Yves](https://github.com/docjyJ), Simon asked me to have a look at your changes. I left some feedback regarding the code itself.

I really like you refactoring the container state logic from an interface to an enum. However, I think that both concerns could have been split into separate PRs. One PR to refactor the interface and another one to implement the new healthy states. But the code is already there so why not.

Looks good otherwise but did not test.

docjyJ:
Hi,

Thanks for the answer, as soon as I have time I'll look into it. 

I will try to separate and make several PR.

Take care,

szaimen:
Conflicts :/

docjyJ:
Don't worry, i'll handle it.

docjyJ:
Solved and up to date and ready (to be tested anyway...)

docjyJ:
I have podman on my machine and I can't launch the container... I can't test it and I don't have time to debug...

szaimen:
> I have podman on my machine and I can't launch the container...

Why may I ask? Do you run into an issue here?

docjyJ:
> Why may I ask? Do you run into an issue here?

```shell
podman run \
--rm \
--name nextcloud-aio-mastercontainer \
--volume nextcloud_aio_mastercontainer:/mnt/docker-aio-config:z \
--volume /var/run/user/1000/podman/podman.sock:/var/run/docker.sock:z \
--env WATCHTOWER_DOCKER_SOCKET_PATH=/var/run/user/1000/podman/podman.sock \
nextcloud/all-in-one:latest
```
```
Trying to fix docker.sock permissions internally...
Adding internal www-data to group root
Cannot connect to the docker socket. Cannot proceed.
Did you maybe remove group read permissions for the docker socket? AIO needs them in order to access the docker socket.
If SELinux is enabled on your host, see https://github.com/nextcloud/all-in-one#are-there-known-problems-when-selinux-is-enabled
If you are on TrueNas SCALE, see https://github.com/nextcloud/all-in-one#can-i-run-aio-on-truenas-scale
```
I don't know what happen

szaimen:
> > Why may I ask? Do you run into an issue here?
> 
> ```shell
> podman run \
> --rm \
> --name nextcloud-aio-mastercontainer \
> --volume nextcloud_aio_mastercontainer:/mnt/docker-aio-config:z \
> --volume /var/run/user/1000/podman/podman.sock:/var/run/docker.sock:z \
> --env WATCHTOWER_DOCKER_SOCKET_PATH=/var/run/user/1000/podman/podman.sock \
> nextcloud/all-in-one:latest
> ```
> 
> ```
> Trying to fix docker.sock permissions internally...
> Adding internal www-data to group root
> Cannot connect to the docker socket. Cannot proceed.
> Did you maybe remove group read permissions for the docker socket? AIO needs them in order to access the docker socket.
> If SELinux is enabled on your host, see https://github.com/nextcloud/all-in-one#are-there-known-problems-when-selinux-is-enabled
> If you are on TrueNas SCALE, see https://github.com/nextcloud/all-in-one#can-i-run-aio-on-truenas-scale
> ```
> 
> I don't know what happen

Maybe this helps? https://github.com/nextcloud/all-in-one?tab=readme-ov-file#are-there-known-problems-when-selinux-is-enabled

docjyJ:
image
:roll_eyes:

docjyJ:
It doesn't work, I give up.
Can you test this?
One way to test would be to go into the Nextcloud container and add `exit 1` in the `heathcheck.sh` script.

According to the docker documentation: https://docs.docker.com/reference/api/engine/version/v1.45/#tag/Container/operation/ContainerInspect

State - Health - Status
> Status is one of none, starting, healthy or unhealthy
> 
>     "none" Indicates there is no healthcheck
>     "starting" Starting indicates that the container is not yet ready
>     "healthy" Healthy indicates that the container is running correctly
>     "unhealthy" Unhealthy indicates that the container has a problem

Adding this information would be good for detecting potential problems and for debugging.

docjyJ:
Another question Why not use `starting` state and `restarting` state from docker ?
>  Enum: "created" "running" "paused" "restarting" "removing" "exited" "dead"
> 
> String representation of the container state. Can be one of "created", "running", "paused", "restarting", "removing", "exited", or "dead".

Could we use this indicator?
It would be easier to maintain.
If you agree, I can try to rewrite the functions `GetContainerRestartingState`, `GetContainerRunningState`, `GetContainerRestartingState`.

docjyJ:
Docker doc for meaning of state https://docs.docker.com/reference/cli/docker/container/ls/#status

docjyJ:
A proposal in commit b903cc3.

(I renamed the enums to match the getters name, it's more consistent but I can rollback)

szaimen:
> Another question Why not use `starting` state and `restarting` state from docker ?
> 
> > Enum: "created" "running" "paused" "restarting" "removing" "exited" "dead"
> > String representation of the container state. Can be one of "created", "running", "paused", "restarting", "removing", "exited", or "dead".
> 
> Could we use this indicator? It would be easier to maintain. If you agree, I can try to rewrite the functions `GetContainerRestartingState`, `GetContainerRunningState`, `GetContainerRestartingState`.

I overread this. Yes sounds good imo :)

php/src/Container/Container.php
```diff
@@ -112,22 +113,15 @@ public function GetVolumes() : ContainerVolumes {
...
+    public function GetRunningState() : RunningState {
         return $this->dockerActionManager->GetContainerRunningState($this);
     }
 
-    public function GetRestartingState() : ContainerState {
-        return $this->dockerActionManager->GetContainerRestartingState($this);
-    }
-
-    public function GetUpdateState() : VersionState {
+    public function GetUpdateState() : UpdateState {
```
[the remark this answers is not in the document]
docjyJ: I understand, it will be less explicit, but it is not important.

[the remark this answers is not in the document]
docjyJ: I understand, it will be less explicit, but it is not important.

php/src/Docker/DockerActionManager.php
```diff
@@ -778,9 +743,8 @@ public function IsMastercontainerUpdateAvailable() : bool
         return true;
     }
 
-    public function sendNotification(Container $container, string $subject, string $message, string $file = '/notify.sh') : void
-    {
-        if ($this->GetContainerStartingState($container) === ContainerState::Running) {
+    public function sendNotification(Container $container, string $subject, string $message, string $file = '/notify.sh'): void {
+        if ($this->GetContainerRunningState($container) === ContainerState::Running) {
```
st3iny, line 747:
```suggestion
        if ($this->GetContainerState($container) === ContainerState::Running) {
```

See above.

php/src/Docker/DockerActionManager.php
```diff
@@ -35,98 +37,62 @@ private function BuildImageName(Container $container) : string {
         return $container->GetContainerName() . ':' . $tag;
     }
 
-    public function GetContainerRunningState(Container $container) : ContainerState
-    {
+    /** @throws GuzzleException */
+    public function GetContainerRunningState(Container $container): ContainerState {
```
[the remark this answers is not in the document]
st3iny: Please also check the other usages I missed during my review.

php/templates/containers.twig
```diff
@@ -40,19 +40,20 @@
...
+            {% set runingState = container.GetRunningState().value %}
+            {% if container.GetDisplayName() != '' and runingState in ['running', 'unhealthy', 'starting'] %}
                 {% set isAnyRunning = true %}
             {% endif %}
-            {% if container.GetDisplayName() != '' and container.GetRestartingState().value == 'restarting' %}
+            {% if container.GetDisplayName() != '' and runingState == 'restarting' %}
                 {% set isAnyRestarting = true %}
             {% endif %}
-            {% if container.GetIdentifier() == 'nextcloud-aio-watchtower' and container.GetRunningState().value == 'running' %}
+            {% if container.GetIdentifier() == 'nextcloud-aio-watchtower' and runingState in ['running', 'unhealthy', 'starting'] %}
```
st3iny, line 50:
```suggestion
            {% if container.GetIdentifier() == 'nextcloud-aio-watchtower' and container.GetRunningState().isNotStopped() %}
```

st3iny (commented):
Another round of reviews and some feedback for you.

docjyJ:
> Another round of reviews and some feedback for you.

Thank, resolved

st3iny (approved):
Awesome! Looks good to me now!

docjyJ:
Health checks may need to be adjusted, probably reduce the interval during startup to speed up startup of all containers.

> *start period* provides initialization time for containers that need time to bootstrap. Probe failure during that period will not be counted towards the maximum number of retries. However, if a health check succeeds during the start period, the container is considered started and all consecutive failures will be counted towards the maximum number of retries.
>
> *start interval* is the time between health checks during the start period. This option requires Docker Engine version 25.0 or later.

See: https://docs.docker.com/reference/dockerfile/#healthcheck

docjyJ:
This should allow for better dependency management.
Can be used for helm chart, I guess.

szaimen:
Thanks for the idea! However our health checks are currently built in a way that they never fail after a specific time. See https://github.com/nextcloud/all-in-one/blob/main/Containers/apache/healthcheck.sh

docjyJ:
> Thanks for the idea! However our health checks are currently built in a way that they never fail after a specific time. See for example https://github.com/nextcloud/all-in-one/blob/main/Containers/apache/healthcheck.sh. for the rest the defaults are good enough imho

The problem with doing this is that docker considers the container ready... This should immediately take the container out of the starting state.

php/src/Docker/DockerActionManager.php
```diff
@@ -35,98 +38,57 @@ private function BuildImageName(Container $container) : string {
...
-        $internalPort = $container->GetInternalPort();
-        if($internalPort === '%APACHE_PORT%') {
-            $internalPort = $this->configurationManager->GetApachePort();
-        } elseif($internalPort === '%TALK_PORT%') {
-            $internalPort = $this->configurationManager->GetTalkPort();
+            return UpdateState::Latest;
         }
 
-        if ($internalPort !== "" && $internalPort !== 'host') {
-            $connection = @fsockopen($containerName, (int)$internalPort, $errno, $errstr, 0.2);
```
szaimen, line 120: @docjyJ we need to keep this. (I oversaw this)

szaimen (changes requested):
Blocking, see above

szaimen:
This is the only reliable way to check if the container is really up and running.

docjyJ:
A fix would be to fail the container if nextcloud:9000 is not reachable and add a startup period (e.g. 10 minutes)
```diff
#!/bin/bash

- nc -z "$NEXTCLOUD_HOST" 9000 || exit 0
+ nc -z "$NEXTCLOUD_HOST" 9000 || exit 0
nc -z 127.0.0.1 8000 || exit 1
nc -z 127.0.0.1 "$APACHE_PORT" || exit 1
if ! nc -z "$NC_DOMAIN" 443; then
    echo "Could not reach $NC_DOMAIN on port 443."
    exit 1
fi
```

```diff
- HEALTHCHECK CMD /healthcheck.sh
+ HEALTHCHECK --start-period=10m CMD /healthcheck.sh
```

szaimen:
> A fix would be to fail the container if nextcloud:9000 is not reachable and add a startup period (e.g. 10 minutes)

This is the problem: we cannot spwcify a time here as is depending on the overall setup like installed apps and features, amount of users, given hardware and else

docjyJ:
> This is the problem: we cannot spwcify a time here as is depending on the overall setup like installed apps and features, amount of users, given hardware and else especially during upgrades

Yes, I see...

Would it be better to manage dependencies like docker compose?

> The solution for detecting the ready state of a service is to use the condition attribute with one of the following options:

> -  service_started
> - service_healthy. This specifies that a dependency is expected to be “healthy”, which is defined with healthcheck, before starting a dependent service.

https://docs.docker.com/compose/how-tos/startup-order/

szaimen:
I fear this is going to have the same problem afaics, no?

php/src/Docker/DockerActionManager.php
docjyJ: I put the connection test back. A container marked healthy but not reachable will be in the starting state

szaimen:
@docjyJ LGTM now 😊

Can you please rebase the PR and squash the commits? 
Afterwards we can merge this :)

docjyJ:
Done

php/src/Container/ContainerState.php
```diff
@@ -2,11 +2,35 @@
 
 namespace AIO\Container;
 
-enum ContainerState: string {
-    case ImageDoesNotExist = 'image_does_not_exist';
-    case NotRestarting = 'not_restarting';
-    case Restarting = 'restarting';
-    case Running = 'running';
```
szaimen, line 9: Sorry, I think I oversaw this. I think we need the Running state as otherwise the logic gets hard to understand, no?

php/src/Docker/DockerActionManager.php
```diff
@@ -35,98 +38,74 @@ private function BuildImageName(Container $container) : string {
...
-        if($internalPort === '%APACHE_PORT%') {
-            $internalPort = $this->configurationManager->GetApachePort();
-        } elseif($internalPort === '%TALK_PORT%') {
-            $internalPort = $this->configurationManager->GetTalkPort();
+        $body = json_decode($response->getBody()->getContents(), true);
+        assert(is_array($body));
+        assert(is_array($body['State']));
+
+        $state = match ($body['State']['Status']) {
+            'running' => ContainerState::Healthy,
```
szaimen, line 76: For example this should be ContainerState::Running, no?

php/src/Container/ContainerState.php
```diff
@@ -2,11 +2,35 @@
...
+        return $this == self::Healthy;
+    }
+
+    public function isUnhealthy(): bool {
+        return $this == self::Unhealthy;
+    }
+
+    public function isRunning(): bool {
+        return $this->isHealthy() || $this->isUnhealthy() || $this->isStarting() || $this->isRestarting();
+    }
```
docjyJ, line 35: The Running state corresponds to the old GetRunningContainerState state.
Maybe there is another clearer state. Healthy means that the container is running without any problem detected by the AIO.

php/src/Controller/DockerController.php
```diff
@@ -28,7 +30,7 @@ private function PerformRecursiveContainerStart(string $id, bool $pullImage = tr
 
         // Don't start if container is already running
         // This is expected to happen if a container is defined in depends_on of multiple containers
-        if ($container->GetRunningState() === ContainerState::Running) {
+        if ($container->GetContainerState()->isRunning()) {
```
docjyJ, line 33: container->GetRunningState does not include Starting and Restarting

php/src/Docker/DockerActionManager.php
```diff
@@ -35,98 +38,74 @@ private function BuildImageName(Container $container) : string {
...
-            throw $e;
-        }
-
-        $responseBody = json_decode((string)$response->getBody(), true);
-
-        if ($responseBody['State']['Running'] === true) {
-            return ContainerState::Running;
-        } else {
-            return ContainerState::Stopped;
-        }
```
docjyJ, line 56: This is true regardless of the Healthy or Starting or Running state.

php/src/Controller/DockerController.php
docjyJ: To avoid it being hard to understand I went through the isSomething function

php/src/Container/ContainerState.php
docjyJ: See my comment

php/src/Docker/DockerActionManager.php
docjyJ: See my comment

szaimen (changes requested):
@docjyJ I discussed the latest changes with @st3iny and we agree that the latest changes do not make much sense or at least are not easy enough to merge them. I am very sorry for that. 

I would like to ask you if you could restore the changes from https://github.com/nextcloud/all-in-one/commit/b20e2a8201619803e8066e4f36a2cdde7cf72b2d by running `git checkout <hash>` and then push the old changes to a new branch via so: `git checkout -b ench/noid/heathcheck-restored`. From that commit hash, I think only such a check like https://github.com/nextcloud/all-in-one/pull/5307/files#diff-502ad3fbc5a2714763795f78cf314d4a76ada0cb3746530f2fb86fa471dd897bR91-R105 is missing. (best in a new commit)

Can you please do the above? I would do it myself but I unfortunately don't have your changes locally available anymore

PR #5308, opened 2024-09-21
Clean code, resolve psalm type error

docjyJ:
Hey

When you are running `composer run psalm -- --show-info=true` Psalm report a lot of type error.

The goal of this PR is to fix these errors through type checking.

It is always more pleasant to develop without waring everywhere.

What do you think? I find it quite relaxing to clean up code. If you think it's a good idea, I can take care of that task.


szaimen:
Hi @docjyJ, this is a very nice idea! Thanks for starting to work on this. I think for reviews it would be best if you could create separate PRs, one for each file if possible or at least grouped in some other way. Thanks a lot! :)

docjyJ:
Ok, so you can review this PR

docjyJ:
See #5369

PR #5314, opened 2024-09-23
helm: adjust initcontainer: use fixed alpine tag

szaimen:
- [x] wait for next major release of aio since this is potentially breaking


st3iny (approved):
Looks good

PR #5323, opened 2024-09-25
FTS: increase ES_JAVA_OPTS

szaimen:
For https://github.com/nextcloud/all-in-one/discussions/5311


Zoey2936:
Then we should also increase the minimum ram requirement in readme and dashboard

szaimen:
> Then we should also increase the minimum ram requirement in readme and dashboard

You are right. Lets not do this then for now. I guess allowing to configure this value is the be the better and only sensible approach

PR #5328, opened 2024-09-25
Update Configuration management

docjyJ:
- type checking when loading configuration.json
- use an object representing the configuration
- stop writing the file on each modification (single write)

It's in progress, what do you think?

### DIFF
--- a/php/psalm.xml
+++ b/php/psalm.xml
@@ -1,17 +1,17 @@
 <?xml version="1.0"?>
 <psalm
-	errorLevel="2"
-	resolveFromConfigFile="true"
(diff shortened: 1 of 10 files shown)


docjyJ:
What do you think of this organization? It's not finished yet, but I would like to know if the management seems good or too complicated?

szaimen:
> What do you think of this organization? It's not finished yet, but I would like to know if the management seems good or too complicated?

I think it is a bit much currently... Lets first focus on https://github.com/nextcloud/all-in-one/pull/5308 and https://github.com/nextcloud/all-in-one/pull/5307 :)

docjyJ:
Yes, step by step

PR #4804, opened 2024-06-09
Add remote borg backup support

timdiels:
This adds support for backing up directly to a remote borg repo. ([I tried reaching out earlier](https://github.com/nextcloud/all-in-one/discussions/1068#discussioncomment-8907198))

### Why?
I'd like to backup/restore via the UI, resist ransomware and store the backups remotely without storing a local (compressed) copy of the nextcloud data.

Alternatives (https://github.com/nextcloud/all-in-one?tab=readme-ov-file#are-remote-borg-backups-supported):

- Mount a network file system: Ransomware can encrypt the backups on the network storage.
- rclone: Ransomware and stores a local copy of the borg repo
- [borg backup the local repo to borgbase](https://github.com/nextcloud/all-in-one/discussions/4391): Stores a local compressed copy of the nextcloud data as at least 1 borg backup
- create your own solution: backups cannot be managed through the UI and others will have to reinvent my work

### How to use it
Instead of entering a local repo path (which remains supported), you can choose to instead enter a remote borg repo url. (These screenshots are from a disaster recovery but it's analogous, just ignore the passphrase):

Screenshot from 2024-06-09 15-26-41

The first you try to initialise the repo, the backup container will create an ssh key and foolishly try to init the borg repo resulting in error because you have to authorise the ssh key it generated first. So, e.g. I copy paste the public key shown here to my borgbase repo:

Screenshot from 2024-06-09 15-27-01

Then I can try again as instructed and it should work. The following backup info is shown:

Screenshot from 2024-06-09 14-59-30

### Changes
- Support for storing backups directly at a remote borg repo (Replace commands to not assume the repo is local, e.g. use BORG_REPO env var, ...)
- Exclude files from backups if we don't plan on restoring them anyway, such as the audit log
- Restore with `borg extract` because, at least with a remote repo, `borg mount` is very slow (20 seconds vs 24 minutes restore for pretty much an empty nextcloud install).

FYI Oddly the original code never restores host-mounts.

### Testing
I hacked the code til I could run local deployments of nextcloud-aio without any builds and manually tested:

- Entering invalid stuff into forms I changed
- Remote borg repo init
- Creating and restoring a backup. I created some files in the master container volume; some in an ignored directory to make sure those are not deleted, some in a directory that is part of the backup, after the backup made some changes and made sure that once restored, the ignored files were left alone, new (non-ignored) files were deleted and changed files restored. I can't really open the user UI with my local test setup.
- Checking a backup
- Setting up a new nextcloud from backup

### TODO
[...]

PR #5370, opened 2024-10-02
Add nextcloud/coding-standard dependency

docjyJ:
### Add `nextcloud/coding-standard`

Attention, all files are affected!
```
Found 43 of 43 files that can be fixed in 4.645 seconds, 18.00 MB memory used
Script php-cs-fixer fix --dry-run --diff handling the cs:check event returned with error code 8
```


docjyJ:
I split into 2 commits:
- b3c7cbe added bibliotech
- cad1a5a run the cs

szaimen:
Thanks again for your contributions @docjyJ! I think this looks good but probably makes sense to wait for e.g. this PR: https://github.com/nextcloud/all-in-one/pull/4804 as otherwise it will be a rebase hell...

docjyJ:
Yes, I know, complicated to merge. Since it's only style, it must not be the worst case.

I dropped the commit from run cs:fix and rebased the branch.

PR #5372, opened 2024-10-02
Use enum instead of interface for state

docjyJ:
I use VersionState instade of bool

### DIFF
--- a/php/src/Container/WorkingState.php
+++ b/php/src/Container/WorkingState.php
@@ -1,0 +1,12 @@
+<?php
+
+namespace AIO\Container;
+
+enum WorkingState: string {
+    case ImageDoesNotExist = 'image_does_not_exist';
+    case NotRestarting = 'not_restarting';
+    case Restarting = 'restarting';
+    case Running = 'running';
+    case Starting = 'starting';
+    case Stopped = 'stopped';
+}
--- a/php/src/Container/Container.php
+++ b/php/src/Container/Container.php
@@ -2,7 +2,6 @@
 
 namespace AIO\Container;
 
-use AIO\Container\State\IContainerState;
 use AIO\Data\ConfigurationManager;
 use AIO\Docker\DockerActionManager;
 use AIO\ContainerDefinitionFetcher;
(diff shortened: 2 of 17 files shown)


st3iny (approved):
Looks good but did not test.

@szaimen I'd say we are good to go if you are fine with the naming:
`IContainerState` was refactored to enums `WorkingState` and `VersionState`

docjyJ:
I didn't test to. Workflow pass and @szaimen you have better test instance than me 😅

docjyJ:
> `IContainerState` was refactored to enums `WorkingState` and `VersionState`

I split them in two, because it didn't make sense for the version and work status to be the same interface...

php/src/Container/WorkingState.php
```diff
@@ -0,0 +1,12 @@
+<?php
+
+namespace AIO\Container;
+
+enum WorkingState: string {
```
szaimen, line 5: @docjyJ I just discussed this with Richard and please rename it to `ContainerState`. Otherwise it looks fine :)

szaimen (changes requested):

php/src/Container/WorkingState.php
docjyJ: Do I rename the state version?

szaimen: you mean versionstate? No, that sounds fine

szaimen:
```suggestion
enum ContainerState: string {
```

docjyJ:
Done, #5307 will be redy to

php/src/Container/WorkingState.php
docjyJ: > you mean versionstate?

Yes sorry, I wrote that quickly on my phone.

szaimen (approved):
LGTM. Thanks a lot! 😊

szaimen:
This is now released with v9.7.0 Beta. Testing and feedback is welcome! See https://github.com/nextcloud/all-in-one#how-to-switch-the-channel

PR #5377, opened 2024-10-04
added docker socket proxy to manual install

sheikhmishar:
I've added and tested docker-socket-proxy for the manual installation of AIO. This will be useful for using AppAPI and ExApp for manual mode users


A4alli:
the internal docker-socket-proxy fails at Init-step on docker-rootless.

PR #5397, opened 2024-10-08
feat(ui): introducing dark-mode and refactor some templates and css

ernolf:
Introducing dark-mode

[AIO-dark-mode.webm](https://github.com/user-attachments/assets/f2a6c208-018a-4ac3-9ee0-52dce008df75)


szaimen:
Woow!!! Many thanks @ernolf for this amazing contribution🎉😊

php/templates/already-installed.twig
```diff
@@ -1,5 +1,20 @@
...
+        </form>
+    </header>
+    <div class="login">
+        <svg class="nextcloud-logo" xmlns="http://www.w3.org/2000/svg" viewBox="0 0 142 100" width="142" height="100"">
+            <use href="/img/nextcloud-logo.svg#logo"></use>
+            <use href="/img/nextcloud-logo.svg#Nextcloud"></use>
+            <text x="10" y="50" fill="var(--color-nextcloud-logo)" class="fallback-text">Nextcloud Logo</text>
+        </svg>
+        <h3>⚠️ Nextcloud All-In-One is already installed</h3>
+    </div>
```
szaimen, line 19:
```suggestion
    <div class="login">
        <svg class="nextcloud-logo" xmlns="http://www.w3.org/2000/svg" viewBox="0 0 142 100" width="142" height="100"">
            <use href="/img/nextcloud-logo.svg#logo"></use>
            <use href="/img/nextcloud-logo.svg#Nextcloud"></use>
            <text x="10" y="50" fill="var(--color-nextcloud-logo)" class="fallback-text">Nextcloud Logo</text>
        </svg>
        <h2>Nextcloud All-In-One is already installed</h2>
        <a href="/" class="button" target="_blank" rel="noopener">Open Nextcloud AIO login ↗</a>
    </div>
```

php/templates/containers.twig
```diff
@@ -5,683 +5,20 @@
...
-    {% if isApacheStarting == true or is_backup_container_running == true or isWatchtowerRunning == true or is_daily_backup_running == true %}
-        <script type="text/javascript" src="automatic_reload.js"></script>
-    {% else %}
-        <script type="text/javascript" src="before-unload.js"></script>
-    {% endif %}
-
-    </main>
+    <div class="container">
+        {{ include('includes/containers-main.twig') }}
+    </div>
```
szaimen, line 20: lets please do this in a follow-up PR to keep the changes minimal and related :)

szaimen (commented):
I have some remarks though and leave it further down but looks really great already. Thanks again! :)

php/public/style.css
```diff
@@ -243,3 +414,58 @@ header > form {
     0% { transform: rotate(0deg); }
     100% { transform: rotate(360deg); }
 }
+
+/* General theme button styling */
+#theme-toggle {
```
docjyJ, line 419: Better to use classes instead of identifiers

php/public/toggle-dark-mode.js
```diff
@@ -0,0 +1,26 @@
+// Function to toggle theme
+function toggleTheme() {
+    const currentTheme = document.documentElement.getAttribute('data-theme');
+    const newTheme = (currentTheme === 'dark') ? 'light' : 'dark';
+    document.documentElement.setAttribute('data-theme', newTheme);
+    localStorage.setItem('theme', newTheme);
+
+    // Change the icon based on the current theme
+    const themeIcon = document.getElementById('theme-icon');
+    themeIcon.textContent = newTheme === 'dark' ? '☀️' : '🌙'; // Switch between moon and sun icons
```
docjyJ, line 10: I'm not a fan of emoji, their rendering is random and depends on the system...

php/templates/already-installed.twig
docjyJ: Yes, and use `a.button` for navigation

php/templates/includes/containers-main.twig
docjyJ, line 1: Can you make 2 commits, 1 for the file naming and 1 for the changes? It will be more readable and will limit conflicts.

php/public/toggle-dark-mode.js
```diff
@@ -0,0 +1,26 @@
+// Function to toggle theme
+function toggleTheme() {
+    const currentTheme = document.documentElement.getAttribute('data-theme');
+    const newTheme = (currentTheme === 'dark') ? 'light' : 'dark';
```
docjyJ, line 4: Would it be possible to add an 'auto' option? Sync with browser theme?

docjyJ (changes requested):
Not tested, but it's cool!

docjyJ:
The ideal would be to have a Vuejs front and a JSON API :face_with_open_eyes_and_hand_over_mouth: :roll_eyes:

docjyJ:
If you don't know, on this site there are lots of tips including sync with OS

https://css-tricks.com/a-complete-guide-to-dark-mode-on-the-web/

szaimen:
> The ideal would be to have a Vuejs front and a JSON API 🫢 🙄

I am not sure about this. I think vuejs is overkill for such a small site. It only adds maintainability issues imho.

php/public/toggle-dark-mode.js
ernolf: > their rendering is random and depends on the system...

yes and no
Meanwhile the display of emojis is in general as good as uniform in every browser, but some browser extensions (such as Chromoji) can affect the display though. However, since this project already makes a lot of use of emojis and all emoji fonts are already loaded in the CSS and these are two very common emojis that are an integral part of utf8 and are already displayed in every shell, for example, and it doesn't matter at all whether they are displayed exactly in one way or another for this purpose, I wanted to leave it as it is.
I can of course also use two svg elements, but I found that to be absolutely unnecessary considering what I wanted to achieve and display.

php/templates/containers.twig
ernolf: OK. :+1:

php/templates/includes/containers-main.twig
ernolf: Yes, look https://github.com/nextcloud/all-in-one/pull/5397#discussion_r1793551278

php/templates/already-installed.twig
ernolf: ✔️done

php/templates/includes/containers-main.twig
ernolf: ✔️done

php/public/toggle-dark-mode.js
szaimen: I am fine eith the emoji 👍

php/public/toggle-dark-mode.js
szaimen: Lets keep the manual mode for the first iteration. We can think about improvements in follow-ups

php/templates/containers.twig
```diff
@@ -5,683 +5,685 @@
...
             <input type="hidden" name="{{csrf.keys.value}}" value="{{csrf.value}}">
             <input type="submit" value="Log out" />
         </form>
     </header>
 
-    <main>
-        <h1>Nextcloud AIO v9.7.0</h1>
+    <div class="container">
+        <main>
+            <h1>Nextcloud AIO v9.6.0</h1>
```
szaimen, line 20:
```suggestion
            <h1>Nextcloud AIO v9.7.0</h1>
```

php/templates/containers.twig
```diff
@@ -5,683 +5,685 @@
...
-        {% endfor %}
-
-        {% if is_daily_backup_running == true %}
-            <p><span class="status running"></span> Daily backup currently running. (<a href="/api/docker/logs?id=nextcloud-aio-mastercontainer" target="_blank" rel="noopener">Logs</a>)</p>
-            {% if automatic_updates == true %}
-                <p>This will update your containers, the mastercontainer and, on Saturdays, your Nextcloud apps if the backup is successful.</p>
-                {% if is_mastercontainer_update_available == true %}
-                    <p>When the mastercontainer is updated it will restart, making it unavailable for a moment. (<a href="/api/docker/logs?id=nextcloud-aio-watchtower" target="_blank" rel="noopener">Logs</a>)</p>
+            {% for container in containers %}
+                {% if container.GetDisplayName() != '' and class(container.GetRunningState()) == 'AIO\\Container\\State\\RunningState' %}
```
szaimen, line 44: Same changes below
```suggestion
                {% if container.GetDisplayName() != '' and class(container.GetRunningState().value == 'running' %}
```

szaimen: you probably need to rebase

szaimen:
Btw @ernolf i think we need to increase for upgrade scenarios: https://github.com/nextcloud/all-in-one/blob/a647d10bf073cc807233290aa3b2cd4b06935699/php/templates/layout.twig#L4

docjyJ:
> I am not sure about this. I think vuejs is overkill for such a small site. And it adds maintainability issues imho.

I just wondered. Why use twig?
Isn't it simpler in standard PHP?

szaimen:
> > I am not sure about this. I think vuejs is overkill for such a small site. And it adds maintainability issues imho.
> 
> I just wondered. Why use twig? Isn't it simpler in standard PHP?

Yes it is :)

docjyJ:
> Yes it is :)

So why twig?

szaimen:
> > Yes it is :)
> 
> So why twig?

Sorry I made a typo. Twig is easier than php :)

docjyJ:
> Sorry I made a typo. Twig is easier than php :)

Maybe... My IDE can't infer types in templates.
What do you use?

szaimen:
> > Sorry I made a typo. Twig is easier than php :)
> 
> Maybe... My IDE can't infer types in templates. What do you use?

Maybe this helps if you should be using phpstorm? https://plugins.jetbrains.com/plugin/7303-twig

docjyJ:
It does not infer type...

szaimen:
> It does not infer type...

Too bad

szaimen:
Lets maybe discuss this in a separate issue. Do you want to create an issue?

ernolf:
> Btw @ernolf i think we need to increase v2 for upgrade scenarios:
> 
> https://github.com/nextcloud/all-in-one/blob/a647d10bf073cc807233290aa3b2cd4b06935699/php/templates/layout.twig#L4

See https://github.com/nextcloud/all-in-one/pull/5397/commits/bf47628073242bd2a045034048e741acbe57b919

szaimen (approved):
LGTM, thanks a lot 💙
This needs to wait for the next stable release though before it can be merged (ca. 1 week)

php/public/style.css
docjyJ: It's OK if is only a sigle element

docjyJ (approved):
LGTM

php/public/toggle-dark-mode.js
```diff
@@ -0,0 +1,26 @@
...
+
+    // Change the icon based on the current theme
+    const themeIcon = document.getElementById('theme-icon');
+    themeIcon.textContent = newTheme === 'dark' ? '☀️' : '🌙'; // Switch between moon and sun icons
+}
+
+// Function to apply saved theme from localStorage
+function applySavedTheme() {
+    const savedTheme = localStorage.getItem('theme');
+    if (savedTheme) {
```
szaimen, line 16: due to the above logic there is no icon shown if you load the page the first time after the update

szaimen:
Another thing: the spinner does not have a dark mode yet and the log out btton should be hidden behind the spinner:
image

szaimen:
Another thing:
it seems like upon reload in dark mode you get a white flashing. not sure how/if we can resolve it but it is quite disturbing:

https://github.com/user-attachments/assets/d12fa7eb-79ed-4388-a215-e9f428fa6ddc

szaimen:
I fear we also need to adjust the mobile layout: the icons on top should not overlay the content and the bottom shuld not leak through:
image

szaimen:
This is now released with v9.8.0 Beta. Testing and feedback is welcome! See https://github.com/nextcloud/all-in-one#how-to-switch-the-channel

PR #5398, opened 2024-10-09
nginx: adjust configs

szaimen:


Zoey2936:
The first change is because of collabora right? And is there a reason for the second change?

szaimen:
> The first change is because of collabora right? And is there a reason for the second change?

The first change is because this disables any compression on the server side so that the js and css files do not get compressed anymore.

The second was a recommendation by @st3iny

Zoey2936 (approved):

szaimen:
Thanks Zoey :)

PR #5409, opened 2024-10-13
[reverse-proxy.md] Add Tailscale integration setup

flll:
- Detailed explanation of integrating Tailscale, Caddy, and AIO
- Add example configuration for Compose file
- Include example Caddyfile configuration
- Provide instructions for setting environment variables
- Include notes on security and network configuration


szaimen:
Hi, thanks a lot for your contribution! I think in general does this look good but it is a bit too verbose for this readme. I think this would better fit as a separate [wiki article](https://github.com/nextcloud/all-in-one/discussions/new?category=wiki) and then maybe link it back to this doc as it is done with [Citrix ADC VPX / Citrix Netscaler](https://github.com/nextcloud/all-in-one/blob/main/reverse-proxy.md#citrix-adc-vpx--citrix-netscaler) for example? :)

flll:
Thank you for your feedback.
We have moved the guide to the wiki.

szaimen (approved):
Thanks a lot! LGTM 😊

PR #5418, opened 2024-10-14
CI: add copyright update workflow

szaimen:


.github/workflows/update-copyright.yml
```diff
@@ -11,3 +11,43 @@ jobs:
...
+          cd .build
+          php license.php
+          cd ../
+
+      - name: Run script 2
+        run: |
+          set -x
+          cd ../
+          # todo: remove the single branch below and clone master
+          git clone https://github.com/nextcloud/github_helper.git --depth 1 --single-branch --branch enh/noid/fix-exit
```
szaimen, line 34: todo

szaimen:
This can be done tremendously easier. for example the authors can be retreived via so:
`git log | grep Author: | sort -u`.

Then simply add the license header via sh script
`SPDX-FileCopyrightText: 2024 Nextcloud GmbH and Nextcloud contributors`
`SPDX-License-Identifier: AGPL-3.0-or-later`

AndyScherzinger:
Sounds good while the year should then be the earliest one meaning to reflect the original creation date.

AndyScherzinger:
The git + grep works fine for the authors file, which is also what I use to fill the file.

PR #5449, opened 2024-10-21
Add ENV variables to adjust max_file_uploads and max_chunk_size

JMarcosHP:


szaimen:
Hi @JMarcosHP , thanks a lot for the PR!

Did you already do the following https://github.com/nextcloud/all-in-one/discussions/5095#discussioncomment-10263837? Did it make it work for you?

JMarcosHP:
> Hi @JMarcosHP , thanks a lot for the PR!
> 
> Did you already do the following [#5095 (reply in thread)](https://github.com/nextcloud/all-in-one/discussions/5095#discussioncomment-10263837)? Did it make it work for you? (We need this information in order to judge if it makes sense to merge your PR going forward)

Hi, yes I've tested my self that parameters and fixed my issue, hope this will help Cloudflare Tunnel users.

JMarcosHP:
I think now this is ready for review and merge.

szaimen:
First of all thanks a lot for your PR!

The thing is that with https://github.com/nextcloud/server/pull/45652 (which brings chunking for public uploads) and https://github.com/nextcloud/server/issues/47682 which consists of https://github.com/nextcloud/server/pull/48461 and https://github.com/nextcloud/desktop/pull/7161 that improves the default chunking situation, I'd like to wait for Nextcloud 31 and a Desktop client release that includes the change first and afterwards re-evaluate the situation to check if we actually need this setting afterwards. 

I hope this makes sense?

PR #5460, opened 2024-10-22
readme: add tailscale network instructions

szaimen:
@flll does this look good to you?


flll:
Thank you!
It looks very good.

flll (approved):

szaimen:
Thank you for creating the guide! 😊

A4alli:
It would be awful to add instructions for non-docker users as well. moreover if nginx, caddy,tailscale and cloudflare all together implement. This setup will not expose any port and there will no need of ddns. All external access will be through tailscale network.

szaimen:
As Nextcloud AIO is buolt around docker, I dont think this makes too much sense...

PR #5467, opened 2024-10-24
local-instance-docs: add summary and add tailscale network as option

szaimen:


flll (approved):
That's great! Thank you for explaining Tailscale so clearly and helping to spread awareness about it. I appreciate you mentioning my name as well.

szaimen:
Thanks in return for this awesome guide! 😊
This was honestly a missing puzzle piece that was miasing since a while but since I dont have any expertise with tailscale I never managed to create something like this.

PR #5481, opened 2024-10-25
aio-interface: fix dark-mode and checkbox style

ernolf:
Follow-up to https://github.com/nextcloud/all-in-one/pull/5397

1. fix(ui): split theme & icon load to prevent flicker, errors
2. fix(ui): correct style for disabled checkbox labels
3. fix(ui): adjust height of main to fit in container


szaimen (approved):
LGTM. Thanks a lot! 😊

PR #5484, opened 2024-10-28
 Specifying additional docker network for Apache container

apparle:
This is a fix for https://github.com/nextcloud/all-in-one/discussions/5482


szaimen:
Hi, thanks a lot for your PR!

I am wondering if this would conflict with another workflow that we already have that also pushes regularly to gh-pages?
See https://github.com/nextcloud/all-in-one/blob/main/.github/workflows/helm-release.yml

compose.yaml
```diff
@@ -17,6 +17,7 @@ services:
       # AIO_COMMUNITY_CONTAINERS: # With this variable, you can add community containers very easily. See https://github.com/nextcloud/all-in-one/tree/main/community-containers#community-containers
       # APACHE_PORT: 11000 # Is needed when running behind a web server or reverse proxy (like Apache, Nginx, Caddy, Cloudflare Tunnel and else). See https://github.com/nextcloud/all-in-one/blob/main/reverse-proxy.md
       # APACHE_IP_BINDING: 127.0.0.1 # Should be set when running behind a web server or reverse proxy (like Apache, Nginx, Caddy, Cloudflare Tunnel and else) that is running on the same host. See https://github.com/nextcloud/all-in-one/blob/main/reverse-proxy.md
+      # APACHE_CONTAINER_ADDITIONAL_NETWORK: frontend_net # (Optional) Connect the apache container to an additional docker network. Needed when behind a web server or reverse proxy (like Apache, Nginx, Caddy, Cloudflare Tunnel and else) running in a different docker network on same server. See https://github.com/nextcloud/all-in-one/blob/main/reverse-proxy.md
```
szaimen, line 20: I would probably name it like that
```suggestion
      # APACHE_ADDITIONAL_NETWORK: frontend_net # (Optional) Connect the apache container to an additional docker network. Needed when behind a web server or reverse proxy (like Apache, Nginx, Caddy, Cloudflare Tunnel and else) running in a different docker network on same server. See https://github.com/nextcloud/all-in-one/blob/main/reverse-proxy.md
```

php/src/Data/ConfigurationManager.php
```diff
@@ -888,6 +888,13 @@ public function DeleteCollaboraDictionaries() : void {
         $this->WriteConfig($config);
     }
 
+    public function GetApacheContainerAdditionalNetwork() : string {
+        $envVariableName = 'APACHE_CONTAINER_ADDITIONAL_NETWORK';
```
szaimen, line 892:
```suggestion
        $envVariableName = 'APACHE_ADDITIONAL_NETWORK';
```

php/src/Docker/DockerActionManager.php
```diff
@@ -897,11 +924,23 @@ public function ConnectMasterContainerToNetwork() : void
...
     }
 
     public function ConnectContainerToNetwork(Container $container) : void
     {
         $this->ConnectContainerIdToNetwork($container->GetIdentifier(), $container->GetInternalPort());
+
+        if ($container->GetIdentifier() === 'nextcloud-aio-apache') {
+            $apacheContainerAdditionalNetwork = $this->configurationManager->GetApacheContainerAdditionalNetwork();
+            if ($apacheContainerAdditionalNetwork !== '') {
+                $this->ConnectContainerIdToExistingNetwork($container->GetIdentifier(), $container->GetInternalPort(), $apacheContainerAdditionalNetwork);
```
szaimen, line 941: Can we maybe simply re-use the ConnectContainerIdToNetwork function?

Possibly something like this? ConnectContainerIdToNetwork(nextcloud-aio-apache, $internalPort, $apacheContainerAdditionalNetwork)

php/src/Docker/DockerActionManager.php
```diff
@@ -844,6 +844,33 @@ private function DisconnectContainerFromBridgeNetwork(string $id) : void
         }
     }
 
+    private function ConnectContainerIdToExistingNetwork(string $id, string $internalPort, string $network) : void
```
szaimen, line 847:
```suggestion
```

reverse-proxy.md
```diff
@@ -44,9 +44,14 @@ All examples below will use port `11000` as `APACHE_PORT`. This port will be exp
...
+    * Reverse Proxy container is connected to `nextcloud-aio` network, either by setting it up under the same service (same docker compose file) or as a secondary external network in the reverse proxy container service.
+    * The front-facing nextcloud containers (`nextcloud-aio-mastercontainer` and `nextcloud-aio-apache`) are connected to the docker network of the reverse proxy container by specifying env variable `APACHE_CONTAINER_ADDITIONAL_NETWORK`.
+
+    After this the reverse-proxy can utilize Docker DNS networking to access nextcloud by container names at `http://nextcloud-aio-apache:$APACHE_PORT` and `https://nextcloud-aio-mastercontainer:8080`.
+
+    A third alternative is to utilize host networking instead of docker bridge networking: Specify `--network host` option (or `network_mode: host` for docker-compose) as setting for the reverse proxy container to connect it to the host network. If you are using a firewall on the server, you need to open ports 80 and 443 for the reverse proxy manually.  By doing so, the default sample configurations that point at `localhost:$APACHE_PORT` should work without having to modify them.
```
szaimen, line 54: I am not satisfied by the formatting here.

you start an enumeration but are not using that for all the options.

Also, it is not the right place to talk about the `https://nextcloud-aio-mastercontainer:8080` here. imho.

Addiitonally, this setting is only going to work if one uses `SKIP_DOMAIN_VALIDATION: true`. So this needs to be accounted for.

szaimen (changes requested):
Thank you for your PR! :)

I have some remarks below

szaimen:
Additionally, is it possible to add a check for this variable in start.sh so that only allowed characters are actually used?
See for example something like this but for strings: https://github.com/nextcloud/all-in-one/blob/af975f6fde8c81cdb9be6b95186c1301f719cbe4/Containers/mastercontainer/start.sh#L179-L188

php/src/Docker/DockerActionManager.php
apparle: I intentionally separated it out because `ConnectContainerIdToNetwork()` also tries to create the network if it does not exist. 
When someone's using this feature, the network should be created externally and must already exist. If it does not exist, it is likely a typo or incorrect usage, and AIO should throw an error instead of silently creating a new network and proceeding.

If code duplication is a concern I can update the `ConnectContainerIdToNetwork()` to take an optional parameter `alsoCreateNetwork = true` with true as default value, and set to false when trying to connect to this feature.

szaimen: I see but isn't it bad if the process fails? Will not people create addiitonal support requests in that case? I would probably rather go the way to also create the network if it does not exist instead of failing out with the last container...

reverse-proxy.md
szaimen: I am mentioning skip-domain-validation because people would first point at the domaincheck container and after the validation point at the apache container to make the validation work...

apparle:
@szaimen I've made a bunch of changes incorporating your suggestions, added support for domaincheck container to make it all seamless and updated documentation. Can you take a look?

docjyJ (approved):
LGTM

php/src/Docker/DockerActionManager.php
```diff
@@ -844,44 +844,49 @@ private function DisconnectContainerFromBridgeNetwork(string $id) : void
...
-                        'Internal' => false,
-                    ]
-                ]
-            );
-        } catch (RequestException $e) {
-            // 409 is undocumented and gets thrown if the network already exists.
-            if ($e->getCode() !== 409) {
-                throw new \Exception("Could not create the nextcloud-aio network: " . $e->getMessage());
-            }
+        if($createNetwork) {
```
szaimen, line 853:
```suggestion
        if ($createNetwork) {
```

php/src/Docker/DockerActionManager.php
```diff
@@ -844,44 +844,49 @@ private function DisconnectContainerFromBridgeNetwork(string $id) : void
...
+                    }
+                }
         }
 
         $url = $this->BuildApiUrl(
             sprintf('networks/%s/connect', $network)
         );
+        $json_payload = [ 'Container' => $id ];
+        if ($alias !== ''  ) {
+            $json_payload['EndpointConfig'] = [ 'Aliases' => [ $alias ] ];
```
szaimen, line 881:
```suggestion
        $jsonPayload = [ 'Container' => $id ];
        if ($alias !== ''  ) {
            $jsonPayload['EndpointConfig'] = ['Aliases' => [ $alias ]];
```

php/src/Docker/DockerActionManager.php
```diff
@@ -844,44 +844,49 @@ private function DisconnectContainerFromBridgeNetwork(string $id) : void
...
+
         try {
             $this->guzzleClient->request(
                 'POST',
                 $url,
                 [
-                    'json' => [
-                        'container' => $id,
-                    ]
+                    'json' => $json_payload
```
szaimen, line 889:
```suggestion
                    'json' => $jsonPayload
```

php/src/Docker/DockerActionManager.php
```diff
@@ -897,11 +902,28 @@ public function ConnectMasterContainerToNetwork() : void
         $this->ConnectContainerIdToNetwork('nextcloud-aio-mastercontainer', '');
         // Don't disconnect here since it slows down the initial login by a lot. Is getting done during cron.sh instead.
         // $this->DisconnectContainerFromBridgeNetwork('nextcloud-aio-mastercontainer');
+
+        $apacheAdditionalNetwork = $this->configurationManager->GetApacheAdditionalNetwork();
+        if ($apacheAdditionalNetwork !== '') {
+            $this->ConnectContainerIdToNetwork('nextcloud-aio-mastercontainer', '', $apacheAdditionalNetwork, false);
```
szaimen, line 908: Shouldnt this rather be done via the compose.yaml itself I am not sure why the mastercontainer should also be inside the additonal network? what is the benefit of that?

php/src/Docker/DockerActionManager.php
```diff
@@ -897,11 +902,28 @@ public function ConnectMasterContainerToNetwork() : void
...
+        // If a reverse-proxy is relying on container name as hostname this allows it to operate as usual and still validate the domain
+        // The domaincheck container and apache container are never supposed to be active at the same time because they use the same APACHE_PORT anyway, so this doesn't add any new constraints.
+        $alias = ($container->GetIdentifier() === 'nextcloud-aio-domaincheck') ? 'nextcloud-aio-apache' : '';
+
+        $this->ConnectContainerIdToNetwork($container->GetIdentifier(), $container->GetInternalPort(), alias: $alias);
+
+        if ($container->GetIdentifier() === 'nextcloud-aio-apache' || $container->GetIdentifier() === 'nextcloud-aio-domaincheck') {
+            $apacheAdditionalNetwork = $this->configurationManager->GetApacheAdditionalNetwork();
+            if ($apacheAdditionalNetwork !== '') {
+                $this->ConnectContainerIdToNetwork($container->GetIdentifier(), $container->GetInternalPort(), $apacheAdditionalNetwork, false, alias: $alias);
```
szaimen, line 924:
```suggestion
                $this->ConnectContainerIdToNetwork($container->GetIdentifier(), $container->GetInternalPort(), $apacheAdditionalNetwork, false, $alias);
```

szaimen (changes requested):
I have some additional suggestions.

Please also add your new env to https://github.com/nextcloud/all-in-one/blob/main/tests/QA/060-environmental-variables.md

reverse-proxy.md
```diff
@@ -44,9 +44,10 @@ All examples below will use port `11000` as `APACHE_PORT`. This port will be exp
...
+    The reverse-proxy container needs to be connected to the nextcloud containers. This can be achieved one of these 3 ways:
+    1. Utilize host networking instead of docker bridge networking: Specify `--network host` option (or `network_mode: host` for docker-compose) as setting for the reverse proxy container to connect it to the host network. If you are using a firewall on the server, you need to open ports 80 and 443 for the reverse proxy manually. With this setup, the default sample configurations with reverse-proxy pointing to `localhost:$APACHE_PORT` should work directly.
+    1. Connect nextcloud's external-facing containers to the reverse-proxy's docker network by specifying env variable APACHE_ADDITIONAL_NETWORK. With this setup, the reverse proxy can utilize Docker bridge network's DNS name resolution to access nextcloud at `http://nextcloud-aio-apache:$APACHE_PORT`. Note, the specified network must already exist before Nextcloud AIO is started.
```
szaimen, line 49:
```suggestion
    1. Connect nextcloud's external-facing containers to the reverse-proxy's docker network by specifying env variable APACHE_ADDITIONAL_NETWORK. With this setup, the reverse proxy can utilize Docker bridge network's DNS name resolution to access nextcloud at `http://nextcloud-aio-apache:$APACHE_PORT`. ⚠️⚠️⚠️ Note, the specified network must already exist before Nextcloud AIO is started. Otherwise it will fail to start the container because the network is not existing.
```

Containers/mastercontainer/start.sh
```diff
@@ -193,6 +193,19 @@ It is set to '$APACHE_IP_BINDING'."
...
+        print_red "You've set APACHE_ADDITIONAL_NETWORK but not to an allowed value.
+It needs to be a string with letters, numbers, hyphens and underscores.
+It is set to '$APACHE_ADDITIONAL_NETWORK'."
+        exit 1
+    fi
+    tmp_out=$(sudo -u www-data docker network connect $APACHE_ADDITIONAL_NETWORK nextcloud-aio-mastercontainer 2>&1)
+    if [ -n "$tmp_out" ] && [[ ! "$tmp_out" =~ "nextcloud-aio-mastercontainer already exists in network $APACHE_ADDITIONAL_NETWORK" ]]; then
+        print_red "Unable to connect to $APACHE_ADDITIONAL_NETWORK. Cannot continue. Error: \n $tmp_out"
+        exit 1
+    fi
```
szaimen, line 207: I am not sure about this. same question as above

szaimen: @apparle can you please clarify?

apparle:
I replied inline but apparently my comments are all "pending" and I didn't realize they aren't posted for everyone. Clearly my interpretation of github UI is broken. Let me post comments directly, not inline.

apparle:
The intent is to allow for the master container to also be exposed via a reverse proxy. For example, I've my reverse proxy point `https://aio-nextcloud.mydomain.com` to `https://nextcloud-aio-mastercontainer:8080` to access the admin page.

The default compose.yml suggests `network_mode: bridge`, then internally connects the mastercontainer to `nextcloud-aio` and then later disconnects the mastercontainer from the bridge through cron job. Since we're manipulating the network, if a user is specifying an additional network for proxy setup through `APACHE_ADDITIONAL_NETWORK`, it's more seamless to have mastercontainer also be accessible through the same connection. 
Yes, the user can create this connection manually -- but then the user must specify not just the network for proxy but also create `nextcloud-aio` doing something like below, which is again not obvious for a naive user.
```
services:
...
    networks:
      - nextcloud-aio 
      - front_net

networks:
  front_net:
    external: true
  nextcloud-aio:
    name: nextcloud-aio
    driver: bridge
```
With the mastercontainer changes I've posed, I imagine the end-user flow to be:
1. User sets up some reverse proxy container with SSL in a docker container. Nginx Proxy Manager with let's encrypt + DuckDNS is extremely simple GUI based and amateur proof. There's at least 3-4 tutorials I've seen on internet which are straight forward to follow.
1. User sets up 2 different routes (just blindly):
    - Some aio URL (`https://aio-nextcloud.mydomain.com`) pointing to `https://nextcloud-aio-mastercontainer:8080` for the admin page.
    - Main nextcloud URL (`https://nextcloud.mydomain.com`) pointing to `http://nextcloud-aio-apache:11000`.
1. User launches the Nextcloud AIO with default compose file and specifies `APACHE_ADDITIONAL_NETWORK=<the network name of reverse proxy>`.
1. Follow all the steps in GUI, and because reverse proxy is already set up, everything including domaincheck just works.

(FWIW, I do intend to post this whole set up somewhere after this feature goes through.)

At a higher level, I suppose it's a software-architecture vs. user-experience question on how much do we do from within the mastercontainer for the sake of consistent user-experience & simplicity between both mastercontainer & apache container vs. leave it up to user to manipulate the compose file. Personally, I think this is a simple enough extension to mastercontainer with minimal downsides, and enables the super-simple reverse proxy + nextcloud setup that I described in above comment. And the `start.sh` change also has the nice side-benefit of doing an early error check with clear error message in red on the docker log, if the specified network has issues.

Containers/mastercontainer/start.sh
szaimen, line 207: so basically do this in a follow-up

php/src/Docker/DockerActionManager.php
```diff
@@ -897,11 +902,28 @@ public function ConnectMasterContainerToNetwork() : void
         $this->ConnectContainerIdToNetwork('nextcloud-aio-mastercontainer', '');
         // Don't disconnect here since it slows down the initial login by a lot. Is getting done during cron.sh instead.
         // $this->DisconnectContainerFromBridgeNetwork('nextcloud-aio-mastercontainer');
+
+        $apacheAdditionalNetwork = $this->configurationManager->GetApacheAdditionalNetwork();
+        if ($apacheAdditionalNetwork !== '') {
+            $this->ConnectContainerIdToNetwork('nextcloud-aio-mastercontainer', '', $apacheAdditionalNetwork, false);
+        }
```
szaimen, line 909: and this too

apparle:
> Hi @apparle thanks for the detailed explanation!
> 
> I was wondering though if we could handle the mastercontainer networking change in a separate PR? (I'd like to merge this PR here without the mastercontainer network change and then do the mastercontainer stuff in a follow-up)
> 
> Is that fine for you?

Sounds good, let me make that change.

szaimen (approved):
LGTM, thanks!

szaimen:
This is now released with v9.9.0 Beta. Testing and feedback is welcome! See https://github.com/nextcloud/all-in-one#how-to-switch-the-channel

ctag:
Just tested on :beta to fix a timeout issue setting up collabora. It worked for fixing the timeout in the settings page, but then I immediately ran into https://github.com/nextcloud/files_pdfviewer/issues/1080 apparently and no documents will open for editing to allow me to test the network/proxy stuff fully.

PR #5499, opened 2024-10-30
Update docker-rootless.md - missprint + improvement

Perlover:
1) No NEXTCLOUD_MOUNT but there is NEXTCLOUD_DATADIR 2) Now in Docker recommended the few different detailed config: https://docs.docker.com/engine/security/rootless/#docker-run--p-does-not-propagate-source-ip-addresses


docker-rootless.md
```diff
@@ -18,7 +18,7 @@ You can run AIO with docker rootless by following the steps below.
...
+All files outside the containers get created, written to and accessed as the user that is running the docker daemon or a subuid of it. So for the built-in backup to work you need to allow this user to write to the target directory. E.g. with `sudo chown -R USERNAME:GROUPNAME /mnt/backup`. The same applies when changing Nextcloud's datadir. E.g. `sudo chown -R USERNAME:GROUPNAME /mnt/ncdata`. When you want to use the NEXTCLOUD_DATADIR option for local external storage, you need to adjust the permissions of the chosen folders to be accessible/writeable by the userid `100032:100032` (if running `grep ^$(whoami): /etc/subuid` as the user that is running the docker daemon returns 100000 as first value).
```
szaimen, line 21: We have two options available:
```suggestion
All files outside the containers get created, written to and accessed as the user that is running the docker daemon or a subuid of it. So for the built-in backup to work you need to allow this user to write to the target directory. E.g. with `sudo chown -R USERNAME:GROUPNAME /mnt/backup`. The same applies when changing Nextcloud's datadir via NEXTCLOUD_DATADIR. E.g. `sudo chown -R USERNAME:GROUPNAME /mnt/ncdata`. When you want to use the NEXTCLOUD_MOUNT option for local external storage, you need to adjust the permissions of the chosen folders to be accessible/writeable by the userid `100032:100032` (if running `grep ^$(whoami): /etc/subuid` as the user that is running the docker daemon returns 100000 as first value). 

```

szaimen (changes requested):
Hi, thanks for your PR!

Hi have one remark below

docker-rootless.md
Perlover: It seems I was mistaken. [The `NEXTCLOUD_MOUNT` environment variable does exist](https://github.com/nextcloud/all-in-one?tab=readme-ov-file#how-to-allow-the-nextcloud-container-to-access-directories-on-the-host). I found it in the documentation, so your latest change appears to be correct.

szaimen (approved):

PR #5539, opened 2024-11-07
Connect mastercontainer to docker network specified by APACHE_ADDITIONAL_NETWORK

apparle:
This is an extension of https://github.com/nextcloud/all-in-one/pull/5484 to allow for `mastercontainer`'s admin page to also be behind a reverse proxy (for security, authentication etc.) instead of just an unsecured localhost port. For example, I've my reverse proxy set up to point `https://aio-nextcloud.mydomain.com` to `https://nextcloud-aio-mastercontainer:8080` for the admin page.

Since the user is specifying `APACHE_ADDITIONAL_NETWORK` as a means to connect the reverse proxy, it is natural to use the same information for mastercontainer's reverse proxy setup as well. Yes in theory, the user can create this connection by docker commands or docker compose file (I do this today), but then the user must specify not just the network for proxy but also create `nextcloud-aio` doing something like below:
```
services:
...
    networks:
      - nextcloud-aio 
      - front_net

networks:
  front_net:
    external: true
  nextcloud-aio:
    name: nextcloud-aio
    driver: bridge
```
This isn't easy or obvious for newcomer, and it'd be much more seamless if handled automatically. FWIW, there's some precedence for mastercontainer manipulating its own network -- the default compose.yml suggests `network_mode: bridge`, then mastercontainer connects itself to `nextcloud-aio` and then later disconnects itself from the bridge through a cron job. Since we're manipulating the network, it's a relatively incremental extension to connect mastercontainer with reverse proxy with minimal downsides. 
And the `start.sh` change also has the nice side-benefit of doing an early error check with clear error message if the specified network has issues.

After this change, I imagine the end-user flow to be:
1. User sets up some reverse proxy container with SSL in a docker container. Eg: Nginx Proxy Manager with let's encrypt + DuckDNS is extremely simple GUI based and amateur proof. There's at least 3-4 tutorials I've seen on internet which are straight forward to follow. And potentially authentication schemes as well.
1. User sets up 2 different routes:
    - Some aio URL (eg: `https://aio-nextcloud.mydomain.com`) pointing to `https://nextcloud-aio-mastercontainer:8080` for the admin page.
    - Main nextcloud URL (eg: `https://nextcloud.mydomain.com`) pointing to `http://nextcloud-aio-apache:11000`.
1. User launches the Nextcloud AIO with default compose file and specifies `APACHE_ADDITIONAL_NETWORK=<the network name of reverse proxy>`.
1. Follow all the steps in GUI, and because reverse proxy is already set up, everything including domaincheck just works.

(FWIW, I do intend to post this as a guide somewhere after this pull request goes through.)

### DIFF
--- a/Containers/mastercontainer/start.sh
+++ b/Containers/mastercontainer/start.sh
@@ -200,9 +200,9 @@ It needs to be a string with letters, numbers, hyphens and underscores.
 It is set to '$APACHE_ADDITIONAL_NETWORK'."
         exit 1
     fi
-    tmp_out=$(sudo -u www-data docker network connect $APACHE_ADDITIONAL_NETWORK nextcloud-aio-mastercontainer 2>&1)
-    if [ -n "$tmp_out" ] && [[ ! "$tmp_out" =~ "nextcloud-aio-mastercontainer already exists in network $APACHE_ADDITIONAL_NETWORK" ]]; then
-        print_red "Unable to connect to $APACHE_ADDITIONAL_NETWORK. Cannot continue. Error: \n $tmp_out"
+    tmp_out=$(sudo -u www-data docker network connect "$APACHE_ADDITIONAL_NETWORK" nextcloud-aio-mastercontainer 2>&1)
+    if [ "$?" -ne 0 ] && [[ ! "$tmp_out" == *"nextcloud-aio-mastercontainer already exists in network $APACHE_ADDITIONAL_NETWORK"* ]]; then
+        print_red "Unable to connect to $APACHE_ADDITIONAL_NETWORK. Cannot continue. Error: $tmp_out"
         exit 1
     fi
 fi
--- a/Containers/mastercontainer/start.sh
+++ b/Containers/mastercontainer/start.sh
@@ -200,10 +200,11 @@ It needs to be a string with letters, numbers, hyphens and underscores.
 It is set to '$APACHE_ADDITIONAL_NETWORK'."
         exit 1
     fi
-    tmp_out=$(sudo -u www-data docker network connect "$APACHE_ADDITIONAL_NETWORK" nextcloud-aio-mastercontainer 2>&1)
-    if [ "$?" -ne 0 ] && [[ ! "$tmp_out" == *"nextcloud-aio-mastercontainer already exists in network $APACHE_ADDITIONAL_NETWORK"* ]]; then
-        print_red "Unable to connect to $APACHE_ADDITIONAL_NETWORK. Cannot continue. Error: $tmp_out"
-        exit 1
+    if ! tmp_out=$(sudo -u www-data docker network connect "$APACHE_ADDITIONAL_NETWORK" nextcloud-aio-mastercontainer 2>&1); then
+        if [[ ! "$tmp_out" == *"nextcloud-aio-mastercontainer already exists in network $APACHE_ADDITIONAL_NETWORK"* ]]; then
+            print_red "Unable to connect to $APACHE_ADDITIONAL_NETWORK, cannot continue. Error: $tmp_out"
+            exit 1
+	fi
     fi
 fi
 if [ -n "$TALK_PORT" ]; then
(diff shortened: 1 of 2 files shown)


Containers/mastercontainer/start.sh
```diff
@@ -200,6 +200,11 @@ It needs to be a string with letters, numbers, hyphens and underscores.
 It is set to '$APACHE_ADDITIONAL_NETWORK'."
         exit 1
     fi
+    tmp_out=$(sudo -u www-data docker network connect $APACHE_ADDITIONAL_NETWORK nextcloud-aio-mastercontainer 2>&1)
+    if [ -n "$tmp_out" ] && [[ ! "$tmp_out" =~ "nextcloud-aio-mastercontainer already exists in network $APACHE_ADDITIONAL_NETWORK" ]]; then
+        print_red "Unable to connect to $APACHE_ADDITIONAL_NETWORK. Cannot continue. Error: \n $tmp_out"
+        exit 1
+    fi
```
docjyJ, line 207:
```suggestion
    tmp_out=$(sudo -u www-data docker network connect $APACHE_ADDITIONAL_NETWORK nextcloud-aio-mastercontainer 2>&1)
    # not tested
    if [ $? -ne 0 ]; then
        print_red "Unable to connect to $APACHE_ADDITIONAL_NETWORK. Cannot continue. Error: \n $tmp_out"
        exit 1
    fi
```
Is it possible to use the exit code? that would be more relevant

apparle: Originally I did use exit code in `$?`, but that is also non-zero when `nextcloud-aio-mastercontainer` is already connected to the specified network; which can happen because someone explicitly connected them via the docker compose file or because it's being restarted etc.
That's why I rewrote it to above code which catches the output and excludes the condition `nextcloud-aio-mastercontainer already exists in network`

apparle: I can potentially change it to this, if that's better?
```
    if [ $? -ne 0 ] && [[ ! "$tmp_out" =~ "nextcloud-aio-mastercontainer already exists in network $APACHE_ADDITIONAL_NETWORK" ]]; then
```

docjyJ: Does the exit code change depending on the error type?

apparle: No, it does not. It's always 1.

docjyJ (approved):
LGTM

szaimen (changes requested):
Honestly I still dislike this idea. I think it is better to simply add docs on the topic and add the external network via docker-compose if you really want to. The mastercontainer will only remove the default `bridge` network and add itself to the `nextcloud-aio` network (and will even create it if it does not exist) but will not modify the network config any further.

apparle:
> > Yes in theory, the user can create this connection by docker commands or docker compose file (I do this today), but then the user must specify not just the network for proxy but also create `nextcloud-aio` doing something like below:
> 
> No, actually the container will always be connected to a nextcloud-aio network even if you don't specify it in the docker-compose file (as soon as you open the aio-interface). So this is a non-argument.
> 
> And sorry for my late reply btw...

Maybe this was the case some time ago, because I remember this being stated in the default compose file itself. If this is no more a requirement then I agree that directly specifying the frontend network in compose file is better than an environment variable. Let me do some testing.
I'm traveling so it'll take some time before I can test.

apparle:
I finally got around to trying this. Indeed it has got simpler and the front facing network can be connected to the mastercontainer normally in the compose file, with no special knowledge or awareness about the `nextcloud-aio` network itself. Here's the snippet of my compose file which works great:
```
services:
    image: docker.io/nextcloud/all-in-one:latest
    container_name: nextcloud-aio-mastercontainer # This line is not allowed to be changed as otherwise AIO will not work correctly
    # network_mode: bridge # Don't connect to the bridge. Instead connect to below specified frontend network specified below
    networks:
      - frontend_net
    volumes:
        <....>
    environment:
      <...>
      - APACHE_ADDITIONAL_NETWORK=frontend_net 
      <...>

networks:
  frontend_net:
    external: true

```

I don't think this PR is needed any more. If anyone comes around looking for this, above snippet is a good example. Closing

PR #5553, opened 2024-11-08
borgbackup: make logs a bit cleaner if local archive is used

szaimen:
Follow-up to https://github.com/nextcloud/all-in-one/pull/4804

Prevent logs like the following when used with local archive:
```
cat: can't open '.pub': No such file or directory
Your public ssh key for borgbackup is: 
```

### DIFF
--- a/Containers/borgbackup/backupscript.sh
+++ b/Containers/borgbackup/backupscript.sh
@@ -66,7 +66,9 @@ if [ -n "$BORG_REMOTE_REPO" ] && ! [ -f "$BORGBACKUP_KEY" ]; then
     ssh-keygen  -f "$BORGBACKUP_KEY" -N ""
     echo "You should configure the remote to accept this public key"
 fi
-echo "Your public ssh key for borgbackup is: $(cat "$BORGBACKUP_KEY.pub")"
+if [ -n "$BORG_REMOTE_REPO" ]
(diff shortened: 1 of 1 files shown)


Containers/borgbackup/backupscript.sh
```diff
@@ -66,7 +66,7 @@ if [ -n "$BORG_REMOTE_REPO" ] && ! [ -f "$BORGBACKUP_KEY" ]; then
     ssh-keygen  -f "$BORGBACKUP_KEY" -N ""
     echo "You should configure the remote to accept this public key"
 fi
-if [ -n "$BORG_REMOTE_REPO" ]
+if [ -n "$BORG_REMOTE_REPO" ] && [ -f "$BORGBACKUP_KEY" ]; then
```
szaimen, line 69: @timdiels does this make sense?

szaimen: okay, figured it out 🙈

timdiels:
Nit: I would merge both ifs

- if remote
  - if key does not exist, create it 
  - if key exists, cat it

szaimen:
> Nit: I would merge both ifs
> 
> * if remote
>   
>   * if key does not exist, create it
>   * if key exists, cat it

I see... Do you want to create a PR? :)

timdiels:
No no, either is fine really

Issue #5178, opened 2024-08-30
use new notification update feature with nc30


szaimen:
See https://github.com/nextcloud/notifications/pull/2020/files#diff-8ed2619991987ffe33630e0f3af607f3ca23cbfb81a0c1669fa087a2d186bc75

szaimen:
Maybe do this in 2 steps:
1. `use `--object-type='update' --object-id='libx'``
2. `Specify the `--output-id-only` option and store it to later be able to delete it again`

szaimen:
> * Specify the `--output-id-only` option and store it to later be able to delete it again

This is not worth it for now because it would increase the complexity of the implementation on the aio side by a lot

PR #5554, opened 2024-11-08
notify-scripts: add object type and object id

szaimen:
Fix part of https://github.com/nextcloud/all-in-one/issues/5178


szaimen:
> I don't know your versioning theme, but 30+ this is okay

Everything that goes in main now is 30+ :)

szaimen:
BTW, thanks a lot for this feature @nickvergessen 😊

PR #5568, opened 2024-11-10
mastercontainer: enable local testability

apparle:
This PR enables the ability to build and test images locally. It does 2 things:
1. Updates the Dockerfile for mastercontainer to separate out the cloning step into its own stage, such that it can be overridden by user input to pick changes from local path with below command. This follows the principle described in https://www.docker.com/blog/dockerfiles-now-support-multiple-build-contexts/
```
cd Containers/mastercontainer
docker buildx build --build-context aio_root=../../ -t ghcr.io/nextcloud-releases/all-in-one:latest --load .
```
2. Since the hash of a locally built image will never match the latest image, mandating "Update mastercontainer" makes local testing impossible. This PR adds ability to skip the mastercontainer update. This also addresses https://github.com/nextcloud/all-in-one/discussions/5468


szaimen:
Hi @apparle thanks a lot for so many useful contributions! 😊

I just invited you to the repo to make collaboration easier :)

szaimen:
btw this PR has some conflicts...

apparle:
> btw this PR has some conflicts...

Merged main and resolved conflicts. Also resolved lint issues.

docjyJ (commented):
I think it would be better to use an environment variable like SKIP_DOMAIN_VALIDATION=true

apparle:
> I think it would be better to use an environment variable like SKIP_DOMAIN_VALIDATION=true

But then it doesn't solve https://github.com/nextcloud/all-in-one/discussions/5468 easily. It would be better to be able to skip the update temporarily rather than take everything down and restart with different env variables. (And then take it down again to remove the env variables when the reason-to-skip is resolved.)

szaimen:
Hi, I would also prefer a solution that overwrites the behaviour via env variable for example. Adding this to the interface means that we also need to support and maintain it... Also sorry for taking so long to answer.

apparle:
> Hi, I would also prefer a solution that overwrites the behaviour via env variable for example. Adding this to the aio interface means that we also need to support and maintain it... Also sorry for taking so long to answer.

Are you suggesting removing the UI elements altogether and only using an environment variable?
Or are you suggesting to add environment variable in addition to the UI changes?

docjyJ:
image

This is what it could look like. What do you think?

Containers/mastercontainer/Dockerfile
```diff
@@ -1,4 +1,14 @@
 # syntax=docker/dockerfile:latest
+
+# Create a minimal stage to just clone the repo
+FROM alpine/git:2.45.2 AS aio_clone
```
Zoey2936, line 4: I think the official alpine image should be used to keep the packages updated

szaimen:
> image
> 
> This is what it could look like. What do you think?

I am not sure about adding such a button 
to the UI honestly. I fear people will just always use that and refrain from updating and then run into update issues later on because they never updated 🙈

docjyJ:
> I am not sure about adding such a button
to the UI honestly. I fear people will just always use that and refrain from updating and then run into update issues later on because they never updated 🙈

The problem is that the containers created by the AIO are on the lastest tag. (apache, nextcloud postgress...)

But the ideal would be that the AIO uses the same version. It is impossible to downgrade to V8 or V9 because the containers will be created in the latest version (V10).

So if the containers are not on the same version, there will be bugs...

apparle:
Above discussion is useful but I'm confused on the next step. 
@szaimen are you convinced that we'll need some UI anyway to solve https://github.com/nextcloud/all-in-one/discussions/5468 . If we add any UI, an environment variable will just add to the confusion on when is it overridden vs. not; might as well do it all through UI. Thoughts?

@docjyJ yes it could be as minimalistic as your example image in https://github.com/nextcloud/all-in-one/pull/5568#issuecomment-2551028534 but I hesitate to make it that easy. I prefer it be a louder warning to deter users from using that unless they really know what they're doing. That's why the current design.

apparle:
@szaimen Ping. Any thoughts on the last comment?

apparle:
So I thought over this more and realized a better way to do this:
I dramatically simplified this by removing the UI element altogether, since we don't want normal users to skip the update ever. Only in the extreme cases, like local development/testing or https://github.com/nextcloud/all-in-one/discussions/5468 should someone need to bypass the container update. 
And for that purpose, an unadvertised URL parameter can be specified :
```
https://localhost:8080/containers
```
should be modified to 
```
https://localhost:8080/containers?bypass_mastercontainer_update
```
With this change, the mastercontainer update is bypassed temporarily.
We can document this in FAQ, which most users aren't going to try until they're completely stuck and go looking for that in FAQ.

Is that a reasonable balance of usability and safe-enough-defaults ?

Containers/mastercontainer/Dockerfile
apparle: Note, this alpine/git itself is not being packaged into the final generated image. It is only a temporary staging area which is used to run git and clone the latest copy of this repo. And then only certain files from the cloned repo are copied over into the final image, and then this alpine/git image is discarded altogether. That final image is still based on official alpine image as usual.
The only reason to do so is to create a temporary `aio_clone` staging area which can be overridden for local testing as highlighted in the first comment of this PR.

docjyJ (approved):
Should work

szaimen:
Hey, I am very sorry for taking so long to review this. Unfortunately I am still undecided regarding this change. I see the need for such a change but I think I still don't like the implementation, however I also don't have a good idea how to further improve this. So this is why I am currently stuck. Hope that makes things a bit clearer why it takes so long...

Containers/mastercontainer/Dockerfile
```diff
@@ -60,15 +74,10 @@ RUN set -ex; \
     grep -q ';listen.allowed_clients' /usr/local/etc/php-fpm.d/www.conf; \
     sed -i 's|;listen.allowed_clients.*|listen.allowed_clients = 127.0.0.1,::1|' /usr/local/etc/php-fpm.d/www.conf; \
     \
-    apk add --no-cache git; \
     wget https://getcomposer.org/installer -O - | php -- --install-dir=/usr/local/bin --filename=composer; \
     chmod +x /usr/local/bin/composer; \
-    cd /var/www/docker-aio; \
-    git clone https://github.com/nextcloud-releases/all-in-one.git --depth 1 .; \
```
szaimen, line 67: okay, I just got an idea: maybe we can simply make an env out of this and allow to overwrite it thus? So something like this:
```
    git clone "$AIO_GIT_URL" --depth 1 --single-branch --branch "$AIO_GIT_BRANCH" .; \
```
This would make the whole PR much simpler :)

szaimen (changes requested):

Containers/mastercontainer/Dockerfile
szaimen: We would then of course need two new envs `AIO_GIT_URL` and `AIO_GIT_BRANCH` in the Dockerfile and also adjust the readme...

php/public/index.php
```diff
@@ -82,6 +82,11 @@
     $dockerController = $container->get(\AIO\Controller\DockerController::class);
     $dockerActionManger->ConnectMasterContainerToNetwork();
     $dockerController->StartDomaincheckContainer();
+
+    // Check if bypass_mastercontainer_update is provided on the URL, a special developer mode to bypass a mastercontainer update and use local image.
+    $params = $request->getQueryParams();
+    $bypass_mastercontainer_update = isset($params['bypass_mastercontainer_update']);
```
szaimen, line 88: This is a nice solution 👍

Containers/mastercontainer/Dockerfile
apparle: Is there any particular concern / reason you don't like the current way -- it is the official recommended way by docker itself - see Example 3 on https://www.docker.com/blog/dockerfiles-now-support-multiple-build-contexts/  

The git version that you propose still requires a "git push" for every modification to be tested. It works, but it is still inconvenient/cumbersome for a newcomer who's just figuring out code and wants to try / test dumb edits locally without making every small change public (as someone with 0 php background until contributing to this project, I was this person).

szaimen: I think the main problem is the added complexity for the edge case of testing things locally (increasing the complexity for this edge case I am really not a fan of ) - which would still be possible with my suggestion (of course by pushing things first to a remote branch)...

szaimen: I would be fine with my suggestion of adding variables that can be overwritten during build time as the added complexity of this fits the intended feature in terms of  complexity/result imo

apparle: Respectfully, I disagree that it's any extra complexity. If anything it also improves the standard deployment image by reducing its size - git (and all associated dependencies) aren't installed in the final deployed image, only the bare-minimum needed files are in the final image. It's a standard docker build pattern very much in the spirit of creating smallest possible containers. 

But this PR is open for far too long, so I don't want to drag this debate further. I've updated the code to match your suggestions.

szaimen (approved):
LGTM, thanks a lot! ☺️ 

And sorry that it took so long!

I will try to include the feature in the next release that comes after v11.1.0 stable :)

asavageiv:
Would it make sense to apply the bypass flag to all container updates or have a separate one? I am developing both the mastercontainer and apache container simultaneously, so it would be useful for me to avoid the registry updates.

szaimen:
This is now released with v11.2.0 Beta. Testing and feedback is welcome! See https://github.com/nextcloud/all-in-one#how-to-switch-the-channel

PR #5601, opened 2024-11-15
helm: refactor securityContext

szaimen:


nextcloud-aio-helm-chart/templates/nextcloud-aio-database-deployment.yaml
```diff
@@ -23,33 +23,32 @@ spec:
...
+      securityContext:
+        # The items below only work in pod context
+        fsGroup: 999
+        fsGroupChangePolicy: "OnRootMismatch"
+        # The items below work in both contexts
+        runAsUser: 999
+        runAsGroup: 999
+        runAsNonRoot: true
+        seccompProfile:
+          type: RuntimeDefault
```
szaimen, line 35: this might be breaking and should add an opt-in?

nextcloud-aio-helm-chart/templates/nextcloud-aio-database-deployment.yaml
```diff
@@ -23,33 +23,32 @@ spec:
...
-            - name: nextcloud-aio-database-dump
-              mountPath: /nextcloud-aio-database-dump
             - name: nextcloud-aio-database
               mountPath: /nextcloud-aio-database
+          securityContext:
+            # The items below only work in container context
+            allowPrivilegeEscalation: false
+            capabilities:
+              drop: ["ALL"]
+              add: ["NET_BIND_SERVICE"]
```
szaimen, line 51: this might be breaking and should add an opt-in?

nextcloud-aio-helm-chart/templates/nextcloud-aio-fulltextsearch-deployment.yaml
```diff
@@ -24,6 +24,9 @@ spec:
       labels:
         io.kompose.service: nextcloud-aio-fulltextsearch
     spec:
+      securityContext:
+        runAsUser: 1000
+        runAsGroup: 0
```
szaimen, line 29: this needs to be added manually to the script

denppa:
Thank you for your good work, it takes a lot of effort to harden the security aspect of things.

I found out that docker's `SYS_ADMIN` does not actually mean anything in Kubernetes. See [here](https://unofficial-kubernetes.readthedocs.io/en/latest/concepts/policy/container-capabilities/), and according to the docs [here](https://kubernetes.io/docs/tasks/configure-pod-container/security-context/) and specs [here](https://github.com/kubernetes/design-proposals-archive/blob/main/auth/no-new-privs.md), all of them seem to also support this unofficial document.

And if you add `CAP_SYS_ADMIN` instead of `SYS_ADMIN`, it will block `allowPrivilegeEscalation` to be set to true. Also, when `CAP_SYS_ADMIN` is added, Collabora forks and works. Therefore, it would be great if you can include this change in the heml charts.

Thanks.

szaimen:
Hi, thanks for your pointer! I'll look into it!

PR #5671, opened 2024-11-28
nextcloud: allow to adjust name of nextcloud container

szaimen:
Fix https://github.com/nextcloud/all-in-one/discussions/5660


wuast94:
that way i only can set the host which we query the ip from. but when using the apache container in beetween a proxy and the nextcloud container. setting IPv4_ADDRESS directly would resolve the issue :)

PR #5767, opened 2024-12-17
rp-docs: add comment to the config directly to adjust localhost:11000

szaimen:
See https://github.com/nextcloud/all-in-one/pull/5750#discussion_r1888705251


Zoey2936:
maybe also add it as a note to npmplus/npm?

szaimen:
> maybe also add it as a note to npmplus/npm?

Fine by me but arent these screenshots? Can you edit them?

Zoey2936 (approved):
I think it is not needed for NPMplus/NPM

PR #5772, opened 2024-12-18
aio-interface: adjust design of buttons according Nextcloud Vue

docjyJ:
- Use the same color.
- Add secondary button and error.
- Remove ugly border

https://github.com/nextcloud-libraries/nextcloud-vue/blob/master/styleguide/assets/default.css

https://github.com/nextcloud-libraries/nextcloud-vue/blob/master/styleguide/assets/dark.css

https://nextcloud-vue-components.netlify.app/#/Components/NcButton

### DIFF
--- a/php/public/style.css
+++ b/php/public/style.css
@@ -11,8 +11,12 @@
     --color-error-text: #c20505;
     --color-success: #46ba61;
     --color-running: #ffd000;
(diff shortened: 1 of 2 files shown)


szaimen (approved):
LGTM. Thanks! ☺️

PR #5775, opened 2024-12-19
disable fts if seccomp is not supported in the kernel

szaimen:
See https://github.com/nextcloud/all-in-one/discussions/5768


szaimen:
> LGTM Maybe add link to official doc : https://www.elastic.co/guide/en/elasticsearch/reference/8.17/bootstrap-checks-syscall-filter.html

👍
 
> (but users should change their kernel)

Agreed!

szaimen:
Thanks for the approval @docjyJ :)

Issue #5810, opened 2025-01-02
apache-aio healthcheck incompatible with k8s


Claiyc:
### Discussed in https://github.com/nextcloud/all-in-one/discussions/5805

aio-apache probes (https://github.com/nextcloud/all-in-one/blob/main/Containers/apache/healthcheck.sh ) are not compatible with kubernetes best practices and creates a chicken-egg problem in some setups. Read the discussion linked above for an example. They should be refactored to not rely on the connectivity interface that is generally used by external users.

### Steps to reproduce
1. Install the provided helm chart on a plain k8s cluster
2. Attempt to access the aio-apache through a standard nginx-ingress
3. Fail because the pod will never be marked as ready (therefore services will not route requests to the pod, therefore the healthcheck never succeeds)

### Expected behavior 
Pods should be marked as ready once cluster-internal communication between all pods succeeds. Pod readiness should not rely on "outside connections"

### Actual behavior 
Pod is never marked as ready and will therefore crash loop.

#### Host OS 
Debian Bullseye

#### Output of `sudo docker info`
invalid

#### Docker run command or docker-compose file that you used
helm install with minimal configuration according to https://github.com/nextcloud/all-in-one/blob/main/nextcloud-aio-helm-chart/readme.md

szaimen:
This was fixed in the PR but is already fixed with v10.2.0 of the helm chart because I temporarily removed the probes: https://github.com/nextcloud/all-in-one/pull/5829

PR #5825, opened 2025-01-06
apache-healthcheck: remove the domain test

szaimen:
As this is not compatible with k8s.

Fix https://github.com/nextcloud/all-in-one/issues/5810


jermerf:
Literally just dealing with this in 10.1.1. Thanks.

PR #5870, opened 2025-01-13
allow to configure redis user

szaimen:


Containers/nextcloud/Dockerfile
```diff
@@ -147,7 +147,7 @@ RUN set -ex; \
     \
     { \
         echo 'session.save_handler = redis'; \
-        echo 'session.save_path = "tcp://${REDIS_HOST}:6379/${REDIS_DB_INDEX}?auth=${REDIS_HOST_PASSWORD}"'; \
+        echo 'session.save_path = "tcp://${REDIS_HOST}:6379?database=${REDIS_DB_INDEX}&auth[]=${REDIS_USER}&auth[]=${REDIS_HOST_PASSWORD}"'; \
```
szaimen, line 150: This needs further thoughts: for example what happens if REDIS_USER is not set?

szaimen: I fear it doesnt as this config is only written once and then used during runtime... But I suppose this needs some testing from my side...

szaimen: I fear this is not easy to solve. I suppose the string `&auth[]=` would need to be saved in a different variable that would also need to be set on top of REDIS_USER. Another solution would be to save all of that inside the REDIS_USER variable and then substract `&auth[]=` in the other containers. But I suppose this is contra-intuitiv.

@derhagen WDYT what would be the best solution for you?

szaimen: Here the idea: https://github.com/nextcloud/all-in-one/pull/5870/commits/4acc9d8f41ad0bc3a22dad61286d096605413171. WDYT?

szaimen: > I guess in the helm chart values.yaml, we could still have a clean REDIS_USER variable that's later extended to something like REDIS_USER_AUTH="&auth[]=${REDIS_USER}", right?

Indeed :)

PR #5886, opened 2025-01-16
Added guide for setting up Nextcloud with FRP

xZinio:
I added a guide with example configuration files on how to setup Nextcloud with FRP (https://github.com/fatedier/frp). I was previously struggling with this exact setup, but got it to work using the help of the community:

https://help.nextcloud.com/t/setup-nextcloud-with-frp-fast-reverse-proxy/214793

The person helping me mentioned at the end I should share my findings and the working config files. 

Now, I would like to share the config files for FRPC and nginx that worked for me here alongside some explanations on how I set it up. I hope that this could be useful for others trying to achieve the same thing. Any suggestions on how to improve the configs or the explanations are very welcome!

### DIFF
--- a/reverse-proxy.md
+++ b/reverse-proxy.md
@@ -807,6 +807,72 @@ For a reverse proxy example guide for Tailscale, see this guide by @flll: https:
 
 </details>
 
+### Fast Reverse Proxy
+<details>
+	
+<summary>click here to expand</summary>
+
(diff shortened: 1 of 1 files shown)


szaimen:
Hi, thanks for your PR!

Unfortunately I am a bit confused why you need nginx in this chain? Isn't it possible to use FRP without using nginx in between FRP and AIO?

xZinio:
I was actually thinking the same thing. I the beginning I would always set 

`localAddr = "127.0.0.1:11000"`

as well as 

`localPort = 8080`

for the AIO (or just the port the AIO was running on. However, with this configuration I would usually run into a `502 Bad Gateway` issue in my browser. I'm also not the only since I found one other user having the exact same problem [here](https://github.com/nextcloud/all-in-one/discussions/611#discussioncomment-8865437).

Perhaps there is way to make it work without using nginx, but I was never able to figure out how. Maybe someone can adjust the guide in the future if someone ever figures it out. I just also found this way using nginx very useful since it was very easy to get my SSL certificates using certbot's nginx plugin this way.

PR #6022, opened 2025-02-10
Allow Jellyfin write access to NEXTCLOUD_MOUNT

Anvil5465:


community-containers/jellyfin/jellyfin.json
```diff
@@ -25,7 +25,7 @@
                 {
                     "source": "%NEXTCLOUD_MOUNT%",
                     "destination": "%NEXTCLOUD_MOUNT%",
-                    "writeable": false
+                    "writeable": true
```
AiroPi, line 28: I am not against this change since it can be useful for some users, but what are the use cases exactly ?

Is it to be able to delete media from the Jellyfin UI ?

AiroPi (approved):
LGTM, since some features of Jellyfin needs write accesses.

community-containers/jellyfin/jellyfin.json
Anvil5465: Yes deleting media is one feature, but also jellyfin offers the ability to save certain types of metadata (lyrics, subtitles, cover art, .nfo files) to the /media folder for easier management

szaimen:
Thanks! What about this writeable:false then? https://github.com/nextcloud/all-in-one/blob/ff7fdb5864ae15698c8edcff6d3243777904bbaf/community-containers/jellyfin/jellyfin.json#L23

AiroPi:
> Thanks! What about this writeable:false then? https://github.com/nextcloud/all-in-one/blob/ff7fdb5864ae15698c8edcff6d3243777904bbaf/community-containers/jellyfin/jellyfin.json#L23

Yes I am sorry I forget what was the purpose of NEXTCLOUD_MOUNT against NEXTCLOUD_DATA_DIR

Anvil5465:
@AiroPi @szaimen yes my thought there was just that my /media folder is on an external mount and so providing write access to just the folder needed and not the entire data directory would be safest. If someone later requests write access for the data directory, we can implement it. 

I'm of course open to setting the container to allow write access to both if you think that would be better

szaimen:
> @AiroPi @szaimen yes my thought there was just that my media is stored on an external mount and so providing write access to only the mount and not the entire data directory would be safest. If someone later requests write access for the data directory, we can implement it.
> 
> I'm of course open to setting the container to allow write access to both if you think that would be better

Okay, lets do it in a follow-up then 👍

community-containers/jellyfin/jellyfin.json
szaimen: > Yes deleting media is one feature, but also jellyfin offers the ability to save certain types of metadata (lyrics, subtitles, trickplay, cover art, .nfo files) to the /media folder for easier management

Wait a second, this would require the other varialbe 4 lines up as well to be set to true

szaimen (changes requested):

Anvil5465:
Sorry @szaimen I misspoke. I was just using /media loosely to mean the location where the media is stored, rather than the literal /media folder

szaimen:
> Sorry @szaimen I misspoke. I was just using /media loosely to mean the location where the media is stored, rather than the literal /media folder

all right :)

szaimen (approved):
LGTM then :)

szaimen:
This is now released with v10.6.0 Beta. Testing and feedback is welcome! See https://github.com/nextcloud/all-in-one#how-to-switch-the-channel

Issue #6014, opened 2025-02-05
Consider setting `set -eu` globally in nextcloud container entrypoint


joshtrichards:
The app container's entrypoint currently only uses `set` in a couple spots. We should test impact of enabling `set -eu` globally at the top, just like in the community micro-services image [here](https://github.com/nextcloud/docker/blob/8d2e9048bca45c0bbda8f36ba72dccc69937b393/docker-entrypoint.sh#L2). It should be safe to use alongside the existing ones already in the AIO entrypoint.

This should prevent problems that lead to broken installations.

This originally came up because there was a forum post a few weeks I saw where rsync failed, yet the failure was not detected and the installation proceeded.

> Ran across an in-the-wild environment where rsync was failing due to I/O disk errors, but initializing continued on anyhow. We should check the exit values and fail hard since it'll result in a broken installation (and may get overlooked until later depending on what files the errors occur on).
> 
> ```
> 2025-01-13T15:13:04.250546270Z Initializing nextcloud 30.0.4.1 ...
> 2025-01-13T15:13:13.543136088Z rsync: [sender] read errors mapping "/usr/src/nextcloud/apps/text/js/pgsql-6nMyK_Wd.chunk.mjs.map": I/O error (5)
> [...]
> 2025-01-13T15:13:13.791311797Z ERROR: apps/text/js/pgsql-6nMyK_Wd.chunk.mjs.map failed verification -- update discarded.
> [...]
> 2025-01-13T15:13:18.615971825Z rsync error: some files/attrs were not transferred (see previous errors) (code 23) at main.c(1338) [sender=3.3.0]
> 2025-01-13T15:13:18.903862977Z Initializing finished
> 2025-01-13T15:13:18.903916218Z New Nextcloud instance.
> 2025-01-13T15:13:18.913472340Z Installing with pgsql database
> 2025-01-13T15:13:18.913679471Z Starting Nextcloud installation...
> ```
>

szaimen:
Hey, I am honestly against adding this globally as things are not possible to easily anymore if we just exit anywhere. So I would only add this at places where it is necessary to run without any issue like for the rsync job. I created https://github.com/nextcloud/all-in-one/pull/6028. However it is not ready yet as we would probably need to think what to do if rsync actually fails: do we touch a file so that when the container restarts, we can print a specific warning or how do we handle this case?

szaimen:
@joshtrichards could you please lend me a hand here?

szaimen:
[...]

PR #6028, opened 2025-02-11
exit if rsync fails

szaimen:
Close https://github.com/nextcloud/all-in-one/issues/6014


Containers/nextcloud/entrypoint.sh
```diff
@@ -217,6 +217,8 @@ if ! [ -f "$NEXTCLOUD_DATA_DIR/skip.update" ]; then
         fi
 
         echo "Initializing nextcloud $image_version ..."
+        # TODO: touch rsync file to track what happens so that the container does not continue restarting? (it is not known what happens if it fails here at this step)
```
szaimen, line 220: @joshtrichards I need help with this. What should happen if we actually exit here? should we just touch a file so that we can track this and then report that rsync fails? or what should be done?

PR #6060, opened 2025-02-18
aio-interface: adjust applicable links to always open in a new tab

szaimen:


docjyJ (commented):
`rel="noopener"` does not seem required:

> Note: Setting target="_blank" on <a> elements implicitly provides the same rel behavior as setting [rel="noopener"](https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/rel/noopener) which does not set window.opener.
https://developer.mozilla.org/en-US/docs/Web/HTML/Element/a#target

szaimen:
> `rel="noopener"` does not seem required:
> 
> > Note: Setting target="_blank" on  elements implicitly provides the same rel behavior as setting [rel="noopener"](https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/rel/noopener) which does not set window.opener.
> > https://developer.mozilla.org/en-US/docs/Web/HTML/Element/a#target

Good catch, thanks! Just removed it :)

docjyJ (commented):
LGTM :heart:

szaimen:
> LGTM ❤️

Thanks for reviewing! ☺️

szaimen:
This is now released with v10.7.0 Beta. Testing and feedback is welcome! See https://github.com/nextcloud/all-in-one#how-to-switch-the-channel

PR #6094, opened 2025-02-26
clamav: build the container also for aarch64/arm64 by using the alpine package

Zoey2936:
- Needs https://github.com/nextcloud-releases/all-in-one/pull/57
- Close https://github.com/nextcloud/all-in-one/discussions/4223
- Close https://github.com/nextcloud/all-in-one/pull/4483
- Supercedes https://github.com/nextcloud/all-in-one/pull/4483

### DIFF
--- a/Containers/clamav/Dockerfile
+++ b/Containers/clamav/Dockerfile
@@ -9,7 +9,8 @@ RUN apk upgrade --no-cache -a && \
     sed -i "s|#\?MaxFileSize.*|MaxFileSize 2G|g" /etc/clamav/clamd.conf && \
     sed -i "s|#\?PCREMaxFileSize.*|PCREMaxFileSize aio-placeholder|g" /etc/clamav/clamd.conf && \
     sed -i "s|#\?StreamMaxLength.*|StreamMaxLength aio-placeholder|g" /etc/clamav/clamd.conf && \
-    sed -i "s|#\?TCPSocket|TCPSocket|g" /etc/clamav/clamd.conf
+    sed -i "s|#\?TCPSocket|TCPSocket|g" /etc/clamav/clamd.conf && \
+    freshclam --foreground --stdout
 
 COPY --chmod=775 start.sh /start.sh
 COPY --chmod=775 healthcheck.sh /healthcheck.sh
--- a/Containers/clamav/Dockerfile
+++ b/Containers/clamav/Dockerfile
@@ -5,7 +5,7 @@ RUN set -ex; \
     apk upgrade --no-cache -a; \
     apk add --no-cache tzdata clamav supervisord; \
     mkdir /run/clamav; \
-    chmod 777 -R /run/clamav /var/log/clamav; \
+    chmod 777 -R /run/clamav /var/log/clamav /var/log/supervisord /var/run/supervisord; \
     sed -i "s|#\?MaxDirectoryRecursion.*|MaxDirectoryRecursion 30|g" /etc/clamav/clamd.conf; \
     sed -i "s|#\?MaxFileSize.*|MaxFileSize 2G|g" /etc/clamav/clamd.conf; \
     sed -i "s|#\?PCREMaxFileSize.*|PCREMaxFileSize aio-placeholder|g" /etc/clamav/clamd.conf; \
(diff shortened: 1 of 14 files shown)


Zoey2936:
the only downside I see is that the alpine package may be a bit oudated compared to the direct image

Zoey2936 (commented):
also needs to be changed then in the release repo

szaimen:
Hi @Zoey2936, first of all thanks a lot for the PR! ☺️

szaimen:
> the only downside I see is that the alpine package may be a bit oudated compared to the direct image

this is fine for me

szaimen:
> also needs to be changed then in the release repo

Indeed. Would you mind adjusting the PR or shall I?

Containers/clamav/start.sh
```diff
@@ -0,0 +1,10 @@
+#!/usr/bin/env sh
+
+if ! freshclam --foreground --stdout; then
+  exit 1
+fi
+
+sed "s|aio-placeholder|$MAX_SIZE|" /etc/clamav/clamd.conf > /tmp/clamd.conf
+
+freshclam --foreground --stdout --daemon & 
+clamd --foreground --config-file=/tmp/clamd.conf
```
szaimen, line 10: Shouldnt we better use supervisord for that? Then it will make sure to automatically restart both services if they should be stopped :)

szaimen (commented):
We also need to adjust all places in the docs and AIO interface and so on. Can you handle this or shall I? :)

Containers/clamav/start.sh
Zoey2936: Thats how it is done in the start script of the official clamav image, but yes, feel free to edit this PR

Containers/clamav/start.sh
```diff
@@ -0,0 +1,10 @@
+#!/usr/bin/env sh
+
+if ! freshclam --foreground --stdout; then
+  exit 1
+fi
```
Zoey2936, line 5: this could also run inside the Dockerfile, which would make the docker image bigger, but the container may starts faster depending on the internet speed it has, but the pull would take longer.
This is needed to download the initial databases which will be updated by the freshclam deamon while running.

Containers/clamav/Dockerfile
```diff
@@ -1,28 +1,21 @@
...
+FROM alpine:3.21.3
 
-COPY clamav.conf /clamav.conf
-COPY --chmod=775 start.script /start.script
+RUN apk upgrade --no-cache -a && \
+    apk add --no-cache tzdata clamav && \
+    mkdir /run/clamav && \
+    chmod 777 -R /run/clamav /var/log/clamav && \
+    sed -i "s|#\?MaxDirectoryRecursion.*|MaxDirectoryRecursion 30|g" /etc/clamav/clamd.conf && \
+    sed -i "s|#\?MaxFileSize.*|MaxFileSize 2G|g" /etc/clamav/clamd.conf && \
```
szaimen, line 9:
```suggestion
    sed -i "s|#\?MaxFileSize.*|MaxFileSize aio-placeholder|g" /etc/clamav/clamd.conf && \
```

Containers/clamav/start.sh
szaimen: This is a very good idea! Then we don't need an external internet connection for clamav :)

Containers/clamav/Dockerfile
Zoey2936: the limit is 2G: https://github.com/Cisco-Talos/clamav/blob/492e505070499a5e7eb60afc89fdec42e55b1d8d/etc/clamd.conf.sample#L585

szaimen: ok

Containers/clamav/start.sh
szaimen: I mean to run it in the Dockerfile already :)

szaimen:
> > We also need to adjust all places in the docs and AIO interface and so on. Can you handle this or shall I? :)
> 
> can you please do this? Thanks!

done :)

Containers/clamav/start.sh
szaimen: done

szaimen (approved):
LGTM. Thanks a lot! ☺️

szaimen:
This is now released with v10.8.0 Beta. Testing and feedback is welcome! See https://github.com/nextcloud/all-in-one#how-to-switch-the-channel

Issue #5954, opened 2025-01-24
follow-up to allowing to add ui-secrets to the aio interface


szaimen:
Sounds good. Somme sugest:

- [x] 1. Add field with hide/show https://github.com/nextcloud/all-in-one/pull/6108
- [ ] 2. Add copy button (Ex: https://codepen.io/foxbeefly/pen/dyWbQgJ)

_Originally posted by @docjyJ in https://github.com/nextcloud/all-in-one/pull/5945#pullrequestreview-2570878000_

szaimen:
Hi @docjyJ are you ok if I close this for now? How important would be the copy button for you? In ny opinion it is not that important...

docjyJ:
Yes

PR #6108, opened 2025-03-01
Hide password by default

docjyJ:
Hi,

I found some time to do this. Now, the password is not displayed by default.

I didn't work on the design, but I moved the code to a components folder and used include with only from twig.
See: https://twig.symfony.com/doc/3.x/tags/include.html

image
image

Fix part of https://github.com/nextcloud/all-in-one/issues/5954

### DIFF
--- a/php/templates/components/container-state.twig
+++ b/php/templates/components/container-state.twig
@@ -16,7 +16,7 @@
   </span>
   {% if c.GetUiSecret() != '' %}
     <details>
-      <summary>Show password</summary>
+      <summary>Show password for {{ c.GetDisplayName() }}</summary>
       <p>{{ c.GetUiSecret() }}</p>
     </details>
   {% endif %}
--- a/php/templates/components/container-state.twig
+++ b/php/templates/components/container-state.twig
@@ -1,15 +1,19 @@
 {# @var c \App\Containers\Container #}
 <li>
-  {% if c.GetStartingState().value == 'starting' %}
-    <span class="status running"></span>
-  {% elseif c.GetRunningState().value == 'running' %}
-    <span class="status success"></span>
-  {% else %}
-    <span class="status error"></span>
-  {% endif %}
   <span>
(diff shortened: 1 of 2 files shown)


php/templates/components/container-state.twig
```diff
@@ -0,0 +1,23 @@
...
+  {% if c.GetStartingState().value == 'starting' %}
+    <span class="status running"></span>
+  {% elseif c.GetRunningState().value == 'running' %}
+    <span class="status success"></span>
+  {% else %}
+    <span class="status error"></span>
+  {% endif %}
+  <span>
+    {{ c.GetDisplayName() }}
+    (<a href="/api/docker/logs?id={{ c.GetIdentifier() }}" target="_blank">Stopped</a>)
```
szaimen, line 12: I would like to keep accessing the logs also when containers are running and starting

php/templates/components/container-state.twig
```diff
@@ -0,0 +1,23 @@
...
+    {% if c.GetDocumentation() != '' %}
+      (<a target="_blank" href="{{ c.GetDocumentation() }}">docs</a>)
+    {% endif %}
+  </span>
+  {% if c.GetUiSecret() != '' %}
+    <details>
+      <summary>Show password</summary>
```
szaimen, line 19:
```suggestion
      <summary>Show password for {{ c.GetDisplayName() }}</summary>
```

szaimen (changes requested):
Hi @docjyJ, thanks a lot for the PR! ☺️
I have two suggestions. See below:

php/templates/components/container-state.twig
docjyJ: My bad, fixed !

szaimen (approved):
LGTM, thanks

szaimen:
This is now released with v10.8.0 Beta. Testing and feedback is welcome! See https://github.com/nextcloud/all-in-one#how-to-switch-the-channel

PR #6173, opened 2025-03-14
Add Community Container documentation 

docjyJ:


szaimen:
> @szaimen WDYT ?

Looks already really cool, thanks for this initiative! 😊

Btw, may we should link this doc in this section https://github.com/nextcloud/all-in-one/tree/main/community-containers#how-to-add-containers and/or put it directly into the readme file as additonal section? Not sure what the best approach would be...

community-containers/helloworld/container-definition.md
```diff
@@ -0,0 +1,68 @@
...
+
+### Ports
+
+| Field         | Description                         |
+|---------------|-------------------------------------|
+| `ip_binding`  | The IP address to bind the port to. |
+| `port_number` | The port number to expose.          |
+| `protocol`    | The protocol to use.                |
+
+### Healthcheck
```
szaimen, line 51: We should probably state in the docs that the options here are not considered inside aio. Same for other things like profiles and user...

szaimen: So these options need to be set in the Dockerfile of the container...

szaimen:
> What elements are not supported?

The healthcheck options and some other things like profiles and user are not actually forwarded to the container. See https://github.com/nextcloud/all-in-one/discussions/5574#discussioncomment-11213399 so we should make that clear in the docs that they need to be set directly in the Dockerfile of the container instead...

Issue #6175, opened 2025-03-15
database import fails when pg_dump generates DOS formatted SQL script


kgraefe:
### Steps to reproduce
Run [Migrate the files and the database](https://github.com/nextcloud/all-in-one/blob/77dd56bade11cbb6ecffe8f2ada381520a16f5d6/migration.md#migrate-the-files-and-the-database), except:
1. To not pollute my server I ran the postgres server for the MySQL -> Postgres migration in a Docker container, using the `postgres:17.4`  image. I did the proper initialization, though:
```sh
❯ export PG_USER="ncadmin"
❯ export PG_PASSWORD="my-temporary-password"
❯ export PG_DATABASE="nextcloud_db"
❯ docker exec -i nextcloud-postgres psql -U postgres <<END
CREATE USER $PG_USER WITH PASSWORD '$PG_PASSWORD' CREATEDB;
CREATE DATABASE $PG_DATABASE WITH OWNER $PG_USER TEMPLATE template0 ENCODING 'UTF8';
GRANT ALL PRIVILEGES ON DATABASE $PG_DATABASE TO $PG_USER;
GRANT ALL PRIVILEGES ON SCHEMA public TO $PG_USER;
END
``` 
3. In step 4 I used `❯ docker exec -it nextcloud-postgres pg_dump -U postgres "$PG_DATABASE" > ./database-dump.sql` to dump the database
2. in step 12 I see lots of `ERROR:  role "ncadmin" does not exist` everywhere during database dump import

### Expected behavior 
The import should be successful.

### Actual behavior 
The import errors out because the database role does not exist.

#### Host OS 
Debian 12

#### Output of `sudo docker info`
<details>
<summary>sudo docker info</summary>

```sh
❯ sudo docker info
Client: Docker Engine - Community
 Version:    28.0.1
 Context:    default
 Debug Mode: false
 Plugins:
  compose: Docker Compose (Docker Inc.)
    Version:  v2.33.1
    Path:     /usr/libexec/docker/cli-plugins/docker-compose

Server:
 Containers: 11
  Running: 9
  Paused: 0
  Stopped: 2
 Images: 14
 Server Version: 28.0.1
 Storage Driver: overlay2
  Backing Filesystem: extfs
  Supports d_type: true
  Using metacopy: false
  Native Overlay Diff: true
  userxattr: false
 Logging Driver: json-file
 Cgroup Driver: systemd
 Cgroup Version: 2
 Plugins:
  Volume: local
[...]

Issue #5364, opened 2024-10-01
Update to postgres 17 with nc31


szaimen:
- [x] needs tests in nextcloud/server https://github.com/nextcloud/server/issues/48495
- [x] update docs: https://github.com/nextcloud/documentation/pull/12363
- [x] wait for 17.1
- [x] wait for nc 31 update
- [x] update in aio

Crash1602:
With NC 31.0.2, Postgres 17 is no longer reported as unsupported.

szaimen:
> With NC 31.0.2, Postgres 17 is no longer reported as unsupported.

Yes, I know :)

PR #6184, opened 2025-03-17
database-import: update logic to consider wrong line endings

szaimen:
- Needs https://github.com/nextcloud/all-in-one/issues/5364 for easy testing
- Close https://github.com/nextcloud/all-in-one/issues/6175#issuecomment-2729169366

### DIFF
--- a/Containers/postgresql/start.sh
+++ b/Containers/postgresql/start.sh
@@ -99,7 +99,7 @@ if ( [ -f "$DATADIR/PG_VERSION" ] && [ "$PG_MAJOR" != "$(cat "$DATADIR/PG_VERSIO
     fi
 
     # Get the Owner
-    DB_OWNER="$(grep -a "$GREP_STRING" "$DUMP_FILE" | head -1 | grep -oP 'Owner:.*$' | sed 's|Owner:||;s| ||g')"
+    DB_OWNER="$(grep -a "$GREP_STRING" "$DUMP_FILE" | head -1 | grep -oP 'Owner:.*$' | sed 's|Owner:||;s|[[:space:]]||g')"
     if [ "$DB_OWNER" = "$POSTGRES_USER" ]; then
         echo "Unfortunately was the found database owner of the dump file the same as the POSTGRES_USER $POSTGRES_USER"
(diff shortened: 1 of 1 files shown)


kgraefe:
Sorry, I got sick. Is there an easy way for me to test it? I would give it a try with my database export, if possible.

Containers/postgresql/start.sh
```diff
@@ -99,7 +99,7 @@ if ( [ -f "$DATADIR/PG_VERSION" ] && [ "$PG_MAJOR" != "$(cat "$DATADIR/PG_VERSIO
     fi
 
     # Get the Owner
-    DB_OWNER="$(grep -a "$GREP_STRING" "$DUMP_FILE" | head -1 | grep -oP 'Owner:.*$' | sed 's|Owner:||;s| ||g')"
+    DB_OWNER="$(grep -a "$GREP_STRING" "$DUMP_FILE" | head -1 | grep -oP 'Owner:.*$' | sed 's|Owner:||;s| ||g;s|\s||g')"
```
kgraefe, line 102: `\s` matches any whitespace. therefore you can remove `;s|| |g` in the `sed` command

kgraefe: also you may want to use `[[:space:]]` which is POSIX, whereas `\s` is only GNU according to [stackexchange](https://unix.stackexchange.com/a/583886/52916) (I could not find it in the manuals)

this works against my local database export:
```sh
❯ grep "Name: oc_appconfig; Type: TABLE; Schema: public; Owner:" database-dump.sql.bak | grep -oP 'Owner:.*$' | sed 's|Owner:||;s|[[:space:]]||g' | hexdump -C 
00000000  6e 63 61 64 6d 69 6e 0a                           |ncadmin.|
00000008
```

szaimen: Thanks for the idea! I just used `[[:space:]]` in the replace command. Can you please check again? :)

kgraefe (approved):

szaimen:
This is now released with v11.0.0 Beta. Testing and feedback is welcome! See https://github.com/nextcloud/all-in-one#how-to-switch-the-channel

PR #6226, opened 2025-03-26
Update reverse-proxy.md

DaCHack:
Adding prominent hint to Traefik users adding the Traefik container to the aio network to change the server domain during the installation process. The nextcloud-aio-apache container is not in place as indicated in the documentation at this point in time

### DIFF
--- a/reverse-proxy.md
+++ b/reverse-proxy.md
@@ -707,6 +707,7 @@ The examples below define the dynamic configuration in YAML files. If you rather
         nextcloud:
           loadBalancer:
             servers:
+    		#When adding Traefik to the aio network: This needs to be changed in the installation process! Point this to nextcloud-aio-domaincheck first, then change to nextcloud-aio-apache after the domaincheck
               - url: "http://localhost:11000" # Adjust to match APACHE_PORT and APACHE_IP_BINDING. See https://github.com/nextcloud/all-in-one/blob/main/reverse-proxy.md#adapting-the-sample-web-server-configurations-below
 
       middlewares:


reverse-proxy.md
```diff
@@ -707,6 +707,7 @@ The examples below define the dynamic configuration in YAML files. If you rather
         nextcloud:
           loadBalancer:
             servers:
+    		#When adding Traefik to the aio network: This needs to be changed in the installation process! Point this to nextcloud-aio-domaincheck first, then change to nextcloud-aio-apache after the domaincheck
               - url: "http://localhost:11000" # Adjust to match APACHE_PORT and APACHE_IP_BINDING. See https://github.com/nextcloud/all-in-one/blob/main/reverse-proxy.md#adapting-the-sample-web-server-configurations-below
```
szaimen, line 711: Hi, if you read through the documentation mentioned here: https://github.com/nextcloud/all-in-one/blob/main/reverse-proxy.md#adapting-the-sample-web-server-configurations-below, you would have seen that pointing at nextcloud-aio-apache is the solution to your problem

DaCHack: It was actually the Root cause of my problem. Apache does not get created at the Point of the domaincheck. Thus, nextcloud-aio-apache cannot be found by Traefik.

szaimen: This should not be a problem as the domaincheck container uses an alias: https://github.com/nextcloud/all-in-one/blob/13e9829a850ae27258b7e05f54849a3d33371cdb/php/src/Docker/DockerActionManager.php#L945-L948

DaCHack: Interesting! Seems this did not work for me. Traefik always reported there is no such host.

flaevers:
Similar issue for me, seemed like.
https://github.com/nextcloud/all-in-one/blob/main/reverse-proxy.md#adapting-the-sample-web-server-configurations-below 
`http://nextcloud-aio-apache.nextcloud-aio:$APACHE_PORT` (e.g. `http://nextcloud-aio-apache.nextcloud-aio:1234`) didn't work for me, I continued with domain validation skipped.

After install, the domain name I had didn't work (bad gateway). I changed it to `url: "http://nextcloud-aio-apache:$APACHE_PORT` - this made the chosen (sub)domain successfully fwd to the nextcloud install.

PR #6231, opened 2025-03-26
clamav: use correct clamd.conf in freshclam

Zoey2936:
From freshclam --help:
--daemon-notify[=/path/clamd.conf]   Send RELOAD command to clamd


szaimen (approved):
LGTM, thanks ☺️

PR #6235, opened 2025-03-26
link to borg logs while borg is running

Zoey2936:
Not sure if mastercontainer is wanted, but I think it should link to borgs logs

### DIFF
--- a/php/templates/containers.twig
+++ b/php/templates/containers.twig
@@ -60,7 +60,7 @@
             {% endfor %}
 
             {% if is_daily_backup_running == true %}
-                <p><span class="status running"></span> Daily backup currently running. (<a href="/api/docker/logs?id=nextcloud-aio-borgbackup" target="_blank">Logs</a>)</p>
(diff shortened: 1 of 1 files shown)


php/templates/containers.twig
```diff
@@ -60,7 +60,7 @@
             {% endfor %}
 
             {% if is_daily_backup_running == true %}
-                <p><span class="status running"></span> Daily backup currently running. (<a href="/api/docker/logs?id=nextcloud-aio-mastercontainer" target="_blank">Logs</a>)</p>
+                <p><span class="status running"></span> Daily backup currently running. (<a href="/api/docker/logs?id=nextcloud-aio-borgbackup" target="_blank">Logs</a>)</p>
```
szaimen, line 63: yes, actually the mastercontainer was the intended as it prints the current state to the logs. Maybe we can do something like this?
```suggestion
                <p><span class="status running"></span> Daily backup currently running. (<a href="/api/docker/logs?id=nextcloud-aio-mastercontainer" target="_blank">Mastercontainer logs</a>) (<a href="/api/docker/logs?id=nextcloud-aio-borgbackup" target="_blank">Borg backup container logs</a>)</p>
```

Zoey2936: For me it was a bit confusing, since it talks about the backup container, but shows me the mastercontainer logs, which are only partly useful

szaimen (approved):
LGTM, thanks!

szaimen:
This is now released with v10.11.0 Beta. Testing and feedback is welcome! See https://github.com/nextcloud/all-in-one#how-to-switch-the-channel

PR #6236, opened 2025-03-26
helm-chart: move alpine init-container image to containers folder

szaimen:
- For https://github.com/nextcloud-releases/all-in-one/pull/65

### DIFF
--- a/Containers/alpine/Dockerfile
+++ b/Containers/alpine/Dockerfile
@@ -1,0 +1,5 @@
+# syntax=docker/dockerfile:latest
+FROM alpine:3.21
+
+RUN set -ex; \
+    apk upgrade --no-cache -a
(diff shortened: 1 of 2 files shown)


Containers/alpine/Dockerfile
```diff
@@ -0,0 +1,5 @@
+# syntax=docker/dockerfile:latest
+FROM alpine:3.21
```
Zoey2936, line 2: (I know unrelated to this PR, but maybe we should change all alpine:3.21.3 base images to alpine:3.21, since we run apk upgrade in any alpine Dockerfile so it will use the latest patch in any case)

szaimen: Good idea! Lets do this in a follow-up :)

PR #6299, opened 2025-04-14
Update reverse-proxy.md for Collabora Compatibility

Zade222:
Added 'location ^~ /*' lines for collabora compatibility as the current config does now allow for collabora use and just results in a time out.

There may be some oddities so I encourage discussing what's been added to drill down to a proper solution for everyone to use. I mainly got to this point after arguing with Google Gemini 2.5 Pro for a fair bit of time but ultimately got collabora working.

### DIFF
--- a/reverse-proxy.md
+++ b/reverse-proxy.md
@@ -420,6 +420,63 @@ server {
 
     server_name <your-nc-domain>;
 
+    # Static files for Collabora Online
+    location ^~ /browser {
+        proxy_pass http://127.0.0.1:11000;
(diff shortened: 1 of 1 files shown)


Zoey2936 (changes requested):
works for me without this change

Zadeis:
> works for me without this change

Does it? For me the current config results in a timeout. Is your nginx proxy on a separate device or local to the nextcloud host?

Zoey2936:
are you sure that you correctly configured the websocket mappings?

PR #6308, opened 2025-04-16
feat(docker-socket-proxy): allow inspect image rule

andrey18106:
Allow `/images/{name}/json`.

Ref: https://github.com/nextcloud/docker-socket-proxy/pull/47, required for: https://github.com/nextcloud/app_api/pull/554


oleksandr-nc (approved):
here is not rush, we need this for Nextcloud 32.

PR #6311, opened 2025-04-17
Docker-socket-proxy: Update haproxy.cfg to fix timeouts

derigs:
Due to some testing its possible to install the agent via occ and the frontend by setting this on 60s. Lower values didn't work in my testing.

### DIFF
--- a/Containers/docker-socket-proxy/haproxy.cfg
+++ b/Containers/docker-socket-proxy/haproxy.cfg
@@ -4,9 +4,9 @@ global
     maxconn 10
 
 defaults
-    timeout connect 10s
-    timeout client 10s
-    timeout server 10s
+    timeout connect 60s
+    timeout client 60s
+    timeout server 60s
 
 frontend http
     mode http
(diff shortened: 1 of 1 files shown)


Containers/docker-socket-proxy/haproxy.cfg
```diff
@@ -4,9 +4,9 @@ global
     maxconn 10
 
 defaults
-    timeout connect 10s
-    timeout client 10s
-    timeout server 10s
+    timeout connect 60s
+    timeout client 60s
+    timeout server 60s
```
oleksandr-nc, line 9: lets put here 1800s (this is needed for context chat in some cases)

oleksandr-nc:
> Lower values didn't work in my testing.

So value of `30s` like [here](https://github.com/nextcloud/docker-socket-proxy/blob/9d09e28612b15ac24ec171b21a95a309d15b79b3/Dockerfile#L12-L14) did not work, right?

If it is so, then we need to update those calues to 60s there too, to keep consistency between AIO and usual DSP.

oleksandr-nc (approved):

szaimen:
This is now released with v10.13.0 Beta. Testing and feedback is welcome! See https://github.com/nextcloud/all-in-one#how-to-switch-the-channel

PR #6333, opened 2025-04-23
Fix typo

benwaffle:
I think this was the intended word? There's probably a way to rewrite this line to make it clearer


szaimen (approved):
Thanks! :)

PR #6343, opened 2025-04-27
Added proper cloudflare proxy guide to reverse-proxy.md

MorrowShore:


Zoey2936 (changes requested):
you edited the cloudflare tunnel guide and added instruction on how to use cloudflare (not cloudflare tunnels) as reverse proxy, it should at least be a different guide, but I'm against adding something like since this would mean that the connection between AIO and cloudflare is fully unencrypted (and being send unencrypted through the internet). Also some screenshots are in the wrong order

MorrowShore:
There you go.
Added as a separate guide and fixed the screenshot order.
It's useful / advantageous because it can be set up without any server side configuration, and gives access to Cloudflare security features.
We're doing this until we figure out the openlitespeed situation.
Although... flexible SSL may not be needed?

Zoey2936:
I still don't think that this is a good idea, since the connections between cloudflare and aio is unencrypted/cleartext but goes over the public internet, but I don't decide if this will be merged, but if it will be merged, then there should be a big note about this problem at the beginning

szaimen (changes requested):
I agree with Zoey here. We should not merge this as it might be a security risk.
However feel free to create a new guide for this at https://github.com/nextcloud/all-in-one/discussions/new?category=wiki

szaimen:
Documented at https://github.com/nextcloud/all-in-one/discussions/6350

PR #6374, opened 2025-05-02
helm: automatically enforce RPSS if configured

szaimen:


nextcloud-aio-helm-chart/templates/nextcloud-aio-namespace-namespace.yaml
```diff
@@ -4,4 +4,8 @@ kind: Namespace
 metadata:
   name: "{{ .Values.NAMESPACE }}"
   namespace: "{{ .Values.NAMESPACE }}"
+  {{- if eq (.Values.RPSS_ENABLED | default "no") "yes" }}
+  labels:
+    pod-security.kubernetes.io/enforce: restricted
+  {{- end }}
```
szaimen, line 10: this needs to be added via update-helm.sh

PR #6443, opened 2025-05-26
aio-interface: allow to manage the community containers via the AIO interface

szaimen:
Close https://github.com/nextcloud/all-in-one/issues/5937

### DIFF
--- a/php/psalm-baseline.xml
+++ b/php/psalm-baseline.xml
@@ -20,6 +20,11 @@
       <code><![CDATA[Container]]></code>
     </ClassMustBeFinal>
   </file>
+  <file src="src/Container/CommunityContainer.php">
+    <ClassMustBeFinal>
+      <code><![CDATA[CommunityContainer]]></code>
+    </ClassMustBeFinal>
+  </file>
   <file src="src/Container/ContainerEnvironmentVariables.php">
     <ClassMustBeFinal>
       <code><![CDATA[ContainerEnvironmentVariables]]></code>
--- a/php/psalm-baseline.xml
+++ b/php/psalm-baseline.xml
@@ -1,86 +1,17 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <files psalm-version="6.11.0@4ed53b7ccebc09ef60ec4c9e464bf8a01bfd35b0">
-  <file src="src/Auth/AuthManager.php">
-    <ClassMustBeFinal>
-      <code><![CDATA[AuthManager]]></code>
-    </ClassMustBeFinal>
-  </file>
-  <file src="src/Auth/PasswordGenerator.php">
-    <ClassMustBeFinal>
-      <code><![CDATA[PasswordGenerator]]></code>
-    </ClassMustBeFinal>
-  </file>
-  <file src="src/Container/AioVariables.php">
-    <ClassMustBeFinal>
-      <code><![CDATA[AioVariables]]></code>
-    </ClassMustBeFinal>
-  </file>
-  <file src="src/Container/Container.php">
-    <ClassMustBeFinal>
-      <code><![CDATA[Container]]></code>
-    </ClassMustBeFinal>
-  </file>
-  <file src="src/Container/CommunityContainer.php">
-    <ClassMustBeFinal>
-      <code><![CDATA[CommunityContainer]]></code>
-    </ClassMustBeFinal>
-  </file>
-  <file src="src/Container/ContainerEnvironmentVariables.php">
-    <ClassMustBeFinal>
-      <code><![CDATA[ContainerEnvironmentVariables]]></code>
-    </ClassMustBeFinal>
-  </file>
-  <file src="src/Container/ContainerPort.php">
-    <ClassMustBeFinal>
-      <code><![CDATA[ContainerPort]]></code>
-    </ClassMustBeFinal>
-  </file>
-  <file src="src/Container/ContainerPorts.php">
-    <ClassMustBeFinal>
-      <code><![CDATA[ContainerPorts]]></code>
-    </ClassMustBeFinal>
-  </file>
-  <file src="src/Container/ContainerVolume.php">
-    <ClassMustBeFinal>
-      <code><![CDATA[ContainerVolume]]></code>
-    </ClassMustBeFinal>
-  </file>
-  <file src="src/Container/ContainerVolumes.php">
-    <ClassMustBeFinal>
-      <code><![CDATA[ContainerVolumes]]></code>
-    </ClassMustBeFinal>
-  </file>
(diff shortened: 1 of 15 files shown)


docjyJ:
I propose a solution

php/psalm-baseline.xml
```diff
@@ -20,6 +20,11 @@
       <code><![CDATA[Container]]></code>
     </ClassMustBeFinal>
   </file>
+  <file src="src/Container/CommunityContainer.php">
+    <ClassMustBeFinal>
+      <code><![CDATA[CommunityContainer]]></code>
+    </ClassMustBeFinal>
```
szaimen, line 26: What would we need to do to fix this psalm issue in the code?

php/public/options-form-submit.js
```diff
@@ -22,7 +24,9 @@ document.addEventListener("DOMContentLoaded", function () {
         });
 
         // Show or hide submit button based on changes
-        optionsFormSubmit.style.display = hasChanges ? 'block' : 'none';
+        optionsFormSubmit.forEach(optionsFormSubmit => {
+            optionsFormSubmit.style.display = hasChanges ? 'block' : 'none';
```
szaimen, line 28: I think this would work incorrectly as it would enable both buttons whenever one of the checkboxes in either form is clicked. How can we fix this?

php/src/Data/ConfigurationManager.php
```diff
@@ -1008,16 +1009,46 @@ public function isBackupSectionEnabled() : bool {
...
+            return $cc;
+        }
+        foreach ($dir as $id) {
+            $json = json_decode(DataConst::GetCommunityContainersDirectory() . '/' . $id . '/' . $id . '.json');
+            if (is_string($json['aio_services_v1'][0]['display_name'])
+                && is_string($json['aio_services_v1'][0]['description'])) {
+                $cc[] = new CommunityContainer(
+                    $id,
+                    $json['aio_services_v1'][0]['display_name'],
+                    $json['aio_services_v1'][0]['documentation']);
```
szaimen, line 1035: in principle, it would also be possible to have two containers with a display name inside the array. So it is not always `[0]`. How could we make sure that it works always?

szaimen (commented):
Hey @docjyJ thanks a lot for helping me with this 🤗

I have a few remarks and questions below. Would you be able to help me to figure out a solution for these? :)

php/src/Data/ConfigurationManager.php
docjyJ: I agree. Currently, there is no other way to get the title of the community containers. We would need to find a method to define the title to display.

php/psalm-baseline.xml
szaimen: okay, so maybe we should add this to the global ignore list to not spam the baseline?

php/templates/includes/community-containers.twig
```diff
@@ -0,0 +1,34 @@
+<h2>Community Containers</h2>
+<p>In this section you can enable or disable optional community containers that are not included by default in the main installation. These containers are provided by the community and can be useful for various purposes.</p>
```
szaimen, line 2: I think we should add a warning here that people need to read the docs before adding any cc. Otherwise I fear people might just add everything and break their instance...

php/templates/containers.twig
```diff
@@ -574,6 +574,7 @@
                     {% if is_backup_container_running == false %}
 
                         {{ include('includes/optional-containers.twig') }}
+                        {{ include('includes/community-containers.twig') }}
```
szaimen, line 577: can we maybe move the section below the timezone change section? I fear people might miss the timezone option otherwise...

php/templates/includes/community-containers.twig
docjyJ: Yes

docjyJ: Maybe we can use a hiden section collapsed

php/templates/containers.twig
docjyJ: Done 9b6508d

php/templates/includes/community-containers.twig
```diff
@@ -0,0 +1,38 @@
+<h2>Community Containers</h2>
+<p>In this section you can enable or disable optional Community Containers that are not included by default in the main installation. These containers are provided by the community and can be useful for various purposes.</p>
+<p><strong>Caution: </strong>Community Containers are not officially component and it maintained by the community. Some containers may not be compatible with your system, may not work as expected or may discontinue. Use them at your own risk.</p>
```
szaimen, line 3: Maybe we could add a hint here to read first through the documentation of each container before adding any...

php/public/options-form-submit.js
szaimen: found a solution

php/templates/includes/community-containers.twig
szaimen: done

php/src/Data/ConfigurationManager.php
docjyJ: Done d2cab4a

szaimen:
Thanks @docjyJ! :) Can you please check my last commit if that makes sense?

php/src/Data/ConfigurationManager.php
```diff
@@ -1008,16 +1009,59 @@ public function isBackupSectionEnabled() : bool {
...
+    public function listAvailableCommunityContainers() : array {
+        $cc = [];
+        $dir = scandir(DataConst::GetCommunityContainersDirectory());
+        if ($dir === false) {
+            return $cc;
+        }
+        foreach ($dir as $id) {
+            $filePath = DataConst::GetCommunityContainersDirectory() . '/' . $id . '/' . $id . '.json';
+            $fileContents = apcu_fetch($filePath);
+            if ($fileContents === false || !is_string($fileContents)) {
```
docjyJ, line 1031:
```suggestion
            if (!is_string($fileContents)) {
```
false is not a string

php/src/Data/ConfigurationManager.php
```diff
@@ -1008,16 +1009,53 @@ public function isBackupSectionEnabled() : bool {
...
+        }
+
+        return $config['aio_community_containers'];
+    }
+
+
+    /** @return list<CommunityContainer> */
+    public function listAvailableCommunityContainers() : array {
+        $cc = [];
+        $dir = scandir(DataConst::GetCommunityContainersDirectory());
```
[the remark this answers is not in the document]
docjyJ: lgtm

docjyJ (approved):

php/src/Data/ConfigurationManager.php
szaimen: you are right

PR #6496, opened 2025-06-02
rp-docs: add Traefik v3

QuentinLemCode:
Hello

This PR update the reverse-proxy documentation to remove the experimental flag HTTP3 from the Traefik configuration.
This flag is obsolete and will prevent Traefik from starting. HTTP3 is enabled just with the `http3` property under `entrypoints.https`

Source : https://doc.traefik.io/traefik/migration/v2-to-v3-details/#http3

### DIFF
--- a/reverse-proxy.md
+++ b/reverse-proxy.md
@@ -745,6 +745,89 @@ The examples below define the dynamic configuration in YAML files. If you rather
 
 </details>
 
+### Traefik 3
+
+<details>
+
+<summary>click here to expand</summary>
+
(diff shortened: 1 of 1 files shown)


szaimen:
Hi, thanks for your PR! I am not sure if this is correct though since our docs is for Treafik v2. However the docs you link state that http3 is only for Traefik v3 no longer experimental... Would you be able to add an additional entry for Traefik v3 or adjust and test our docs for Traefik v3?

QuentinLemCode:
@szaimen Sure, here is the updated version 👌

PR #6523, opened 2025-06-10
Remove libretranslate

docjyJ:


szaimen:
Hey, I fear this does not work like that... In best case we simply remove the wholte directory but in that case we need to adjust the logic internally to ignore apps that are enabled but not existing on the file system...

szaimen:
@docjyJ we could maxbe adjust the display name to `LibreTranslate (deprecated)` in https://github.com/nextcloud/all-in-one/blob/d922017777ef81b95ea5485eab14c9e4f871dcf9/community-containers/libretranslate/libretranslate.json#L5
WDYT about the idea? :)

docjyJ:
Yes why not

docjyJ:
Done

community-containers/libretranslate/libretranslate.json
```diff
@@ -2,7 +2,7 @@
     "aio_services_v1": [
         {
             "container_name": "nextcloud-aio-libretranslate",
-            "display_name": "LibreTranslate",
+            "display_name": "LibreTranslate - DEPRECATED",
```
szaimen: I am not sure about the caps lock... What about this wording? :)

```suggestion
            "display_name": "LibreTranslate (deprecated)",
```

docjyJ: ( and ) are not valide char :/

szaimen: lets add them then :)

szaimen: done

szaimen (approved):
LGTM, thanks! ☺️

szaimen:
This is now released with v11.3.0 Beta. Testing and feedback is welcome! See https://github.com/nextcloud/all-in-one#how-to-switch-the-channel

PR #6533, opened 2025-06-12
Update watchtower to a well-maintained repo and add podman compatibility

apparle:
Since watchtower original repo seems to be abandoned even for security updates (see https://github.com/containrrr/watchtower/issues/2067), this PR shifts to using a better maintained fork (see https://github.com/nicholas-fedor/watchtower/discussions/267#discussioncomment-13201594).
Additionally, it also adds an environment variable to get watchtower updates working in podman containerization environments.


szaimen:
Hi, thank you for your PR! I am fine with this change but have some suggestion below:

Can you please change the build process and use `FROM ghcr.io/nicholas-fedor/watchtower:1.11.3` directly? Then dependabot which is already configured for the Dockerfile will take care of the updates to the image automatically and we decrease the build time and make the Dockerfile simpler. (and we don't need to add another workflow that updates the version of the image manually)...

Thanks in advance! :)

apparle:
> Can you please change the build process and use `FROM ghcr.io/nicholas-fedor/watchtower:1.11.3` directly? Then dependabot which is already configured for the Dockerfile will take care of the updates to the image automatically and we decrease the build time and make the Dockerfile simpler. (and we don't need to add another workflow that updates the version of the image manually)...

Done

szaimen (approved):
Thank you, LGTM! 

I will also try to include this one in the next release after AIO v11.1.0 stable :)

docjyJ:
nice

invis-z:
Since this PR mentions podman, I would like to ask whether it would be a good idea to add an option to disable watchtower all together? With podman, there is already a way to keep containers up to date. 

AFAIK, watchtower is only used to update the master container? All the other container updates are handled by the master container itself. 

Adding this as an option seems also simplify the support of a registry mirror as watchtower is not supporting it.

apparle:
> Since this PR mentions podman, I would like to ask whether it would be a good idea to add an option to disable watchtower all together? With podman, there is already a way to keep containers up to date.

I'll recommend this be a separate issue.
Nextcloud AIO has been strongly opinionated on NOT allowing mechanisms to disable updates / skip updates, because that opens the door to fraken-updated set ups and many support tickets that come with it.


> AFAIK, watchtower is only used to update the master container? All the other container updates are handled by the master container itself.

While podman does have it's own auto-update, I don't think that's a good idea in AIO setup . AIO container tries to be graceful -- stop all containers, do a backup, then attempt to update the master container, then update all the subcontainers, which guarantees consistency of updates as well as consistent backup to rollback to, if needed. Podman's auto-update would be a side-band restart of just the mastercontainer which may lead to inconsistencies (unless you're doing clever tricks with systemd services to recreate the graceful behavior). 


> Adding this as an option seems also simplify the support of a registry mirror as watchtower is not supporting it.

Regarding mirrors, have you tried modifying `/etc/containers/registries.conf` (root containers) or `~/.config/containers/registries.conf` (rootless) for podman ?
From my experiments in past, podman does allow for clever ways to redirect / alias in registries (eg: see https://github.com/nextcloud/all-in-one/discussions/5994#discussioncomment-12337032), and then anyone using podman socket just gets the right aliasing behavior. And I'm guessing that'll include watchtower as well, once this PR gets merged.

apparle:
FWIW, just modifying the script to disable the updates is not the hard part - that can be easily done in mastercontainer's codebase itself or in the watchtower script by detecting that container is running under podman and container has `io.containers.autoupdate=registry` set on it. 
But whether to do it -- that's a different conversation that should be in a different issue.

invis-z:
Thanks for the explanation. The part about mirror is based on a previous discussion where docker instead of podman is used, I didn't know about all the podman tricks so thanks for pointing me to the right direction. 

I was wondering why aio uses watchtower the way it does, and thought it would be easily replaceable. Wanting to gracefully doing updates after backups totally makes sense and I no longer think the option to disable is useful.

szaimen:
This is now released with v11.2.0 Beta. Testing and feedback is welcome! See https://github.com/nextcloud/all-in-one#how-to-switch-the-channel

PR #6544, opened 2025-06-13
borg-test: config file is not required for restoring

szaimen:
Close https://github.com/nextcloud/all-in-one/discussions/6538
Waiting for feedback here: https://github.com/nextcloud/all-in-one/discussions/6538#discussioncomment-13459122


szaimen:
It seems like a valid config file is actually required. So the check seems legit

PR #6550, opened 2025-06-15
feat(helm): allow custom data storage class for Nextcloud data PVC

frederikb96:
New Helm value `STORAGE_CLASS_DATA`

Summary
- allow specifying `STORAGE_CLASS_DATA` for the `nextcloud-aio-nextcloud-data` PVC
- update templates and generation script to support the new value
- document the new value in `values.yaml`
- Bump helm version


nextcloud-aio-helm-chart/Chart.yaml
```diff
@@ -1,6 +1,6 @@
 name: nextcloud-aio-helm-chart
 description: A generated Helm Chart for Nextcloud AIO from Skippbox Kompose
-version: 11.0.0
+version: 11.1.0
```
szaimen: please revert to 11.0.0. This will be done autoamtically when 11.1.0 becomes stable
```suggestion
version: 11.0.0
```

nextcloud-aio-helm-chart/update-helm.sh
```diff
@@ -407,6 +411,7 @@ sed -i 's|17179869184|"17179869184"|' /tmp/sample.conf
 echo "" >> /tmp/sample.conf
 # shellcheck disable=SC2129
 echo 'STORAGE_CLASS:        # By setting this, you can adjust the storage class for your volumes' >> /tmp/sample.conf
+echo 'STORAGE_CLASS_DATA:        # Allows to set a dedicated storage class for the Nextcloud data volume' >> /tmp/sample.conf
```
szaimen: Maybe we should make the description a bit better by saying the `STORAGE_CLASS` should be a fast storage and `STORAGE_CLASS_DATA` can be a bit slower?

szaimen (commented):
Hi @frederikb96 thanks a lot for the PR!

I was wondering if it makes sense to allow to do the same for the `nextcloud-aio-database-dump-persistentvolumeclaim` and `nextcloud-aio-talk-recording-persistentvolumeclaim` volumes? WDYT?

Also some small suggestions below:

frederikb96:
Hi :) Ah yes I guess that makes sense. Just realized that I will not have time for this the next days. So feel free to adjust as you like, or maybe I will come back to it at a later point :)

nextcloud-aio-helm-chart/Chart.yaml
szaimen: done

nextcloud-aio-helm-chart/update-helm.sh
szaimen: done

szaimen (approved):
LGTM, thank you! :)

szaimen:
This is now released with v11.1.0 of the helm chart. Testing and feedback is welcome! See https://github.com/nextcloud/all-in-one/releases/tag/helm-chart-11.1.0

PR #6574, opened 2025-06-22
Make daily-backup.sh more robust by checking for uninitialized state

apparle:
Adding a guard check in daily-backup.sh to detect uninitialized state to make it more robust. 
This is continuation with the discussion on https://github.com/nextcloud/all-in-one/discussions/5994#discussioncomment-12335738

**Background summary**
Podman containers have native integration with systemd with the ability to start/stop/restart services using `systemctl start XXXX.service` at reboots, other service interactions etc. Because AIO container itself is just the wrapper, there's a need for programmatic mechanism to do "Start Containers" or "Stop Containers"  similar to the buttons exposed in the WebUI of mastercontainer; and then the systemd service can use them for graceful start/stop of whole of AIO+Nextcloud. The `daily-backup.sh` script is quite useful for this purpose:
- Start command: `podman exec -it --env START_CONTAINERS=1 nextcloud-aio-mastercontainer /daily-backup.sh`
- Stop command: `podman exec -it --env STOP_CONTAINERS=1 nextcloud-aio-mastercontainer /daily-backup.sh`

This has been working working quite well - been using it for ~6 months with once-a-week reboots.
But as pointed out in earlier discussion, above will reach undefined behavior if someone reboots the machine / restarts the service on a fresh installation with configuration not yet done.
This PR improves that by handling this corner case and exiting early.

PS: Using `grep` to do the check is an implementation choice. Another alternative would be to add `jq` package to mastercontainer image and use that to parse configuration.json more reliably.

### DIFF
--- a/Containers/mastercontainer/daily-backup.sh
+++ b/Containers/mastercontainer/daily-backup.sh
@@ -3,9 +3,9 @@
 echo "Daily backup script has started"
 
 # Check if initial configuration has been done, otherwise this script should do nothing.
-configFile=/mnt/docker-aio-config/data/configuration.json
-if [ ! -f "$configFile" ] || ! grep -q -E '"wasStartButtonClicked"\s*:\s*1\s*,' "$configFile"; then
-    echo "Initial configuration not done yet. Exiting..."
+CONFIG_FILE=/mnt/docker-aio-config/data/configuration.json
+if ! [ -f "$CONFIG_FILE" ] || ! grep -q "wasStartButtonClicked.*1" "$CONFIG_FILE"; then
+    echo "Initial configuration via AIO interface not done yet. Exiting..."
     exit 0
 fi
 
--- a/Containers/mastercontainer/daily-backup.sh
+++ b/Containers/mastercontainer/daily-backup.sh
@@ -2,6 +2,13 @@
 
 echo "Daily backup script has started"
 
+# Check if initial configuration has been done, otherwise this script should do nothing.
+configFile=/mnt/docker-aio-config/data/configuration.json
+if [ ! -f "$configFile" ] || ! grep -q -E '"wasStartButtonClicked"\s*:\s*1\s*,' "$configFile"; then
+    echo "Initial configuration not done yet. Exiting..."
+    exit 0
+fi
+
 # Daily backup and backup check cannot be run at the same time
 if [ "$DAILY_BACKUP" = 1 ] && [ "$CHECK_BACKUP" = 1 ]; then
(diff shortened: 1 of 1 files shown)


Containers/mastercontainer/daily-backup.sh
```diff
@@ -2,6 +2,13 @@
 
 echo "Daily backup script has started"
 
+# Check if initial configuration has been done, otherwise this script should do nothing.
+configFile=/mnt/docker-aio-config/data/configuration.json
+if [ ! -f "$configFile" ] || ! grep -q -E '"wasStartButtonClicked"\s*:\s*1\s*,' "$configFile"; then
+    echo "Initial configuration not done yet. Exiting..."
```
szaimen:
```suggestion
CONFIG_FILE=/mnt/docker-aio-config/data/configuration.json
if ! [ -f "$CONFIG_FILE" ] || ! grep -q "wasStartButtonClicked.*1" "$CONFIG_FILE"; then
    echo "Initial configuration via AIO interface not done yet. Exiting..."
```

szaimen (commented):
Hi, thanks for the PR! 

I have some small suggestions regarding the syntax below. 

I hope the suggestions make sense to you!

Containers/mastercontainer/daily-backup.sh
```diff
@@ -2,6 +2,13 @@
...
+    exit 0
```
szaimen, line 9: btw, should we exit 1 here? (I mean we do not handle exit codes in the wrapper script yet but it still might make sense. WDYT?

apparle: Using a non-zero exit code will actually create problems for my intended usage because systemd does interpret non-zero error code as failure of service and kill the process.  

At the moment I'd say, leave it at 0.
Down the road if Nextcloud AIO project wants to add first-class support for podman+systemd, I'll recommend we peel this out into it's own `start_containers.sh` and `stop_containers.sh` that have more information reported to logs, systemd notification support (either through systemd socket + [systemd-notify](https://www.freedesktop.org/software/systemd/man/latest/systemd-notify.html) and graceful handling of in-progress updates & backups.

szaimen: Ok

szaimen (approved):
LGTM, thanks!

szaimen:
This is now released with v11.2.1 Beta. Testing and feedback is welcome! See https://github.com/nextcloud/all-in-one#how-to-switch-the-channel

PR #6592, opened 2025-06-30
readme: improve the `How to stop/start/update containers or trigger the daily backup from a script externally?` section

thiswillbeyourgithub:
Signed-off-by: thiswillbeyourgithub <26625900+thiswillbeyourgithub@users.noreply.github.com>

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -1035,7 +1035,7 @@ You can do so by running the `/daily-backup.sh` script that is stored in the mas
 - `DAILY_BACKUP` if set to `1`, it will automatically stop the containers and create a backup. If you want to start them again afterwards, you may have a look at the `START_CONTAINERS` option.
 - `STOP_CONTAINERS` if set to `1`, it will automatically stop the containers at the start of the script. Implied by `DAILY_BACKUP=1`.
 - `START_CONTAINERS` if set to `1`, it will automatically start the containers at the end of the script, without updating them. Implied by `DAILY_BACKUP=1`.
-- `CHECK_BACKUP` if set to `1`, it will start the borg backups integrity check. Note that the backup check is non blocking so containers can be kept running while the check lasts. That means you can't pass `DAILY_BACKUP=1` at the same time. The output of the check can be found in the logs of the container `nextcloud-aio-borgbackup`.
+- `CHECK_BACKUP` if set to `1`, it will start the integrity check of all borg backups made by AIO. Note that the backup check is non blocking so containers can be kept running while the check lasts. That means you can't pass `DAILY_BACKUP=1` at the same time. The output of the check can be found in the logs of the container `nextcloud-aio-borgbackup`.
 
 One example to do a backup would be `sudo docker exec -it --env DAILY_BACKUP=1 nextcloud-aio-mastercontainer /daily-backup.sh`, which you can run via a cronjob or put it in a script.
 


readme.md
```diff
@@ -1034,10 +1034,12 @@ You can do so by running the `/daily-backup.sh` script that is stored in the mas
...
 - `DAILY_BACKUP` if set to `1`, it will automatically stop the containers and create a backup. If you want to start them again afterwards, you may have a look at the `START_CONTAINERS` option.
 - `START_CONTAINERS` if set to `1`, it will automatically start the containers without updating them.
-- `STOP_CONTAINERS` if set to `1`, it will automatically stop the containers.
-- `CHECK_BACKUP` if set to `1`, it will start the backup check. This is not allowed to be enabled at the same time like `DAILY_BACKUP`. Please be aware that this option is non-blocking which means that the backup check is not done when the process is finished since it only start the borgbackup container with the correct configuration.
+- `STOP_CONTAINERS` if set to `1`, it will automatically stop the containers, either during a backup or during a check (even though the backup check is non blocking).
```
szaimen: Not sure why you adjusted this...
```suggestion
- `STOP_CONTAINERS` if set to `1`, it will automatically stop the containers.
```

readme.md
```diff
@@ -1034,10 +1034,12 @@ You can do so by running the `/daily-backup.sh` script that is stored in the mas
...
+- `CHECK_BACKUP` if set to `1`, it will start the borg backups integrity check. Note that the backup check is non blocking so containers can be kept running while the check lasts. That means you can't pass `DAILY_BACKUP=1` at the same time. The output of the check can be found in the logs of the container `nextcloud-aio-borgbackup`.
```
```suggestion
- `CHECK_BACKUP` if set to `1`, it will start the borg backup integrity check. Note that the backup check is non blocking so containers can be kept running while the check lasts. That means you can't pass `DAILY_BACKUP=1` at the same time. The output of the check can be found in the logs of the container `nextcloud-aio-borgbackup`.
```

szaimen (commented):
Hi, thanks for the PR! Just a few comments below

readme.md
thiswillbeyourgithub: The fact that checking the backup is non blocking does not mean that the containers are not stopped by default, nor updated, nor restarted. So to me it's worth clarifying that the STOP_CONTAINERS has an effect both on creating backups and on checking backups.

readme.md
thiswillbeyourgithub: Well we are checking the backups made by borg, not checking the integrity of the backup system called "borg backup". As the package in the ppa is called "borgbackup" I think it would be worth being precise.

All of this is of course obvious for people that know what borg is etc, but slightly less noob friendly. And the concept of "data integrity checking" is not entirely accessible to rookies. IMHO of course.

thiswillbeyourgithub:
I replied, but if you are still not convinced I'm fine with committing your changes.

readme.md
szaimen: Actually checking the backup integrity should not automatically stop the containers. This is why I find the information confusing here...

thiswillbeyourgithub: Oh I understand more now. I pushed something more then. Tell me what you think

readme.md
szaimen: I mean this sentence and information makes sense. I was only trying to fix a typo (removing the s here)...

thiswillbeyourgithub: My point is that it's not a typo but a plural marker.

thiswillbeyourgithub:
All good?

readme.md
szaimen: I see. Maybe the following wording is better then?
```suggestion
- `CHECK_BACKUP` if set to `1`, it will start the integrity check of all borg backups made by AIO. Note that the backup check is non blocking so containers can be kept running while the check lasts. That means you can't pass `DAILY_BACKUP=1` at the same time. The output of the check can be found in the logs of the container `nextcloud-aio-borgbackup`.
```

thiswillbeyourgithub: I agree thanks. For some reason I can't apmrove so I pushed it myself. I hope that's okay.

szaimen (approved):
LGTM, thanks! ☺️

PR #6606, opened 2025-07-02
Make elasticsearch connection configurable

thegcat:
We would like to use an external ElasticSearch server and would need some configurability, especially wrt user and index names for the ElasticSearch connection (we currently have amongst others 2 nextcloud-aio that should be consolidated to one ES server).

This PR makes those connection settings configurable for advanced users without changing the defaults and current settings for basic users and helm users.


szaimen:
Hi, before I can review this, please make sure to fix DCO as we cannot merge this in the current state. See https://github.com/nextcloud/all-in-one/pull/6606/checks?check_run_id=45202785412

thegcat:
Done.

szaimen (approved):
LGTM, thank you!

thegcat:
Thanks for catching the variables, it seems I was not careful enough 😅 And thanks for merging!

szaimen:
This is now released with v11.5.0 Beta. Testing and feedback is welcome! See https://github.com/nextcloud/all-in-one#how-to-switch-the-channel

thegcat:
Oh, I did not think of reporting back, but we backported this to our slightly modified branch and everything works well as far as we can tell.

PR #6628, opened 2025-07-07
rp-docs: increase Treafik timeout 

seclution:
Increase timeouts to prevent connection reset on uploads >100MB in clean Traefik setup

### DIFF
--- a/reverse-proxy.md
+++ b/reverse-proxy.md
@@ -765,6 +765,9 @@ The examples below define the dynamic configuration in YAML files. If you rather
     entryPoints:
       https:
         address: ":443" # Create an entrypoint called "https" that uses port 443
+          transport:
+            respondingTimeouts:
+              readTimeout: 30m # Allows uploads > 100MB; prevents connection reset due to chunking (public upload-only links)
         # If you want to enable HTTP/3 support, uncomment the line below
         # http3: {}
     
--- a/reverse-proxy.md
+++ b/reverse-proxy.md
@@ -678,6 +678,9 @@ The examples below define the dynamic configuration in YAML files. If you rather
     entryPoints:
       https:
         address: ":443" # Create an entrypoint called "https" that uses port 443
+        transport:
+          respondingTimeouts:
+            readTimeout: 24h # Allows uploads > 100MB; prevents connection reset due to chunking (public upload-only links)
         # If you want to enable HTTP/3 support, uncomment the line below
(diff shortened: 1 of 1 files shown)


reverse-proxy.md
```diff
@@ -765,6 +765,9 @@ The examples below define the dynamic configuration in YAML files. If you rather
     entryPoints:
       https:
         address: ":443" # Create an entrypoint called "https" that uses port 443
+          transport:
+            respondingTimeouts:
+              readTimeout: 30m # Allows uploads > 100MB; prevents connection reset due to chunking (public upload-only links)
```
szaimen: usually we put a timeout of 24h. See for example https://github.com/seclution/all-in-one/blob/99b54546761f69b61c9d1236373afee8087c7934/reverse-proxy.md?plain=1#L531C20-L531C26

seclution: Thanks for the addition!

In our case, a 30-minute timeout was sufficient, but higher values like 24h should also work without issues. I intentionally chose a shorter duration to avoid potential side effects from long-lived open connections, although I can't assess the exact impact this might have on Traefik.

I also assume that Traefik v2 requires a similar configuration, but I’m unable to test this myself. Maybe this note could be added to the v2 section for completeness?

szaimen: I just checked and it looks like traefik v2.1. supports the same config. See https://doc.traefik.io/traefik/v2.1/routing/entrypoints/#respondingtimeouts

szaimen: Btw, I checked https://doc.traefik.io/traefik/v3.0/routing/entrypoints/#respondingtimeouts and it looks like the indentation is not correct...

```suggestion
        transport:
          respondingTimeouts:
            readTimeout: 30m # Allows uploads > 100MB; prevents connection reset due to chunking (public upload-only links)
```

szaimen (changes requested):

reverse-proxy.md
szaimen: also, please adjust the timeout to 24h:
```suggestion
              readTimeout: 24h # Allows uploads > 100MB; prevents connection reset due to chunking (public upload-only links)
```

szaimen: done

szaimen: done

szaimen: done

PR #6648, opened 2025-07-11
mastercontainer: add check for http proxy variables

szaimen:


munderl:
Is there a reason for this change? We rely on this because our nextcloud sever is not allowed to access the internet directly but over a proxy only, where all connections are checked by our data center for abnormalities. Currently we have to keep the master container on the older version to keep the cloud running.

raspitoaster:
Same here - if there are no issues known and the change is just to clarify that the environment variables are not supported, then why not raise a warning at startup instead of an "exit 1"? The merged commit will break existing installations that use the vars.

PR #6685, opened 2025-07-25
daily-backup: implement a dedicated imagepull before stopping containers which should reduce the downtime

szaimen:
Close https://github.com/nextcloud/all-in-one/discussions/4501

### DIFF
--- a/php/src/Cron/PullContainerImages.php
+++ b/php/src/Cron/PullContainerImages.php
@@ -1,0 +1,20 @@
+<?php
+declare(strict_types=1);
+
+// increase memory limit to 2GB
+ini_set('memory_limit', '2048M');
+
+// Log whole log messages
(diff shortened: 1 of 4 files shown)


php/src/Cron/PullContainerImages.php
```diff
@@ -0,0 +1,20 @@
...
+
+require __DIR__ . '/../../vendor/autoload.php';
+
+$container = \AIO\DependencyInjection::GetContainer();
+
+/** @var \AIO\Controller\DockerController $dockerController */
+$dockerController = $container->get(\AIO\Controller\DockerController::class);
+
+// Pull all containers
+$dockerController->PullAllContainerImage();
```
szaimen:
```suggestion
$dockerController->PullAllContainerImages();
```

PR #6687, opened 2025-07-25
feat(borgbackup): add support for custom remote path

bwcknr:
Enable configuration of BORG_REMOTE_PATH environment variable for borgbackup operations. This allows users to specify a custom path to the borg executable on remote systems, which is useful when borg is installed in non-standard locations.


Zoey2936:
see: https://github.com/nextcloud/all-in-one/discussions/6340
I already asked if a PR like yours would be merged, and the answer was sadly no.

If you created this PR because of rsync.net, you can just write them an E-Mail, they will add the borg command back for you

bwcknr:
> see: #6340 I already asked if a PR like yours would be merged, and the answer was sadly no.
> 
> If you created this PR because of rsync.net, you can just write them an E-Mail, they will add the borg command back for you

I have a different opinion on that topic as it is not a 'problem with the one provider' but merely a basic borg feature.
Please see https://github.com/nextcloud/all-in-one/discussions/6340#discussioncomment-13889069

Zoey2936:
> I have a different opinion on that topic as it is not a 'problem with the one provider' but merely a basic borg feature.
Please see https://github.com/nextcloud/all-in-one/discussions/6340#discussioncomment-13889069

~~And why should your opinion change anything here?~~ I [already] asked if something like this would be merged, and it was denied[.]~~, so what do you expect?~~ Even if I would like to see this merged, what changed? Why should the answer now be yes? The reason to add this is even less important now that rsync will readd the borg command back if you ask.

Also borg has many options and still nextcloud-aio will never support all of them, so what is the argument here? [Multiple Usecases would be one, from my view]

Edit: Sorry this messages sounded rude, this was not intended, I edited it a bit to sound less rude

bwcknr:
> > I have a different opinion on that topic as it is not a 'problem with the one provider' but merely a basic borg feature.
> > Please see [#6340 (reply in thread)](https://github.com/nextcloud/all-in-one/discussions/6340#discussioncomment-13889069)
> 
> And why should your opinion change anything here? I asked if something like this would be merged, and it was denied, so what do you expect? Even if I would like to see this merged, what changed? Why should the answer now be yes? The reason to add this is even less important now that rsync will readd the borg command back if you ask.
> 
> Also borg has many options and still nextcloud-aio will never support all of them, so what is the argument here?

@szaimen is this how the community works here?

Zoey2936:
sorry if I maybe sound a bit rude, but my questions still stands, what has changed since my question if a PR like this would be merged was denied, except that such a PR is not required anymore to use rsync. I don't think “there is an option so it needs to be provided” is a big argument.

Edit: You don't need to convince me, I just want to tell you that without new/different arguments, your chance is low that this will be merged

Containers/borgbackup/start.sh
```diff
@@ -18,6 +18,9 @@ else
 fi
 export BORG_UNKNOWN_UNENCRYPTED_REPO_ACCESS_IS_OK=yes
 export BORG_RELOCATED_REPO_ACCESS_IS_OK=yes
+if [ -n "$BORG_REMOTE_PATH" ]; then
+    export BORG_REMOTE_PATH
+fi
```
Zoey2936, line 23: I think this can be removed, since the env is not modified here, it does not need to re-exported

Zoey2936, line 23: I think this can be removed, since the env is not modified here, it does not need to re-exported, but I'm not fully sure

php/tests/tests/restore-instance.spec.js
```diff
@@ -53,6 +57,8 @@ test('Restore instance', async ({ page: setupPage }) => {
   await containersPage.locator('#borg_restore_host_location').fill(borgBackupLocation);
   await containersPage.locator('#borg_restore_password').click();
   await containersPage.locator('#borg_restore_password').fill(borgBackupPassword);
+  // Clear remote path field for local backup test
+  await containersPage.locator('input[name="borg_remote_path"]').fill('');
```
Zoey2936, line 61: It is unclear, are they required to work or was this only added for local testing? (comment applies to all three changes in this file)

php/templates/containers.twig
```diff
@@ -417,6 +419,7 @@
                                             <form method="POST" action="/api/configuration" class="xhr">
                                                 <label>Local backup location</label> <input type="text" name="borg_backup_host_location" placeholder="/mnt/backup"/><br>
                                                 <label>Remote borg repo</label> <input type="text" name="borg_remote_repo" placeholder="ssh://user@host:port/path/to/repo"/><br>
+                                                <label>Remote borg path</label> <input type="text" name="borg_remote_path" placeholder="borg1"/><br>
```
Zoey2936, line 422: I think it should be made more clear that these options are optional to set, or the placeholder should default to borg instead of borg1, or both (comment applies to all three changes in this file)

bwcknr:
> sorry if I maybe sound a bit rude, but my questions still stands, what has changed since my question if a PR like this would be merged was denied, except that such a PR is not required anymore to use rsync. I don't think “there is an option so it needs to be provided” is a big argument.

Sorry, but my question for clarification also still stands @szaimen

I maybe just need to reevaluate if contributing and using would be beneficial in my contexts.

szaimen:
Hi, I might reconsider the decision if there are multiple providers that require this setting.

For now rsync.net seems to be the only one and even for that a feasible workaround is available.

HHousen:
@szaimen This is also an issue for Hetzner Storage Boxes. Hetzner defaults to using borg 1.2.x while Nextcloud AIO uses 1.4.x. It seems like this setup technically works, but it isn't ideal to be using a different minor version locally vs remotely. Hetnzer also recommends using `--remote-path=borg-1.4` to choose a borg version: https://docs.hetzner.com/storage/storage-box/access/access-ssh-rsync-borg/#borgbackup

So, rsync.net requires `remote-path` since they don't provide the `borg` command by default and Hetzner recommends it so that the remote repo is on the same minor version as locally.  So, not necessarily a problem with more than one provider, but it certainly would help with more than one provider.

Zoey2936:
https://github.com/borgbackup/borg/discussions/8621#discussioncomment-11735759 the borg dev says that it is not a problem to use diffrent minor versions if the major version is the same. So currently it should make no problems, but once borg2 is out this is probably needed

HHousen:
> [borgbackup/borg#8621 (comment)](https://github.com/borgbackup/borg/discussions/8621#discussioncomment-11735759) the borg dev says that it is not a problem to use diffrent minor versions if the major version is the same. So currently it should make no problems, but once borg2 is out this is probably needed

Nice! Good to know the main dev says that rather than it just seeming to work.

Mycah:
I don't understand why adding this command line isn't possible when there is already documentation on how to update the retention period with some env changes.  Why can't the same method be used to update the remote borg command?

https://github.com/nextcloud/all-in-one#how-to-adjust-borgs-retention-policy

PR #6688, opened 2025-07-25
feat(borgbackup): add support for bind-mounted SSH keys

bwcknr:
Enable users to provide existing SSH keys for remote borgbackup repositories by bind mounting them to the mastercontainer. Keys mounted at id_borg_mount and id_borg_mount.pub are automatically copied to the expected locations on startup, allowing reuse of existing SSH credentials.

### DIFF
--- a/Containers/mastercontainer/start.sh
+++ b/Containers/mastercontainer/start.sh
@@ -331,6 +331,27 @@ chown www-data:www-data -R /mnt/docker-aio-config/session/
 chown www-data:www-data -R /mnt/docker-aio-config/caddy/
 chown root:root -R /mnt/docker-aio-config/certs/
 
+# Handle SSH key propagation for borgbackup
+BORG_KEY_PATH="/mnt/docker-aio-config/data/id_borg"
+BORG_KEY_PUB_PATH="/mnt/docker-aio-config/data/id_borg.pub"
+BORG_KEY_MOUNT_PATH="/mnt/docker-aio-config/data/id_borg_mount"
+BORG_KEY_PUB_MOUNT_PATH="/mnt/docker-aio-config/data/id_borg_mount.pub"
+
+# Check if borg keys don't exist and mounted keys do exist, then copy them
+if [ ! -f "$BORG_KEY_PATH" ] && [ -f "$BORG_KEY_MOUNT_PATH" ]; then
+    print_green "Copying mounted SSH private key for borgbackup..."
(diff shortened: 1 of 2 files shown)


Zoey2936 (changes requested):
I think it would be much simpler to just tell the users to which volume/folder they need to upload the file and which permissions it needs. Additional mounts and script changes make this just more complex than it needs to be.

wickner-consid:
> I think it would be much simpler to just tell the users to which volume/folder they need to upload the file and which permissions it needs. Additional mounts and script changes make this just more complex than it needs to be.

The whole point of configurable containerized applications is to avoid manual intervention - asking users to manually upload files and set permissions after deployment goes against core container principles. This bind-mount approach ensures declarative configuration where everything is set up correctly at startup without any manual steps. I believe eliminating operational overhead for users justifies the modest implementation complexity on your side.

Zoey2936:
somehow the commit order is wrong, I replied to this comment: https://github.com/nextcloud/all-in-one/pull/6688#issuecomment-3118169390

> The whole point of configurable containerized applications is to avoid manual intervention - asking users to manually upload files and set permissions after deployment goes against core container principles. This bind-mount approach ensures declarative configuration where everything is set up correctly at startup without any manual steps. I believe eliminating operational overhead for users justifies the modest implementation complexity on your side.

Which is still required if the ssh keys needs to be changed after they are once copied/generated

(and just for you, I don't decide here anything, I speak for my own, the decisions will be made by szaimen)

bwcknr:
> (and just for you, I don't decide here anything, I speak for my own, the decisions will be made by szaimen)

"nextcloud/all-in-one"-your codebase. No need to get personal here

Zoey2936:
(I wrote that because you wrote “on your side”, I just wanted to make this clear for you, that I speak for me, not for nextcloud/all-in-one)

bwcknr:
> I think it would be much simpler to just tell the users to which volume/folder they need to upload the file and which permissions it needs. Additional mounts and script changes make this just more complex than it needs to be.

The whole point of configurable containerized applications is to avoid manual intervention - asking users to manually upload files and set permissions after deployment goes against core container principles. This bind-mount approach ensures declarative configuration where everything is set up correctly at startup without any manual steps. I believe eliminating operational overhead for users justifies the modest implementation complexity on your side.

szaimen:
Hi, I agree with Zoey here that we should rather document how to add your own key into the container at the correct place e.g. via `docker cp`.

PR #6702, opened 2025-07-29
Add support for no_pull_container_ids param for easier local build and testing

asavageiv:
This allows easier local testing of the containers because you can build locally with the correct label then start the containers again.

For example:
1. Stop the containers via the AIO admin UI
2. Build apache with `git push && docker buildx build -t ghcr.io/nextcloud-releases/aio-apache:beta --build-arg AIO_GIT_URL="https://github.com/asavageiv/all-in-one.git" --build-arg AIO_GIT_BRANCH="$(git branch --show-current)" --load --no-cache Containers/apache/`
3. Go to the admin UI with no_pull_container_ids set, like this: `/containers?no_pull_container_ids=nextcloud-aio-apache`
4. Push start containers

Now you will be running with the locally built apache image.

You can adjust the tag on the docker build command for `:beta` or `:latest` as needed.


Zoey2936:
What is the difference to https://github.com/nextcloud/all-in-one/pull/5568/files?

asavageiv:
> What is the difference to https://github.com/nextcloud/all-in-one/pull/5568/files?

#5568 is only for the mastercontainer. This works for all the other containers. If you want to make changes locally to both the mastercontainer and the others then you would use both of them together.

Zoey2936:
maybe then also add some documentation for this, like done in #5568?

asavageiv:
> maybe then also add some documentation for this, like done in #5568?

Done. I noticed that in the code the `container_names` are referred to as `id`s. Everything should probably consistently call them `names` matching the Docker CLI and the key in `containers.json`, but that is a separate refactor that might touch a lot of files. Should I open an issue for it?

asavageiv:
Actually I think this needs more work. Not all the docker files have the `AIO_GIT_URL` and `AIO_GIT_BRANCH` args.

Zoey2936:
> Actually I think this needs more work. Not all the docker files have the `AIO_GIT_URL` and `AIO_GIT_BRANCH` args.

yes, since they are not required since they are built based on the local files and don't clone anything while building

asavageiv:
Looks like the `nextcloud/Dockerfile` also needed it. Added the fix

Zoey2936:
right, I'm not sure why it is even cloned, since it could also be directly copied, but yes

asavageiv:
> right, I'm not sure why it is even cloned, since it could also be directly copied, but yes

I was wondering that too. I can think of two possible reasons:
1. We want to always build against the "latest" image of the repo for some reason. This is odd because usually we want everything to be as it is locally so it's all the same "version".
2. The Dockerfile was written with `Containers/nextcloud` as the build context and it was easier to clone the repo than add a symlink or rewrite the Dockerfile with a different build context.

The initial clone was added 3 years ago in #2! Maybe szaiman remembers why.

Containers/nextcloud/Dockerfile
```diff
@@ -1,6 +1,9 @@
 # syntax=docker/dockerfile:latest
 FROM php:8.3.23-fpm-alpine3.22
 
+ARG AIO_GIT_URL="https://github.com/nextcloud-releases/all-in-one.git"
+ARG AIO_GIT_BRANCH="main"
```
szaimen: 2. Can you please move these args 7 lines down into the AIO settings block?

php/public/index.php
```diff
@@ -86,6 +86,10 @@
     // Check if bypass_mastercontainer_update is provided on the URL, a special developer mode to bypass a mastercontainer update and use local image.
     $params = $request->getQueryParams();
     $bypass_mastercontainer_update = isset($params['bypass_mastercontainer_update']);
+    $no_pull_container_ids = '';
+    if (isset($params['no_pull_container_ids'])) {
+        $no_pull_container_ids = $params['no_pull_container_ids'];
+    }
```
szaimen: 3. Can we maybe simplify this by introducing one setting that bypasses all container pulls? 

e.g.:
```suggestion
    $do_not_pull_sibling_containers = isset($params['do_not_pull_sibling_containers']);
```

szaimen (changes requested):
Hi, thanks for the PR and sorry for needing so long to review this!

Some comments:
1. Can you please fix DCO? See https://github.com/nextcloud/all-in-one/pull/6702/checks?check_run_id=47130206913

szaimen:
> 2\. The Dockerfile was written with `Containers/nextcloud` as the build context and it was easier to clone the repo than add a symlink or rewrite the Dockerfile with a different build context.

Indeed, this is the reason. See https://github.com/nextcloud-releases/all-in-one/blob/main/.github/workflows/build_images.yml

php/src/Controller/DockerController.php
```diff
@@ -19,12 +19,12 @@ public function __construct(
     ) {
     }
 
-    private function PerformRecursiveContainerStart(string $id, bool $pullImage = true) : void {
+    private function PerformRecursiveContainerStart(string $id, bool $pullImage = true, array $noPullContainerIds = []) : void {
```
szaimen: If we simplify the logic (3.), we can also simplify the logic in this file

php/templates/containers.twig
```diff
@@ -352,6 +352,7 @@
                                     <form method="POST" action="/api/docker/start" class="xhr">
                                         <input type="hidden" name="{{csrf.keys.name}}" value="{{csrf.name}}">
                                         <input type="hidden" name="{{csrf.keys.value}}" value="{{csrf.value}}">
+                                        <input type="hidden" name="no_pull_container_ids" value="{{no_pull_container_ids}}">
```
szaimen: the name needs to be adjusted then here as well

develop.md
```diff
@@ -57,3 +57,15 @@ docker buildx build -t ghcr.io/nextcloud-releases/all-in-one:latest --build-arg
...
+## How to locally build and test changes to other containers using the no_pull_container_ids param
+1. Push changes to your own git fork and branch
+1. Use below commands to build the container image for a custom git url and branch:
+```
+cd Containers/{container}
+docker buildx build -t ghcr.io/nextcloud-releases/all-in-one:latest --build-arg AIO_GIT_URL="https://github.com/my-fork-repo/all-in-one.git" --build-arg AIO_GIT_BRANCH="my-feature-branch" --load .
+```
+1. Stop the containers using the AIO admin interface
+1. Since the hash of a locally built image doesn't match the latest release. Reload the AIO admin interface with the param `no_pull_container_ids={container_id}`, e.g. `?no_pull_container_ids=nextcloud-aio-apache,nextcloud-aio-nextcloud`.
+  - `container_id` matches the `container_name` from `containers.json`
```
szaimen: this docs need to be adjusted then as well. For example it should be `docker buildx build -t ghcr.io/nextcloud-releases/{container}:develop --build-arg AIO_GIT_URL="https://github.com/my-fork-repo/all-in-one.git" --build-arg AIO_GIT_BRANCH="my-feature-branch" --load .`

Probably we should also add a hint that the build-args are only needed for the nextcloud container?

asavageiv:
> Indeed, this is the reason. See https://github.com/nextcloud-releases/all-in-one/blob/main/.github/workflows/build_images.yml

Are you open to a PR that rewrites them all to use the project root as the build context and avoid the clone? I'd love to be able to develop locally without pushing my in-flight work and take advantage of Docker build step caching. We could get rid of the build args completely that way.

php/public/index.php
asavageiv: Done

develop.md
asavageiv: Done

szaimen:
> Are you open to a PR that rewrites them all to use the project root as the build context and avoid the clone?

In general yes but see https://github.com/nextcloud-releases/all-in-one/pull/72#issuecomment-3191368656

asavageiv:
Ok, I pulled the Dockerfile changes into this PR which now Depends on https://github.com/nextcloud-releases/all-in-one/pull/72 for updating the `build_images` workflow.

I think this is ready to commit once that other PR goes in.

php/templates/containers.twig
```diff
@@ -352,6 +352,7 @@
                                     <form method="POST" action="/api/docker/start" class="xhr">
                                         <input type="hidden" name="{{csrf.keys.name}}" value="{{csrf.name}}">
                                         <input type="hidden" name="{{csrf.keys.value}}" value="{{csrf.value}}">
+                                        <input type="hidden" name="bypass_container_update" value="{{bypass_container_update}}">
```
szaimen, line 355: I think this change here and the additions in index.php can be removed since the change in Dockercontroller.php should already make it work, no?

asavageiv: I think they're necessary to forward the query param given in GET `/containers?bypass_container_update` to the JSON body of the POST to `/api/containers/start`, right? The query params aren't forwarded automatically I don't think.

develop.md
```diff
@@ -48,12 +48,26 @@ This is documented here: https://github.com/nextcloud-releases/all-in-one/tree/m
...
+1. Ensure you are on the developer channel per the instructions above.
+1. Use the commands below from the project root to build the container image:
+```
+# For the "nextcloud" container
+docker buildx build --file Containers/nextcloud/Dockerfile --tag ghcr.io/nextcloud-releases/aio-nextcloud:develop --load .
+
+# For all other containers
+docker buildx build --file Containers/{container}/Dockerfile --tag ghcr.io/nextcloud-releases/aio-{container}:develop --load Containers/{container}
+```
+    - Note that the build-args are only needed for the `apache` and `nextcloud` containers
```
szaimen:
```suggestion
```

php/templates/containers.twig
szaimen: Looks like you are right. Just tested and confirmed 👍

php/src/Controller/DockerController.php
```diff
@@ -191,8 +191,9 @@ public function StartContainer(Request $request, Response $response, array $args
         $config['install_latest_major'] = $installLatestMajor;
         $this->configurationManager->WriteConfig($config);
 
+        $pullImage = !isset($request->getParsedBody()['bypass_container_update']);
```
szaimen, line 196:
```suggestion
        // Do not pull container images in case 'bypass_container_update' is set via url params
        // Needed for local testing
        $pullImage = !isset($request->getParsedBody()['bypass_container_update']);
        if ($pullImage === false) {
            error_log('WARNING: Not pulling container images. Instead, using local ones.');
        }
```

szaimen (approved):
LGTM! Thank you! ☺️

szaimen:
This is now released with v11.7.0 Beta. Testing and feedback is welcome! See https://github.com/nextcloud/all-in-one#how-to-switch-the-channel

PR #6676, opened 2025-07-19
Configure WOPI requests to remain within the Docker network

asavageiv:
Depends on https://github.com/nextcloud/all-in-one/pull/6568

This PR configures Collabora and Nextcloud to use the docker hostnames when communicating internally.

This fixes a few issues when AIO is behind a reverse proxy on another server outside of the docker network/LAN/default WOPI allow list.
- It increases performance saving a potentially expensive round trip
- It avoids needing the AIO host server's (possibly dynamic) public IP in the WOPI allow list

I think that it is fine for the Collabora <-> Nextcloud communication to be over HTTP and not HTTPS because it is within the docker network.

Before:
```
2025-06-20T03:52:43.330179992Z Activating Collabora config...
2025-06-20T03:52:43.901663506Z ✓ Reset callback url autodetect
2025-06-20T03:52:43.901688553Z Checking configuration
2025-06-20T03:52:43.901693541Z 🛈 Configured WOPI URL: https://nc.mydomain.me
2025-06-20T03:52:43.901697511Z 🛈 Configured public WOPI URL: https://nc.mydomain.me
2025-06-20T03:52:43.901701373Z 🛈 Configured callback URL: 
2025-06-20T03:52:43.901705189Z 
2025-06-20T03:52:44.622263890Z ✓ Fetched /hosting/discovery endpoint
2025-06-20T03:52:44.623698285Z ✓ Valid mimetype response
2025-06-20T03:52:44.624908960Z ✓ Valid capabilities entry
2025-06-20T03:52:44.897462696Z ✓ Fetched /hosting/capabilities endpoint
2025-06-20T03:52:44.897794791Z ✓ Detected WOPI server: Collabora Online Development Edition 25.04.2.2
2025-06-20T03:52:44.903921218Z 
2025-06-20T03:52:44.903961154Z Collabora URL (used for Nextcloud to contact the Collabora server):
2025-06-20T03:52:44.903975151Z   https://nc.mydomain.me
2025-06-20T03:52:44.903986670Z Collabora public URL (used in the browser to open Collabora):
2025-06-20T03:52:44.903998135Z   https://nc.mydomain.me
2025-06-20T03:52:44.904009056Z Callback URL (used by Collabora to connect back to Nextcloud):
2025-06-20T03:52:44.904020350Z   autodetected (will use the same URL as your user for browsing Nextcloud)
```

After:
```
Activating Collabora config...
✓ Set WOPI url to http://nextcloud-aio-collabora:9980
[...]

PR #6703, opened 2025-07-29
add CGNAT range to Caddy and document support for additional trusted proxies in rp docs

asavageiv:
When a reverse proxy is on a remote host that proxies from a public IP to a VPN IP (e.g. a Tailscale node) then the VPN IP needs to be a trusted proxy to correctly forward the client IP. This permits that.

### DIFF
--- a/reverse-proxy.md
+++ b/reverse-proxy.md
@@ -1003,12 +1003,10 @@ Add the IP it uses connect to AIO to the Nextcloud trusted_proxies like this:
 sudo docker exec --user www-data -it nextcloud-aio-nextcloud php occ config:system:set trusted_proxies 2 --value=ip.address.of.proxy
 ```
 
-Advice: In case you run AIO behind Cloudflare Tunnel, you must add all [Cloudflare IP-Ranges](https://www.cloudflare.com/ips/) to the WOPI allowlist.
-
 ### Collabora WOPI allow list
 If your reverse proxy connects to Nextcloud with an IP address that is different from the one for your domain<sup>*</sup> and you are using the Collabora server then you must also add the IP to the WOPI request allow list via `Administration Settings > Administration > Office > Allow list for WOPI requests`.
 
-<small>*: For example, the reverse proxy has a public globally routable IP and connects to your AIO instance via Tailscale with an IP in the `100.64.0.0/10` range, or you are using a Cloudflare tunnel ([cloudflare notes](https://github.com/nextcloud/all-in-one?tab=readme-ov-file#notes-on-cloudflare-proxytunnel)).</small>
+<small>*: For example, the reverse proxy has a public globally routable IP and connects to your AIO instance via Tailscale with an IP in the `100.64.0.0/10` range, or you are using a Cloudflare tunnel ([cloudflare notes](https://github.com/nextcloud/all-in-one?tab=readme-ov-file#notes-on-cloudflare-proxytunnel): You must add all [Cloudflare IP-Ranges](https://www.cloudflare.com/ips/) to the WOPI allowlist.)</small>
 
 ### External reverse proxies connecting via VPN (e.g. Tailscale)
 


Containers/mastercontainer/start.sh
```diff
@@ -313,9 +313,6 @@ fi
 if [ -n "$NEXTCLOUD_TRUSTED_DOMAINS" ]; then
     print_red "The environmental variable NEXTCLOUD_TRUSTED_DOMAINS has been set which is not supported by AIO. Please remove it!"
 fi
-if [ -n "$TRUSTED_PROXIES" ]; then
-    print_red "The environmental variable TRUSTED_PROXIES has been set which is not supported by AIO. Please remove it!"
-fi
```
Zoey2936: I think this should be kept, since the env you added is called ADDITIONAL_TRUSTED_PROXY

asavageiv: Whoops. Good call. Done

szaimen:
Hi, thanks for the PR!

However adjusting the trusted proxies is already possible. See https://github.com/nextcloud/all-in-one/discussions/2045

Maybe we should add a link to the RP docs to make this more visible?

asavageiv:
Thanks for the reviews!

I also see that this conflicts with `ADDITIONAL_TRUSTED_PROXY` being set by `CADDY_IP_ADDRESS`, so this won't work as-is.

My initial PR description is incomplete and the name `ADDITIONAL_TRUSTED_PROXY` is misleading. Using `ADDITIONAL_TRUSTED_PROXY` because it also adds the IP to the `wopi_allowlist` in `Containers/nextcloud/entrypoint.sh`. It may be that adding it to the WOPI allow list is not necessary if #6676 gets merged. I'm not sure.

I wanted an environment variable to keep a clean and clear interface for supporting reverse proxies. That way if internal things change, like the need for it to be on the allow list, users won't have to make a change and us developers won't have to try and communicate that change. It parallels the way `APACHE_ADDITIONAL_NETWORK` is used for RPs on the same host.

But, I could be even more explicit about the intent with the environment variable name with something like `REVERSE_PROXY_IP`. That would give us a clean interface and maybe allow us to clean up some of the allow list and trusted proxy code internally in the future.

I'm fine with either of these options:
1. (preferred) Use a `REVERSE_PROXY_IP` environment variable in `compose.yaml`
2. Update the RP docs to say "if you're running a RP on a remote host, adjust `trusted_proxies` and the `WOPI allowlist` like this: ..." 

Let me know what you prefer @szaimen. There may be other reasons to prefer one or the other that I'm missing.

asavageiv:
I forgot to mention, the external proxy IP also needs to be in the apache container Caddyfile. If it is only set in Nextcloud then the internal Caddy will have its IP detected as `127.0.0.1`, so I'm not sure how to pass it there if it's not an ENV variable.

szaimen:
Hey, I am back from vacation and currently going through my notifications.

> I forgot to mention, the external proxy IP also needs to be in the apache container Caddyfile. If it is only set in Nextcloud then the internal Caddy will have its IP detected as `127.0.0.1`, so I'm not sure how to pass it there if it's not an ENV variable.

I am honestly not sure about this. This is only needed in your custom setup because you use a public ip-address for the connection between RP and AIO instance IIRC...

szaimen:
> Thanks for the reviews!
> 
> I also see that this conflicts with `ADDITIONAL_TRUSTED_PROXY` being set by `CADDY_IP_ADDRESS`, so this won't work as-is.
> 
> My initial PR description is incomplete and the name `ADDITIONAL_TRUSTED_PROXY` is misleading. Using `ADDITIONAL_TRUSTED_PROXY` because it also adds the IP to the `wopi_allowlist` in `Containers/nextcloud/entrypoint.sh`. It may be that adding it to the WOPI allow list is not necessary if #6676 gets merged. I'm not sure.
> 
> I wanted an environment variable to keep a clean and clear interface for supporting reverse proxies. That way if internal things change, like the need for it to be on the allow list, users won't have to make a change and us developers won't have to try and communicate that change. It parallels the way `APACHE_ADDITIONAL_NETWORK` is used for RPs on the same host.
> 
> But, I could be even more explicit about the intent with the environment variable name with something like `REVERSE_PROXY_IP`. That would give us a clean interface and maybe allow us to clean up some of the allow list and trusted proxy code internally in the future.
> 
> I'm fine with either of these options:
> 
> 1. (preferred) Use a `REVERSE_PROXY_IP` environment variable in `compose.yaml`
> 2. Update the RP docs to say "if you're running a RP on a remote host, adjust `trusted_proxies` and the `WOPI allowlist` like this: ..."
> 
> Let me know what you prefer @szaimen. There may be other reasons to prefer one or the other that I'm missing.

I need to think about this...

asavageiv:
I drew my set up below for clarification. I'm surprised this is not a common configuration.

I updated the PR to allow `ADDITIONAL_TRUSTED_PROXY` xor the Caddy community container and removed `CADDY_IP_ADDRESS` placeholder in favor of returning the `CADDY_IP_ADDRESS` value as the value of `ADDITIONAL_TRUSTED_PROXY` if the community container is enabled. This should keep things working for the community container and solve my issue.

asavageiv nextcloud

szaimen:
> I drew my set up below for clarification. I'm surprised this is not a common configuration.

I don't think it is common to use public ip ranges in the VPN subnet. Using a private ip range should automatically solve the problem, no?

szaimen:
> > I'm fine with either of these options:
> > 
> > 1. (preferred) Use a `REVERSE_PROXY_IP` environment variable in `compose.yaml`
> > 2. Update the RP docs to say "if you're running a RP on a remote host, adjust `trusted_proxies` and the `WOPI allowlist` like this: ..."
> > 
> > Let me know what you prefer @szaimen. There may be other reasons to prefer one or the other that I'm missing.
> 
> I need to think about this...

Ok, thought about this now and I would prefer option 2 :)

asavageiv:
> > I drew my set up below for clarification. I'm surprised this is not a common configuration.
> 
> I don't think it is common to use public ip ranges in the VPN subnet. Using a private ip range should automatically solve the problem, no?

[Tailscale assigns addresses in the CGNAT range.](https://tailscale.com/kb/1015/100.x-addresses) A private address would solve it, but unfortunately, it's a Tailscale decision.

szaimen:
> > > I drew my set up below for clarification. I'm surprised this is not a common configuration.
> > 
> > 
> > I don't think it is common to use public ip ranges in the VPN subnet. Using a private ip range should automatically solve the problem, no?
> 
> [Tailscale assigns addresses in the CGNAT range.](https://tailscale.com/kb/1015/100.x-addresses) A private address would solve it, but unfortunately, it's a Tailscale decision.

Thanks for the link!

image

It is fine for me then to hard-code the address range into caddy to trust it by default. Then we don't need to add a config option for this 👍

Containers/apache/start.sh
```diff
@@ -46,7 +46,11 @@ echo "$CADDYFILE" > /tmp/Caddyfile
 
 # Change the trusted_proxies in case of reverse proxies
 if [ "$APACHE_PORT" != '443' ]; then
+  if [ -z "$ADDITIONAL_TRUSTED_PROXY" ]; then
     CADDYFILE="$(sed 's|# trusted_proxies placeholder|trusted_proxies static private_ranges|' /tmp/Caddyfile)"
+  else
+    CADDYFILE="$(sed "s|# trusted_proxies placeholder|trusted_proxies static private_ranges $ADDITIONAL_TRUSTED_PROXY|" /tmp/Caddyfile)"
+  fi
```
szaimen: So basically add the range here:
```suggestion
CADDYFILE="$(sed 's|# trusted_proxies placeholder|trusted_proxies static private_ranges 100.64.0.0/10|' /tmp/Caddyfile)"
```

asavageiv:
Thanks again for the thought.

Whitelisting 100.64.0.0/10 will work, but the addresses are shared with all the users using the same IP address for an ISP. For example, Starlink uses CGNAT and my gateway router is assigned an IP in the 100.64.0.0/10 range which is CGNATed to a global IPV4 IP we are all sharing (98.97.63.64). If we whitelist the whole range, then everyone sharing that global IP is also whitelisted. Is this safe enough?

> Ok, thought about this now and I would prefer option 2 :)

For option 2, we still need a way to configure the aio-apache Caddyfile.

szaimen:
> Whitelisting 100.64.0.0/10 will work, but the addresses are shared with all the users using the same IP address for an ISP. For example, Starlink uses CGNAT and my gateway router is assigned an IP in the 100.64.0.0/10 range which is CGNATed to a global IPV4 IP we are all sharing (98.97.63.64). If we whitelist the whole range, then everyone sharing that global IP is also whitelisted. Is this safe enough?

I dont have a starlink connection for testing but it should security wise not be possible to access other ip-adresses inside the ip range anyway, no? So I think this is safe enough



> > Ok, thought about this now and I would prefer option 2 :)
> 
> For option 2, we still need a way to configure the aio-apache Caddyfile.

Hm, no as the logic inside the container does this automatically if you use this: https://github.com/nextcloud/all-in-one/pull/6703#discussion_r2273690576

asavageiv:
> I dont have a starlink connection for testing but it should security wise not be possible to access other ip-adresses inside the ip range anyway, no?

It's up to the ISP. It is common for other IPs within the range to be reachable. Whether or not they have properly configured firewalls is another thing, but for AIO users if they have `APACHE_PORT: 11000` and `APACHE_IP_BINDING: 0.0.0.0` and their ISP doesn't isolate the addresses then their server would be accessible. Someone would still have to set up a reverse proxy with the correct domain name to connect to it, so maybe this is an acceptable risk.

szaimen:
> so maybe this is an acceptable risk.

Yes, I'd say so 👍

asavageiv:
Ok, implemented Option 2 along with adding the CGNAT range to the trusted proxies list

szaimen:
> Ok, implemented Option 2 along with adding the CGNAT range to the trusted proxies list

Thanks @asavageiv! ☺️

Btw, @asavageiv I sent you an invite to the repo for easier collaboration in the future :)

reverse-proxy.md
```diff
@@ -990,7 +990,19 @@ After starting AIO, you should be able to access the AIO Interface via `https://
 ⚠️ **Important:** do always use an ip-address if you access this port and not a domain as HSTS might block access to it later! (It is also expected that this port uses a self-signed certificate due to security concerns which you need to accept in your browser)<br>
 Enter your domain in the AIO interface that you've used in the reverse proxy config and you should be done. Please do not forget to open/forward port `3478/TCP` and `3478/UDP` in your firewall/router for the Talk container!
 
-## 5. Optional: get a valid certificate for the AIO interface
+## 5. Configure Nextcloud and Collabora trusted_proxies and WOPI allow list
```
szaimen: I guess we should add this step to the content section

reverse-proxy.md
```diff
@@ -990,7 +990,19 @@ After starting AIO, you should be able to access the AIO Interface via `https://
...
+
+```
+sudo docker exec --user www-data -it nextcloud-aio-nextcloud php occ config:system:set trusted_proxies 2 --value=ip.address.of.proxy
+```
+
+If using the Collabora server and the IP address of your domain is different from the one used to connect from your reverse proxy to AIO<sup>*</sup> then you must also add the IP to the WOPI request allow list via `Administration Settings > Administration > Office > Allow list for WOPI requests`.
+
+<small>*: For example, the reverse proxy has a public globally routable IP and connects to your AIO instance via Tailscale with an IP in the `100.64.0.0/10` range.</small>
+
+## 6. Optional: get a valid certificate for the AIO interface
```
szaimen, line 1015: we have the problem now that adjusting the heading will break old links from external to this section. This is probably not a big issue with this specific section but it is for `## 6. How to debug things?` which should be adjusted to `## 7. How to debug things?`. Do you have an idea how we can solve this?

Containers/apache/start.sh
```diff
@@ -46,7 +46,9 @@ echo "$CADDYFILE" > /tmp/Caddyfile
 
 # Change the trusted_proxies in case of reverse proxies
 if [ "$APACHE_PORT" != '443' ]; then
-    CADDYFILE="$(sed 's|# trusted_proxies placeholder|trusted_proxies static private_ranges|' /tmp/Caddyfile)"
+    # 100.64.0.0/10 is the CGNAT range used by Tailscale nodes.
+    # See https://github.com/nextcloud/all-in-one/pull/6703 for discussion.
```
szaimen:
```suggestion
    # Here the 100.64.0.0/10 range gets added which is the CGNAT range used by Tailscale nodes
    # See https://github.com/nextcloud/all-in-one/pull/6703 for reference
```

reverse-proxy.md
szaimen: The link then also needs to be adjusted in these places: https://github.com/search?q=repo%3Anextcloud%2Fall-in-one%20how-to-debug-things&type=code

szaimen: btw, the link also needs to be adjusted here: https://github.com/search?q=repo%3Anextcloud%2Fall-in-one+get-a-valid-certificate&type=code

asavageiv: This is a great point and something I hadn't considered. What we do depends a bit on the requirements for external linking. Do we want the whole web to be able to link to a section in this document? Just the "how to debug" section? Or just within the repository or Nextcloud organization?

We want to be able to reword and reorganize the document, but we also want somewhat stable links.

There are a few options:
1. Append a new section to the end or make it a subsection of an existing section. Easy right now but might make the doc harder to read over time. So I don't like this one. I want minimal barriers to making the documentation clearer.
1. Insert it, but add anchors to preserve backwards compatibility. E.g. 
    ```
    ## 7. How to debug things?
    <a id="6-how-to-debug-things"></a>
    ```
    Over time could lead to cruft build-up in the docs unless we prefer manually creating stable links over the auto-generated ones in the future. That is, in the future put `<a id="how-to-debug"` on things externally linked so there's no reference to the document structure in the link, just the content. It has the advantage of maintaining links on the WWW. We could also have both links and remove the old ones in the future.
1. Insert it and update all the links we know about. This breaks all links outside of the repo and leaves us always having to remember to check and search for any link changes in any documentation we update.

I like the idea of an explicit anchor for anything we want linked from outside the document. `<a id="how-to-debug">` makes it obvious "this thing is a stable link and shouldn't be changed" but lets us reorganize the document in the future. We probably want at least temporary backwards compatibility with error message links from older AIO versions, so I put in an additional `<a id="6-how-to-debug-things"></a><!-- for backwards compatibility -->` which we can remove later.

There might be other ways I'm not thinking of, but I think you get the idea.

szaimen: I like the solution that you commited! Thanks for this! :)

reverse-proxy.md
```diff
@@ -990,7 +991,28 @@ After starting AIO, you should be able to access the AIO Interface via `https://
 ⚠️ **Important:** do always use an ip-address if you access this port and not a domain as HSTS might block access to it later! (It is also expected that this port uses a self-signed certificate due to security concerns which you need to accept in your browser)<br>
 Enter your domain in the AIO interface that you've used in the reverse proxy config and you should be done. Please do not forget to open/forward port `3478/TCP` and `3478/UDP` in your firewall/router for the Talk container!
 
-## 5. Optional: get a valid certificate for the AIO interface
+## 5. Optional: Configure AIO for reverse proxies outside your LAN
+If your reverse proxy is on a host outside your LAN<sup>*</sup> you must make the following configuration changes
```
szaimen:
```suggestion
## 5. Optional: Configure AIO for reverse proxies that connect to nextcloud using an ip-address and not localhost nor 127.0.0.1
If your reverse proxy connects to nextcloud using an ip-address and not localhost or 127.0.0.1<sup>*</sup> you must make the following configuration changes
```

reverse-proxy.md
```diff
@@ -990,7 +991,28 @@ After starting AIO, you should be able to access the AIO Interface via `https://
...
+Add the IP it uses connect to AIO to the Nextcloud trusted_proxies like this:
+
+```
+sudo docker exec --user www-data -it nextcloud-aio-nextcloud php occ config:system:set trusted_proxies 2 --value=ip.address.of.proxy
+```
+
+### Collabora WOPI allow list
+If your reverse proxy connects to Nextcloud with an IP address that is different from the one for your domain<sup>*</sup> and you are using the Collabora server then you must also add the IP to the WOPI request allow list via `Administration Settings > Administration > Office > Allow list for WOPI requests`.
+
+<small>*: For example, the reverse proxy has a public globally routable IP and connects to your AIO instance via Tailscale with an IP in the `100.64.0.0/10` range.</small>
```
szaimen: Maybe we should mention cloudflare tunnel here?
See https://github.com/nextcloud/all-in-one?tab=readme-ov-file#notes-on-cloudflare-proxytunnel
image

asavageiv: Agreed, I think it's a good idea. Added a proposed small change.

reverse-proxy.md
```diff
@@ -990,7 +991,28 @@ After starting AIO, you should be able to access the AIO Interface via `https://
...
+
+<small>*: The IP address it uses to connect to AIO is not in a private IP range such as these: `127.0.0.1/8,192.168.0.0/16,172.16.0.0/12,10.0.0.0/8,fd00::/8,::1`</small>
+
+### Nextcloud trusted proxies
+Add the IP it uses connect to AIO to the Nextcloud trusted_proxies like this:
+
+```
+sudo docker exec --user www-data -it nextcloud-aio-nextcloud php occ config:system:set trusted_proxies 2 --value=ip.address.of.proxy
+```
+
```
szaimen, line 1005:
```suggestion

Advice: In case you run AIO behind Cloudflare Tunnel, you must add all [Cloudflare IP-Ranges](https://www.cloudflare.com/ips/) to the WOPI allowlist.

```

reverse-proxy.md
szaimen: Thanks! :)

szaimen:
This is now released with v11.7.0 Beta. Testing and feedback is welcome! See https://github.com/nextcloud/all-in-one#how-to-switch-the-channel

PR #6779, opened 2025-08-21
Add docs how to adjust MTU size of the docker network

Fly7113:
With the following proposed changes I'd like to introduce a new environment variable to change the MTU size of the network created by the mastercontainer when the Nextcloud containers are spawned.
We've been running this setup for a while, and it's been working nicely. Since I saw a few questions in the discussions and in the issues, I thought it'd be a good idea to create a PR to share our solution.

### DIFF
--- a/compose.yaml
+++ b/compose.yaml
@@ -35,6 +35,7 @@ services:
       # SKIP_DOMAIN_VALIDATION: false # This should only be set to true if things are correctly configured. See https://github.com/nextcloud/all-in-one?tab=readme-ov-file#how-to-skip-the-domain-validation
       # TALK_PORT: 3478 # This allows to adjust the port that the talk container is using which is exposed on the host. See https://github.com/nextcloud/all-in-one#how-to-adjust-the-talk-port
       # WATCHTOWER_DOCKER_SOCKET_PATH: /var/run/docker.sock # Needs to be specified if the docker socket on the host is not located in the default '/var/run/docker.sock'. Otherwise mastercontainer updates will fail. For macos it needs to be '/var/run/docker.sock'
+      # MTU_SIZE: 1500 # Allows to adjust the MTU size of the docker network. See https://github.com/nextcloud/all-in-one#how-to-adjust-the-mtu-size-of-the-docker-network
     # security_opt: ["label:disable"] # Is needed when using SELinux
 
 #   # Optional: Caddy reverse proxy. See https://github.com/nextcloud/all-in-one/discussions/575
--- a/compose.yaml
+++ b/compose.yaml
@@ -7,7 +7,7 @@ services:
     volumes:
       - nextcloud_aio_mastercontainer:/mnt/docker-aio-config # This line is not allowed to be changed as otherwise the built-in backup solution will not work
       - /var/run/docker.sock:/var/run/docker.sock:ro # May be changed on macOS, Windows or docker rootless. See the applicable documentation. If adjusting, don't forget to also set 'WATCHTOWER_DOCKER_SOCKET_PATH'!
-    network_mode: bridge # Add the container to the same network as docker run would do. Comment this line and uncomment the line below and the networks section at the end of the file if you want to define a custom MTU size for the docker network
+    network_mode: bridge # This adds the container to the same network as docker run would do. Comment this line and uncomment the line below and the networks section at the end of the file if you want to define a custom MTU size for the docker network
     # networks: ["nextcloud-aio"]
     ports:
       - 80:80 # Can be removed when running behind a web server or reverse proxy (like Apache, Nginx, Caddy, Cloudflare Tunnel and else). See https://github.com/nextcloud/all-in-one/blob/main/reverse-proxy.md
--- a/compose.yaml
+++ b/compose.yaml
@@ -7,7 +7,9 @@ services:
     volumes:
       - nextcloud_aio_mastercontainer:/mnt/docker-aio-config # This line is not allowed to be changed as otherwise the built-in backup solution will not work
       - /var/run/docker.sock:/var/run/docker.sock:ro # May be changed on macOS, Windows or docker rootless. See the applicable documentation. If adjusting, don't forget to also set 'WATCHTOWER_DOCKER_SOCKET_PATH'!
-    network_mode: bridge # add to the same network as docker run would do
+    network_mode: bridge # add to the same network as docker run would do. Comment this line and uncomment the networks section below and the one at the end of the file if you want to define a custom MTU size for the docker network
+    # networks:
+    #   nextcloud-aio
     ports:
       - 80:80 # Can be removed when running behind a web server or reverse proxy (like Apache, Nginx, Caddy, Cloudflare Tunnel and else). See https://github.com/nextcloud/all-in-one/blob/main/reverse-proxy.md
       - 8080:8080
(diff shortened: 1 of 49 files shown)


Zoey2936:
As an alternative solution, you could also just create the docker network manually with the MTU you want

Fly7113:
That could be a solution, but *it is* hard to implement an easy solution that works flawlessly with the AIO mechanism.
I can say this with confidence because, when developing a solution for our infrastructure, we got to a point where it was easier to start tinkering with a custom mastercontainer image rather than writing an external script.

szaimen:
Hi @Fly7113, 

I thought about this a few days and I agree with Zoey here that this is too much of an edge case to be baked directly into the mastercontainer. 

However we can and should update the docs and probably also the compose.yaml file to mention how to create the `nextcloud-aio` network beforehand with custom settings if you really need that.

Fly7113:
I see.
I am trying to write a succinct documentation on how to create the `nextcloud_aio` network beforehand to specify a different MTU (or other kind of customisations).

Personally I think that creating the network inside the docker-compose.yml is the cleaner way to archive this, but if no services/containers are attached to the network specified in the docker compose file, then the network doesn't get created.
As a workaround one could simply attach the mastercontainer to the network by specifying a `networks:` entry in the service declaration, but that is not possible when the `network_mode:` entry is specified. To solve this one would need to make the mastercontainer join only the `nextcloud_aio` network and not the default `bridge` one.

Could this be okay or is there a specific reason as to why the mastercontainer would need to join the default `bridge` network?

szaimen:
Yes, using the networks option is the correct approach. You could add a comment to the network_mode: bridge to comment that if you want to configure custom network options for the `nextcloud-aio` network and to uncomment the network options in that case

szaimen:
Feel free to commit a WIP for the compose.yaml file and I will comment the current state :)

Fly7113:
Let me know if the documentation and the comments look good to you.

compose.yaml
```diff
@@ -69,3 +71,10 @@ volumes: # If you want to store the data on a different drive, see https://githu
   # caddy_config:
   # caddy_data:
   # caddy_sites:
+
+# # Adjust the MTU size of the docker network. See https://github.com/nextcloud/all-in-one#how-to-adjust-the-mtu-size-of-the-docker-network
+# networks:
+#   nextcloud-aio:
+#     name: nextcloud-aio
+#     driver_opts:
+#       com.docker.network.driver.mtu: 1440
```
szaimen:
```suggestion
#       com.docker.network.driver.mtu: 1440

```

readme.md
```diff
@@ -353,6 +354,17 @@ See https://dev.to/ozorest/fedora-32-how-to-solve-docker-internal-network-issue-
...
+### How to adjust the MTU size of the docker network
+You can adjust the MTU size of the docker network by defining nextcloud's network in the `docker-compose.yml` file by adding the following lines and providing a value for the MTU size:
+```yaml
+networks:
+  nextcloud-aio:
+    name: nextcloud-aio
+    driver_opts:
+      com.docker.network.driver.mtu: <value>
+```
+When doing so, you have to comment the line `network_mode: "bridge"` and uncomment the `networks:` section in the mastercontainer's service definition.
```
szaimen: Usually in the readme we describe how to adjust the docker run command and not how to modify the compose file. 

Maybe we should add explanations for both in this case? WDYT?

compose.yaml
```diff
@@ -7,7 +7,9 @@ services:
     volumes:
       - nextcloud_aio_mastercontainer:/mnt/docker-aio-config # This line is not allowed to be changed as otherwise the built-in backup solution will not work
       - /var/run/docker.sock:/var/run/docker.sock:ro # May be changed on macOS, Windows or docker rootless. See the applicable documentation. If adjusting, don't forget to also set 'WATCHTOWER_DOCKER_SOCKET_PATH'!
-    network_mode: bridge # add to the same network as docker run would do
+    network_mode: bridge # add to the same network as docker run would do. Comment this line and uncomment the networks section below and the one at the end of the file if you want to define a custom MTU size for the docker network
```
szaimen:
```suggestion
    network_mode: bridge # Add the container to the same network as docker run would do. Comment this line and uncomment the networks section below and the one at the end of the file if you want to define a custom MTU size for the docker network
```

compose.yaml
```diff
@@ -7,7 +7,9 @@ services:
...
+    # networks:
+    #   nextcloud-aio
```
szaimen:
```suggestion
    # networks: ["nextcloud-aio"]
```

readme.md
Fly7113: Good point. Maybe we can just describe the commands to setup the network and connect the container to it, and then point to the comments in the `compose.yml`?

szaimen: sounds good

compose.yaml
```diff
@@ -7,7 +7,8 @@ services:
     volumes:
       - nextcloud_aio_mastercontainer:/mnt/docker-aio-config # This line is not allowed to be changed as otherwise the built-in backup solution will not work
       - /var/run/docker.sock:/var/run/docker.sock:ro # May be changed on macOS, Windows or docker rootless. See the applicable documentation. If adjusting, don't forget to also set 'WATCHTOWER_DOCKER_SOCKET_PATH'!
-    network_mode: bridge # add to the same network as docker run would do
+    network_mode: bridge # Add the container to the same network as docker run would do. Comment this line and uncomment the line below and the networks section at the end of the file if you want to define a custom MTU size for the docker network
```
szaimen:
```suggestion
    network_mode: bridge # This adds the container to the same network as docker run would do. Comment this line and uncomment the line below and the networks section at the end of the file if you want to define a custom MTU size for the docker network
```

readme.md
```diff
@@ -353,6 +354,15 @@ See https://dev.to/ozorest/fedora-32-how-to-solve-docker-internal-network-issue-
...
 If you get an error during the domain validation which states that your ip-address is an internal or reserved ip-address, you can fix this by first making sure that your domain indeed has the correct public ip-address that points to the server and then adding `--add-host yourdomain.com:<public-ip-address>` to the docker run command of the mastercontainer (but before the last line `ghcr.io/nextcloud-releases/all-in-one:latest`! If it was started already, you will need to stop the mastercontainer, remove it (no data will be lost) and recreate it using the docker run command that you initially used) which will allow the domain validation to work correctly. And so that you know: even if the `A` record of your domain should change over time, this is no problem since the mastercontainer will not make any attempt to access the chosen domain after the initial domain validation.
 
+### How to adjust the MTU size of the docker network
+You can adjust the MTU size of the docker network by creating it beforehand with the custom MTU:
+```
+docker network create --driver bridge --opt "com.docker.network.driver.mtu"="<value>" nextcloud-aio
```
```suggestion
`docker network create --driver bridge --opt com.docker.network.driver.mtu=1440 nextcloud-aio`
```

readme.md
```diff
@@ -353,6 +354,15 @@ See https://dev.to/ozorest/fedora-32-how-to-solve-docker-internal-network-issue-
...
+```
+and then by adding `--network=nextcloud-aio` to the docker run command of the mastercontainer (but before the last line `ghcr.io/nextcloud-releases/all-in-one:latest`).
```
szaimen: Actually AIO adds the container automatically to the network when you open the AIO interface. So this does not need to be done manually but maybe we can mention it. Also we should mention that the network needs to be created before visitng the aio interface the first time...

Fly7113: That's correct. I was still thinking about the compose problem of creating the network right away. I'll add this clarification.

readme.md
Fly7113: I see you added in the suggestion the inline code backticks around the command, but it is already in a code block (triple backticks)

szaimen: Ah you are right. I missed this. However adjusting thr value tag to use a dedicated value maybe makes it clearer?

readme.md
Fly7113: I added the clarification, let me know what you think of it!

szaimen (approved):
LGTM. Thank you very much! ☺️

PR #6802, opened 2025-08-28
compose.yaml: add some more explanation and links to docs

szaimen:


szaimen:
@Zoey2936 WDYT? Should we maybe also add a link to the top of the file that explains how to use this file?

PR #6818, opened 2025-09-03
update docs with new tailscale guide

szaimen:


szaimen:
@Perseus333 I agree that your guide is more simple and easier to use than the other guide. Since I usually prefer KISS, I'd like to promote your guide. Hope that is fine for you? :)

Perseus333:
@szaimen That was the original intent of the guide. Thanks for promoting my guide!

PR #6825, opened 2025-09-05
Use a fixed port for internal WOPI requests and callbacks

asavageiv:
This fixes the issue where Caddy fails to start when APACHE_PORT was 443.

See also [#6676](https://github.com/nextcloud/all-in-one/pull/6676#issuecomment-3258047014)

@szaimen This is one quick fix. Note that it should be OK if APACHE_PORT is also 23973. It would just end up with a block like this:
```
http://nextcloud-aio-apache:23973,
http://:23973
```
which is fine.
If APACHE_PORT is 443 we get this:
```
http://nextcloud-aio-apache:23973,
https://nextcloud.example.com:443
```
which should also work.

### DIFF
--- a/Containers/apache/Caddyfile
+++ b/Containers/apache/Caddyfile
@@ -15,7 +15,7 @@
 }
 
 https://{$ADDITIONAL_TRUSTED_DOMAIN}:443,
-http://{$APACHE_HOST}:{$APACHE_PORT}, # For Collabora callback
+http://{$APACHE_HOST}:23973, # For Collabora callback and WOPI requests, see containers.json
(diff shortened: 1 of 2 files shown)


szaimen:
Thank you! I will review on monday

szaimen (approved):
LGTM, thanks! ☺️

szaimen:
This is now released with v11.8.0 Beta. Testing and feedback is welcome! See https://github.com/nextcloud/all-in-one#how-to-switch-the-channel

asavageiv:
Just switched and everything is looking good! I'm back to running a pure version of AIO with no local modifications. IP addresses are detected correctly and the performance is great. Thank you for working with me to get these changes in!

PR #6835, opened 2025-09-08
facerecognition-cc: update json to use actual secret for api key

szaimen:
- This is possible since https://github.com/nextcloud/all-in-one/pull/6568
- Close part of https://github.com/nextcloud/all-in-one/discussions/3889

PR #6831, opened 2025-09-07
feat: Add shared_secrets and shared_environment for community containers

Grotax:
## Overview

This PR introduces two new concepts for community containers to securely share secrets and environment variables with all AIO containers, including Nextcloud.

Related: https://github.com/nextcloud/all-in-one/discussions/3889

I also added some explanation to the readme to make it a bit easier for new contributes.

### 🔐 `shared_secrets`
To allow sharing secret from community containers to all other containers it is now possible to add a shared_secret array to the container json file.
That shared_secret behaves the same as the secret except that it is shard to all other containers. 

### 🌐 `shared_environment` 
- Environment variables that are automatically injected into **ALL** AIO containers
- Uses same format as regular `environment` array
- Supports full placeholder replacement (`%SECRET_NAME%`, `%TIMEZONE%`, etc.)
- Perfect for sharing service URLs, API keys, and configuration flags

## Example Usage

```json
{
    "secrets": ["FACERECOGNITION_API_KEY"],
    "shared_secrets": ["FACERECOGNITION_API_KEY"], 
    "shared_environment": [
        "FACERECOGNITION_API_KEY=%FACERECOGNITION_API_KEY%",
        "FACERECOGNITION_URL=http://nextcloud-aio-facerecognition:5000"
    ],
    "nextcloud_exec_commands": [
        "php /var/www/html/occ config:system:set facerecognition.external_model_api_key --value %FACERECOGNITION_API_KEY%"
    ]
}

PR #6838, opened 2025-09-08
feat: Share declared secrets between all enabled containers

Grotax:
As discussed in #6831 this PR makes the secrets shared by default.

This allows for example that a community container can declare a new secret that will also be available in the nextcloud container context and can be used via string substitution in the nextcloud_exec area.

That also means that community containers sharing secrets from the core containers do not need to declare that but it doesn't hurt either and adds some clarity maybe.

Replaces #6831, #6835

I tested the fresh AIO creation but not an upgrade :)

### DIFF
--- a/php/src/Controller/DockerController.php
+++ b/php/src/Controller/DockerController.php
@@ -213,6 +213,9 @@ readonly class DockerController {
         $config['AIO_TOKEN'] = bin2hex(random_bytes(24));
         $this->configurationManager->WriteConfig($config);
 
+        // Generate all secrets from enabled containers upfront for placeholder substitution
+        $this->generateAllEnabledContainerSecrets();
+
         // Stop domaincheck since apache would not be able to start otherwise
         $this->StopDomaincheckContainer();
 
@@ -276,6 +279,9 @@ readonly class DockerController {
             return;
         }
 
+        // Generate all secrets from enabled containers for placeholder substitution
+        $this->generateAllEnabledContainerSecrets();
+
         $id = 'nextcloud-aio-domaincheck';
 
         $cacheKey = 'domaincheckWasStarted';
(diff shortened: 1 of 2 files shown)


community-containers/facerecognition/facerecognition.json
```diff
@@ -21,7 +24,7 @@
                 "php /var/www/html/occ app:install facerecognition",
                 "php /var/www/html/occ app:enable facerecognition", 
                 "php /var/www/html/occ config:system:set facerecognition.external_model_url --value nextcloud-aio-facerecognition:5000",
-                "php /var/www/html/occ config:system:set facerecognition.external_model_api_key --value some-super-secret-api-key",
+                "php /var/www/html/occ config:system:set facerecognition.external_model_api_key --value %FACERECOGNITION_API_KEY%",
```
szaimen: Lets do this in a separate pr. See https://github.com/nextcloud/all-in-one/pull/6835

php/src/Docker/DockerActionManager.php
```diff
@@ -221,9 +221,8 @@ public function CreateContainer(Container $container): void {
             $requestBody['HostConfig']['Binds'] = $volumes;
         }
 
-        foreach ($container->GetSecrets() as $secret) {
-            $this->configurationManager->GetAndGenerateSecret($secret);
-        }
+        // Generate all secrets from enabled containers to make them available for placeholder replacement
+        $this->generateAllEnabledContainerSecrets();
```
szaimen: This is now being called every time a container is created. I guess we should move this one level up for performance optimization. I guess it should be called here: https://github.com/nextcloud/all-in-one/blob/e4c2f44d81e13f52ce720bfe62119d5b88d99160/php/src/Controller/DockerController.php#L220

szaimen (changes requested):
Thanks! I have two comments

php/src/Docker/DockerActionManager.php
Grotax: Yea that worked out, also need to add a call to the method in StartDomaincheckContainer since that is called during the init phase and would miss secrets like the instance id 
It think that should be fine.

community-containers/facerecognition/facerecognition.json
Grotax: Ok, removed it.

php/src/Controller/DockerController.php
```diff
@@ -310,4 +316,26 @@ private function StopDomaincheckContainer() : void
...
+                // Generate all secrets from this container
+                foreach ($container->GetSecrets() as $secret) {
+                    // Only process each secret once for efficiency
+                    if (!in_array($secret, $processedSecrets, true)) {
+                        $this->configurationManager->GetAndGenerateSecret($secret);
+                        $processedSecrets[] = $secret;
+                    }
+                }
+            } catch (\Exception $e) {
+                // Container might not exist or have issues, continue with others
```
asavageiv, line 336: At a minimum we need to log what the exception is here. Silently swallowing all exceptions is rarely the right move and makes debugging super hard.

Looking at GetAndGenerateSecret, the thing I'd worry about is File IO failing, in which case I'm not sure we really want to continue.

Was there an Exception you saw during development that made you add this?

Grotax: True, swallowing would make debugging issues hard.

In my short tests there were no exceptions.

PR #6841, opened 2025-09-08
Make secrets global and init on first use

asavageiv:
This allows all containers to use any secret declared anywhere in their placeholders but they will not be generated and written to the configuration until they are used.

### DIFF
--- a/php/src/Container/Container.php
+++ b/php/src/Container/Container.php
@@ -19,8 +19,6 @@ readonly class Container {
         private ContainerEnvironmentVariables $containerEnvironmentVariables,
         /** @var string[] */
         private array                         $dependsOn,
-        /** @var string[] */
-        private array                         $secrets,
         private string                        $uiSecret,
         /** @var string[] */
         private array                         $devices,
@@ -80,10 +78,6 @@ readonly class Container {
 
     public function GetMaxShutdownTime() : int {
         return $this->maxShutdownTime;
-    }
-
-    public function GetSecrets() : array {
-        return $this->secrets;
     }
 
     public function GetUiSecret() : string {
--- a/php/src/ContainerDefinitionFetcher.php
+++ b/php/src/ContainerDefinitionFetcher.php
@@ -239,9 +239,12 @@ readonly class ContainerDefinitionFetcher {
                 $internalPort = $entry['internal_port'];
             }
 
-            $secrets = [];
             if (isset($entry['secrets'])) {
-                $secrets = $entry['secrets'];
+                // All secrets are registered with the configuration when they 
+                // are discovered so they can be later generated at time-of-use.
+                foreach ($entry['secrets'] as $secret) {
+                    $this->configurationManager->RegisterSecret($secret);
+                }
             }
 
             $uiSecret = '';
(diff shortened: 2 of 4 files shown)


Grotax:
What we need is a decision how it should work for the configuration.

This will of coure require that the Nextcloud container actually declares all the secrets it uses.

For example:
```
      "secrets": [
        "DATABASE_PASSWORD",
        "REDIS_PASSWORD",
        "NEXTCLOUD_PASSWORD",
        "TURN_SECRET",
        "SIGNALING_SECRET",
        "FULLTEXTSEARCH_PASSWORD",
        "IMAGINARY_SECRET",
        "WHITEBOARD_SECRET",
        "ONLYOFFICE_SECRET",
        "RECORDING_SECRET",
        "FACERECOGNITION_API_KEY"
      ],
      
```

This list will become longer of course with every community container added that needs to share a secret with Nextcloud.


From a functional view this would have also worked before if the Nextcloud container declared FACERECOGNITION_API_KEY it would have worked with the API, it is just more strict now.

asavageiv:
Agreed. If we truly want the secrets to be global because they can actually be used by any container then I think we should go ahead and make them global by putting them all at the top of containers.json in a separate `secrets` key. Having them all declared in one place would make it more obvious what is really going on. There aren't going to be so many of them that I think this would become a maintenance burden.

Declaring them with a container, but actually they become available everywhere is misleading and probably a security risk. All it takes is someone sneaking a placeholder into a community container ENV and they can read all the Nextcloud data. However, I know the container json changes are reviewed, so maybe it's acceptable.

I trust @szaimen to decide

szaimen:
> Declaring them with a container, but actually they become available everywhere is misleading and probably a security risk. All it takes is someone sneaking a placeholder into a community container ENV and they can read all the Nextcloud data. However, I know the container json changes are reviewed, so maybe it's acceptable.

I don't think this is a security risk since you point out correctly that all json changes are reviewed. So in my opinion this is acceptable. BTW, we already have this pattern for the nextcloud_exec_commands which are commands for a different container and not for the container where the commands are defined. So I think declaring secrets with a container but making them available gloabally is acceptable. Rather, this allows nextcloud_exec_commands to actually work as you would expect as dev since string replacement works then correctly...

> If we truly want the secrets to be global because they can actually be used by any container then I think we should go ahead and make them global by putting them all at the top of containers.json in a separate secrets key. Having them all declared in one place would make it more obvious what is really going on. 

I don't think that we should refactor this and have a global option for secrets that are not bound to a specific container anymore. See the reasons above.

asavageiv:
Ok, I pushed a different approach then. Instead of putting the generation in the Controller or ActionManager, the secrets are registered when they are discovered and then generated at time-of-use. This has the bonus of avoiding writing secrets to the config file if they are never used, only declared, and keeps the logic of the Controller and ActionManager a little more focused.

It complicates the DefinitionFetcher, but I think it more explicitly links how the containers.json and config file work together for secrets.

PTAL

szaimen (approved):
I like it, great solution! Thanks a lot! 😊

asavageiv:
Thanks @Grotax for pointing out the issue and inspiration for the solution.

Thanks @szaimen for the review and feedback

szaimen:
This is now released with v11.8.0 Beta. Testing and feedback is welcome! See https://github.com/nextcloud/all-in-one#how-to-switch-the-channel

PR #6843, opened 2025-09-09
Added the section Removing the reverse proxy

Verhoeckx:
Rewrote step 4

### DIFF
--- a/reverse-proxy.md
+++ b/reverse-proxy.md
@@ -1058,17 +1058,21 @@ If you, at some point, want to remove the reverse proxy, here are some general s
     sudo docker nextcloud-aio-mastercontainer && sudo docker rm nextcloud-aio-mastercontainer  
     ```
 3. Remove the software and configuration file that you used for the reverse proxy (see section 1).
-4. Restore the settings in the main configuration file of Nextcloud AIO.
-    1. Open the main configuration file.
-        ```
-        sudo docker run -it --rm --volume nextcloud_aio_mastercontainer:/mnt/docker-aio-config:rw alpine sh -c "apk add --no-cache nano && nano /mnt/docker-aio-config/data/configuration.json"
-        ```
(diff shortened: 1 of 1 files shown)


Verhoeckx:
Question: are the environmental variables this way set temporarily or permanently?

Verhoeckx:
To solve the DCO error I ran (as in the description) the following two commands in Github Codespaces:
```
git rebase HEAD~2 --signoff
git push --force-with-lease origin main
```

If this is not the solution, then I don't know what is.

reverse-proxy.md
```diff
@@ -1048,3 +1048,31 @@ If something does not work, follow the steps below:
...
+   --restart always \
+   --publish 80:80 \
+   --publish 8080:8080 \
+   --publish 8443:8443 \
+   --env APACHE_IP_BINDING=0.0.0.0 \
+   --env APACHE_PORT=443 \   
+   --volume nextcloud_aio_mastercontainer:/mnt/docker-aio-config \
+   --volume /var/run/docker.sock:/var/run/docker.sock:ro \
+   ghcr.io/nextcloud-releases/all-in-one:latest
+    ```
```
szaimen, line 1076: I am not sure if we should mention the command here again. Maybe it is better to just link to https://github.com/nextcloud/all-in-one?tab=readme-ov-file#how-to-use-this and its docker run command and mention that you need to add the two variables below to the command?

szaimen:
Unfortunately, it looks like DCO is still not solved but I ask internally if it is fine to merge this nonetheless since it is "only" a documentation update.

Verhoeckx:
I don't mind fixing the DCO error, if it's clear what I need to do for that.

szaimen:
> git rebase HEAD~2 --signoff
> git push --force-with-lease origin main

So this doeesnt work for you? What hapens if you run it again?

PR #6867, opened 2025-09-15
Make AIO interface not dependent on absolute location

Fly7113:
As a followup from #6782, here are the proposed changes to make AIO interface work relative to the browser location, without assuming the interface resides on the root path of the domain.
The changes can be summarised in:
- links in Twig templates: all the frontend links have been changed to be relative
- paths in location headers: (almost) all the location headers now redirect to the same page from where the request originated, which is (almost) always the "/container" page.
- getting the AIO path to inform Nextcloud of the correct location of the interface: a new frontend script collects and sends the path to the backend via the "start" button.

I tested every basic usage, including unauthenticated situations (which require a particular logic, as seen in the AuthMiddleware.php file) and modifying the settings.


Fly7113:
Hm, I tired to squash the commits but it looks like I did something wrong... I hope this doesn't create confusion.

szaimen:
First of all, thanks for creating this PR!

> Hm, I tired to squash the commits but it looks like I did something wrong... I hope this doesn't create confusion.

Squashing is not necessarily needed but can you please rebase the branch from main? This would be important. Thanks in advance!

Fly7113:
Rebased!
And squashed commits, so that the older and outdated commits (relative to my old work) are not visible. Less confusion!

Fly7113:
Today, with the release of v11.8.0, I had the chance to test also the update process, and everything went smoothly!

szaimen:
Thanks a lot for your effort! 

Just to be clear: I try to review it in the coming 1-2 weeks and try to include it for the next major version which is planned to be released in ca 1 month.

php/public/base_path.js
```diff
@@ -0,0 +1,3 @@
+document.addEventListener("DOMContentLoaded", function(event) {
+    document.getElementById("base_path") && (document.getElementById("base_path").value = window.location.pathname.slice(0, -11));
```
szaimen:
```suggestion
    basePath = document.getElementById("base_path")
    if (basePath) {
        basePath.value = window.location.pathname.slice(0, -11));
    }
```

php/src/Controller/LoginController.php
```diff
@@ -19,33 +19,33 @@ public function __construct(
...
-        return $response->withHeader('Location', '/')->withStatus(422);
+        return $response->withHeader('Location', '.')->withStatus(422);
     }
 
     public function GetTryLogin(Request $request, Response $response, array $args) : Response {
         $token = $request->getQueryParams()['token'] ?? '';
         if($this->authManager->CheckToken($token)) {
             $this->authManager->SetAuthState(true);
-            return $response->withHeader('Location', '/')->withStatus(302);
+            return $response->withHeader('Location', '../..')->withStatus(302);
```
szaimen, line 38: I am not sure about this syntax. Never seen this before. Shouldnt be the following more correct?
```suggestion
            return $response->withHeader('Location', '../../../')->withStatus(302);
```

php/src/Controller/LoginController.php
```diff
@@ -19,33 +19,33 @@ public function __construct(
...
         }
 
-        return $response->withHeader('Location', '/')->withStatus(302);
+        return $response->withHeader('Location', '../..')->withStatus(302);
```
szaimen, line 41: same

php/src/Controller/LoginController.php
```diff
@@ -19,33 +19,33 @@ public function __construct(
...
     }
 
     public function Logout(Request $request, Response $response, array $args) : Response
     {
         $this->authManager->SetAuthState(false);
         return $response
-            ->withHeader('Location', '/')
+            ->withHeader('Location', '.')
```
szaimen: Shouldnt this refer to the parent dir?
```suggestion
            ->withHeader('Location', '../')
```

php/src/Middleware/AuthMiddleware.php
```diff
@@ -27,7 +27,15 @@ public function __invoke(ServerRequestInterface $request, RequestHandlerInterfac
...
                 $status = 302;
-                $headers = ['Location' => '/'];
+                if(count(explode('/', $request->getUri()->getPath())) > 2) {
+                    $location = '..';
+                    for($i = 0; $i < count(explode('/', $request->getUri()->getPath())) - 3; $i++) {
+                        $location = $location . '/..';
+                    }
+                } else {
+                    $location = '.';
+                }
```
szaimen: I fear this syntax is very confusing. Can you please try to rewrite it to make it clearer? Also please add comments

php/templates/already-installed.twig
```diff
@@ -3,11 +3,11 @@
...
         <svg class="nextcloud-logo" xmlns="http://www.w3.org/2000/svg" viewBox="0 0 142 100" width="142" height="100">
-            <use href="/img/nextcloud-logo.svg#logo"></use>
-            <use href="/img/nextcloud-logo.svg#Nextcloud"></use>
+            <use href="img/nextcloud-logo.svg#logo"></use>
+            <use href="img/nextcloud-logo.svg#Nextcloud"></use>
             <text x="10" y="50" fill="var(--color-nextcloud-logo)" class="fallback-text">Nextcloud Logo</text>
         </svg>
         <h2>Nextcloud All-In-One is already installed</h2>
-        <a href="/" class="button">Open Nextcloud AIO</a>
+        <a href="." class="button">Open Nextcloud AIO</a>
```
szaimen, line 11: Shouldn't this refer to the parent dir?
```suggestion
        <a href="../" class="button">Open Nextcloud AIO</a>
```

php/templates/setup.twig
```diff
@@ -3,14 +3,14 @@
...
+            <use href="img/nextcloud-logo.svg#logo"></use>
+            <use href="img/nextcloud-logo.svg#Nextcloud"></use>
             <text x="10" y="50" fill="var(--color-nextcloud-logo)" class="fallback-text">Nextcloud Logo</text>
         </svg>
         <h1>All-in-One setup</h1>
         <p>The official Nextcloud installation method. Nextcloud All-in-One provides easy deployment and maintenance with most features included in this one Nextcloud instance.</p>
         <p>⚠️ <strong>Please note down the passphrase to access the AIO interface and don't lose it!</strong></p>
         <strong>Passphrase</strong><br/><span id="initial-password" class="monospace">{{ password }}</span><br>
-        <a href="/" class="button" target="_blank">Open Nextcloud AIO login ↗</a>
+        <a href="." class="button" target="_blank">Open Nextcloud AIO login ↗</a>
```
szaimen, line 14: Shouldnt this refer to the parent dir?
```suggestion
        <a href="../" class="button" target="_blank">Open Nextcloud AIO login ↗</a>
```

php/public/base_path.js
Fly7113: Will update accordingly and add comments!

php/src/Middleware/AuthMiddleware.php
Fly7113: Sure thing, I'll refactor it a bit and add some comments.

php/src/Controller/LoginController.php
Fly7113: Not really: when 'GetTryLogin' is invoked the path is set to '/api/auth/getlogin', so we are in 'auth' folder ('getlogin' is just a file in the folder) and we need to get up by one level to get to the 'api' folder and then by another one to get to the root (/base) folder.
Therefore we need to go up by two folder levels, hence we just need '../..'.
The trailing slash is optional, as in paths in bash for example.

php/src/Controller/LoginController.php
Fly7113: Same as above

php/src/Controller/LoginController.php
Fly7113: In this case it should be the same as the previously mentioned cases since we are in '/api/auth/logout': I'll update accordingly

php/templates/already-installed.twig
Fly7113: In this case, since the path doesn't end with a '/' then we are in the root (/base) folder and on a specific file. By using '.' we are referencing the same folder we currently are in, so the root (/base) one: once we reference this folder 'index.php' gets invoked and redirects us to the appropriate page ('/containers')

php/templates/setup.twig
Fly7113: Same as above

Fly7113:
Hello, thanks for finding the time to check it out.
I tested throughly in my "subfoldered" instance (daily driving it), and I also had the occasion to test it in a fresh environment without a base path.

I understand that the relative "Location" header can be confusing at times, especially since it's kind of a new feature: the HTTP spec has been updated in June 2014 to allow relative URIs in the "Location" header [RFC 7231]

Let me know what you think about the improved logic in the "Auth" middleware: I made the root scenario the first if case and added comments to explain the logic.

[RFC 7231]: http://tools.ietf.org/html/rfc7231#section-7.1.2

szaimen:
@Fly7113 the PR looks good to me generally now. Can you please rebase one last time? I would then merge it :)

daks:
Great, thanks both of you :)

szaimen:

PR #6908, opened 2025-09-27
Add LLDAP automatic configuration

docjyJ:
After automatic configuration for ClamAV, I would like to do the same for lldap.

I don't know if shared secrets work?

I think that if lldap is enabled, then the secret will be set, otherwise the secret will be empty. Do you agree?

Or else an enable_ldap variable would be required.

I would like your opinion on this feature.


szaimen:
Hey :)

> After automatic configuration for ClamAV, I would like to do the same for lldap.

Sounds good!

> I don't know if shared secrets work?

Yes, should work but please add `LLDAP_LDAP_USER_PASS` to secrets of the stalwart.json

> I think that if lldap is enabled, then the secret will be set, otherwise the secret will be empty. Do you agree?

No, the secret will be set all the time. So I guess we need a different way to detect if lldap is enabled. For example you could try to use dig to check if nextcloud-aio-lldap is there during the startup of stalwart...

> Or else an enable_ldap variable would be required.

Yeah, I guess this would be the cleanest way...

> I would like your opinion on this feature.

done :)

szaimen:
@docjyJ any update here? :)

docjyJ:
No, I'm stuck. Because we have to save the LDAP configuration ID, and that's not possible at the moment.

I add the discussion label.

PR #6909, opened 2025-09-27
use nextcloud_exec_commands to configure LLDAP

docjyJ:
You said:

> FYI: you can now use string replacement in nextcloud_exec_commands since https://github.com/nextcloud/all-in-one/pull/6835.


docjyJ:
There remains the issue of the configuration ID.

szaimen:
Hey :)

The problem that I see with this automatic configuration is that it will run the command every time the nextcloud container restarts. So it will add new configurations all the time. If we want to automate this, we need to add a check that the config is not existing yet.

I am also not sure if we run into a limitation regarding max characters at some point because the commands are added to the Nextcloud container via 1 env...

docjyJ:
> The problem that I see with this automatic configuration is that it will run the command every time the nextcloud container restarts. So it will add new configurations all the time. If we want to automate this, we need to add a check that the config is not existing yet.

One thing that would be nice is to be able to define an ID, such as `occ ldap:create-empty-config --name aio-lldap`. I will look at the LDAP code and maybe try to open a PR.

> I am also not sure if we run into a limitation regarding max characters at some point because the commands are added to the Nextcloud container via 1 env...

It seems that the limit is over 100,000 characters. See https://unix.stackexchange.com/questions/357843/setting-a-long-environment-variable-breaks-a-lot-of-commands

szaimen:
> > The problem that I see with this automatic configuration is that it will run the command every time the nextcloud container restarts. So it will add new configurations all the time. If we want to automate this, we need to add a check that the config is not existing yet.
> 
> One thing that would be nice is to be able to define an ID, such as `occ ldap:create-empty-config --name aio-lldap`. I will look at the LDAP code and maybe try to open a PR.

Sounds good :)

> > I am also not sure if we run into a limitation regarding max characters at some point because the commands are added to the Nextcloud container via 1 env...
> 
> It seems that the limit is over 100,000 characters. See https://unix.stackexchange.com/questions/357843/setting-a-long-environment-variable-breaks-a-lot-of-commands

Ok cool. Lets hope that docker envs have the same limit :)

docjyJ:
PR open on Nextcloud, feedback welcome.

https://github.com/nextcloud/server/pull/55424

PR #6914, opened 2025-09-28
Add support of overleaf as community container

docjyJ:


community-containers/overleaf/overleaf.json
```diff
@@ -0,0 +1,96 @@
...
+          "writeable": true
+        }
+      ],
+      "backup_volumes": [
+        "nextcloud_aio_overleaf_mongo"
+      ]
+    },
+    {
+      "container_name": "nextcloud-aio-overleaf",
+      "display_name": "Overleaf",
```
szaimen: I guess this main container should be the first because IIRC we use the first displayname for the checkbox for the community container in the AIO interface...

szaimen (commented):
Hey, thanks a lot already for starting to work on this! ☺️

I have one comment after having a first look:

community-containers/overleaf/overleaf.json
docjyJ: This is AI generated, I need to read the documentation. Thank you, I'll take note of that.

This week-end, the noise was too much. thats why it in draft.

docjyJ:
Another thing, it's a shame to run a second Redis and Database server. Is it possible to use the existing servers?

szaimen:
> Another thing, it's a shame to run a second Redis and Database server. Is it possible to use the existing servers?

For stability and compatibility reasons I would recommend to use dedicated redis and database even though it might come with a small overhead. 

However it might be possible to reuse the ghcr.io/nextcloud-releases/aio-redis image also for overleaf.

Also not sure if overleaf always needs an external database? Maybe it can run with sqlite instead?

docjyJ:
Okay! I'll take a look when I have time.

szaimen:
Hi @docjyJ any update here? :)

docjyJ:
Not enough time to work on it. I need a test environment to speed up development. I should convert an old PC into a test server.

szaimen:
Hey, what is the current state of this? Can I help to push this foward? :)

docjyJ:
It's quite complex, I haven't done anything more. Yes, you can commit to this branch!

community-containers/overleaf/overleaf.json
szaimen: done

community-containers/overleaf/overleaf.json
```diff
@@ -0,0 +1,101 @@
...
+      "command": "--replSet overleaf",
+      "volumes": [
+        {
+          "source": "nextcloud_aio_overleaf_mongo",
+          "destination": "/data/db",
+          "writeable": true
+        },
+                {
+          "source": "./bin/shared/mongodb-init-replica-set.js",
+          "destination": "/docker-entrypoint-initdb.d/mongodb-init-replica-set.js",
```
szaimen, line 92: I fear there are some customizations made to the mongodb image which we cannot do from inside AIO. 

So the only way I see is to built a custom mongodb image (maybe build it in a github repo) and build the customizations directly into the image...

@docjyJ would you be up for this?

szaimen: Hi @docjyJ WDYT?

PR #6916, opened 2025-09-28
add comment about possibility of migration AIO->VM

alexanderdd:


readme.md
```diff
@@ -52,7 +52,7 @@ Included are:
 - Possibility included to [pass the needed device for hardware transcoding](https://github.com/nextcloud/all-in-one#how-to-enable-hardware-acceleration-for-nextcloud) to the Nextcloud container
 - Possibility included to [store all docker related files on a separate drive](https://github.com/nextcloud/all-in-one#how-to-store-the-filesinstallation-on-a-separate-drive)
 - [LDAP can be used as user backend for Nextcloud](https://github.com/nextcloud/all-in-one/tree/main#ldap)
-- Migration from any former Nextcloud installation to AIO is possible. See [this documentation](https://github.com/nextcloud/all-in-one/blob/main/migration.md)
+- Migration from any former Nextcloud installation to AIO is possible. See [this documentation](https://github.com/nextcloud/all-in-one/blob/main/migration.md). Migration in the other direction (e.g. from AIO to a VM-based installation) is also possible.
```
szaimen: Hey, I am not sure if this is the best place. Maybe add another `-` on a new line and mention that there? Also possibly we should be a bit more verbose how to do it? :)

alexanderdd: Yes, explaining how to do the reverse direction would also be good. But my goal for this PR is to at least mention that migration is possible both ways.

IMO for now it is ok to have it all in one bullet point.

szaimen (approved):
LGTM, thanks!

Issue #6893, opened 2025-09-25
Nextcloud AIO Mastercontainer stuck after update because of caddy lockfile


axx-fdas:
### Steps to reproduce
It happened randomly after roughly one year of runtime, it could maybe be reproduced by killing the AIO mastercontainer without a graceful shutdown.

### Expected behavior 
The AIO mastercontainer should start normally after restarting itself for updates.
On a fresh start it shoud probably delete all lock files.

### Actual behavior 
The AIO mastercontainer was stuck on start, even re-starting it via docker did not change anything. 
The container produced no logs on starting.
I randomly discovered while looking through the persistentent volume of the container that  `/mnt/docker-aio-config/caddy/locks` contained a file (sadly I don't have the name anymore).
After deleting this lockfile the container could be started normally, and the system is running ever since.

#### Host OS

```
root@nextcloud:~# cat /etc/debian_version 
12.12
```

#### Output of `sudo docker info`

```
root@nextcloud:~# docker info
Client: Docker Engine - Community
 Version:    28.4.0
 Context:    default
 Debug Mode: false
 Plugins:
  buildx: Docker Buildx (Docker Inc.)
    Version:  v0.28.0
    Path:     /usr/libexec/docker/cli-plugins/docker-buildx
  compose: Docker Compose (Docker Inc.)
    Version:  v2.39.4
    Path:     /usr/libexec/docker/cli-plugins/docker-compose

Server:
 Containers: 16
  Running: 14
  Paused: 0
  Stopped: 2
 Images: 28
 Server Version: 28.4.0
 Storage Driver: overlay2
  Backing Filesystem: extfs
  Supports d_type: true
  Using metacopy: false
  Native Overlay Diff: true
  userxattr: false
 Logging Driver: json-file
 Cgroup Driver: systemd
 Cgroup Version: 2
 Plugins:
  Volume: local
  Network: bridge host ipvlan macvlan null overlay
  Log: awslogs fluentd gcplogs gelf journald json-file local splunk syslog
 CDI spec directories:
  /etc/cdi
  /var/run/cdi
 Swarm: inactive
 Runtimes: io.containerd.runc.v2 runc
 Default Runtime: runc
 Init Binary: docker-init
 containerd version: 05044ec0a9a75232cad458027ca83437aae3f4da
 runc version: v1.2.5-0-g59923ef
 init version: de40ad0
 Security Options:
  apparmor
  seccomp
   Profile: builtin
  cgroupns
[...]

PR #6919, opened 2025-09-29
delete caddy locks if existing on startup

szaimen:
- Fix https://github.com/nextcloud/all-in-one/issues/6893


Containers/apache/start.sh
```diff
@@ -66,6 +66,11 @@ caddy fmt --overwrite /tmp/Caddyfile
 # Add caddy path
 mkdir -p /mnt/data/caddy/
 
+# Fix caddy startup
+if [ -d "/mnt/docker-aio-config/caddy/locks" ]; then
+    rm -rf /mnt/docker-aio-config/caddy/locks/*
```
szaimen: Path needs to be adjusted

szaimen: done

PR #6949, opened 2025-10-08
docs: Add README for `mastercontainer`

joshtrichards:
Added README.md for the Nextcloud All-in-One mastercontainer, detailing its purpose, responsibilities, usage, and related resources.

### DIFF
--- a/Containers/mastercontainer/README.md
+++ b/Containers/mastercontainer/README.md
@@ -14,8 +14,7 @@ of all other containers in the Nextcloud All-in-One stack. It hosts:
 - A dedicated PHP SAPI/backend (php-fpm) for AIO itself (not Nextcloud Server)
 - An Apache service for accessing the AIO interface via a self-signed HTTPS VirtualHost on 8080/tcp
 - A Caddy reverse proxy service enabling HTTPS access to the AIO frontend on port 8443/tcp.
-  - This uses a self-signed certificate by default.
-  - The self-signed certificate can be automatically replaced by a Let's Encrypt issued certificate if port 80
+  - Caddy will automatically issue a Let's Encrypt issued certificate if port 80 and 8443
     is open/forwarded and a domain pointer is in place; then, simply open the Nextcloud AIO interface using the
     domain (`https://your-domain-that-points-to-this-server.tld:8443`). The Let's Encrypt certificate request will
     use an [ACME HTTP-01](https://letsencrypt.org/docs/challenge-types/#http-01-challenge) challenge.


Containers/mastercontainer/README.md
```diff
@@ -0,0 +1,67 @@
+# Nextcloud All-in-One `mastercontainer` Container
```
szaimen: This seems to double the container part...
```suggestion
# Nextcloud All-in-One `mastercontainer`
```

Containers/mastercontainer/README.md
```diff
@@ -0,0 +1,67 @@
...
+
+This folder contains the OCI/Docker container definition, along with associated resources and
+configuration files, for building the `mastercontainer` as part of the Nextcloud All-in-One
+project. This container hosts [the Nextcloud All-in-One management portal and API](
```
szaimen: I usually call it the `AIO interface`
```suggestion
project. This container hosts [the Nextcloud AIO interface and API](
```

Containers/mastercontainer/README.md
```diff
@@ -0,0 +1,67 @@
...
+for it (which is completely independent of the Nextcloud Server).
+
+## Overview
+
+The mastercontainer acts as the central orchestration service for the deployment and management
+of all other containers in the Nextcloud All-in-One stack. It hosts:
+
+- A dedicated PHP SAPI/backend (php-fpm) for AIO itself (not Nextcloud Server)
+- An Apache service for accessing the AIO frontend via a cleartext HTTP VirtualHost on
+  8000/tcp as well as a self-signed HTTPS VirtualHost on 8080/tcp
```
szaimen: I would not like to expose the fact about the unencrypted port 8000 publicly as it is unsecure. 
```suggestion
- An Apache service for accessing the AIO interface via a self-signed HTTPS VirtualHost on 8080/tcp
```

Containers/mastercontainer/README.md
```diff
@@ -0,0 +1,67 @@
...
+- A Caddy reverse proxy service for accessing the AIO frontend (optionally) via HTTPS with a
+  valid certificate on 8443/tcp.
```
szaimen: Maybe we should point out that port 80 is needed to retrieve the Lets encrypt certificate via http challenge for the interface running on port 8443?

Containers/mastercontainer/README.md
```diff
@@ -0,0 +1,67 @@
...
+It triggers the initial installation and ensures the smooth operation of the Nextcloud
+All-in-One stack.
+
+## Contents
+
+- **Dockerfile**: Instructions for building the mastercontainer image.
+- **Entrypoint script**: The `start.sh` script is used for container initialization and runtime
+  configuration before starting supervisord.
+- **Nextcloud All-in-One Controller App**: The core AIO orchestrator that handles configuration
+  and settings for the containers.
```
szaimen: Maybe we should add a link here also to https://github.com/nextcloud/all-in-one/tree/main/php?

Containers/mastercontainer/README.md
```diff
@@ -0,0 +1,67 @@
...
+Contributions are welcome! Please follow the Nextcloud project's guidelines and submit pull
+requests or issues via the main repository.
+
+## License
+
+This folder and its contents are licensed under the
+[GNU AGPLv3](https://www.gnu.org/licenses/agpl-3.0.html), in line with the rest of Nextcloud
+All-in-One.
+
+[^app]: The Nextcloud All-in-One management portal allows users to install, configure, and
```
szaimen: I usually call it the ´AIO interface´
```suggestion
[^app]: The Nextcloud AIO interface allows users to install, configure, and
```

Containers/mastercontainer/README.md
joshtrichards: Yeah, I went back and forth on that one. I agree it's weird *with* it even if it's more consistent. It's less weird I guess w/o it even if it's inconsistent.

Containers/mastercontainer/README.md
joshtrichards: Hmm. I see your point. I was mostly thinking of these docs as being dev/contributor (+power users, customizers, etc.) documentation so I lean towards it being included.

8000/tcp is only supposed to be accessed via the 8080/tcp (HTTPS) via ProxyPass here I think right?

https://github.com/nextcloud/all-in-one/blob/6b3eac0be9b21098673f3233795230a0289a9890/Containers/mastercontainer/mastercontainer.conf#L44

Maybe we should consider adding something like this to the mastercontainer.conf Apache config:

```apache
<VirtualHost *:8000>
[...]
  <Directory />
    Require ip 127.0.0.1
  </Directory>
[...]
```

Then the unecrypted port will never be accessible directly anyhow. ;-)

Containers/mastercontainer/README.md
joshtrichards: Done

Containers/mastercontainer/README.md
joshtrichards: Done

Containers/mastercontainer/README.md
joshtrichards: Done

Containers/mastercontainer/README.md
joshtrichards: Done

Containers/mastercontainer/README.md
szaimen: > 8000/tcp is only supposed to be accessed via the 8080/tcp (HTTPS) via ProxyPass here I think right?
> 
> https://github.com/nextcloud/all-in-one/blob/6b3eac0be9b21098673f3233795230a0289a9890/Containers/mastercontainer/mastercontainer.conf#L44

Yes, correct!

> Maybe we should consider adding something like this to the mastercontainer.conf Apache config:
> 
> ```apacheconf
> <VirtualHost *:8000>
> [...]
>   <Directory />
>     Require ip 127.0.0.1
>   </Directory>
> [...]
> ```
> 
> Then the unecrypted port will never be accessible directly anyhow. ;-)

Good idea! Would you mind adding this to the mastercontainer.conf? :)

Zoey2936: but could port 8000 then not also be removed? So that the ssl config of port 8080 is moved to the port 8000, the original port 8080 then removed and port 8000 then changed to be the new port 8080?

szaimen: I dont think so because caddy also proxies to the unencrypted port...

Zoey2936: But it could be changed to proxy to the encrypted 8080 (also maybe #6998 is possible?)
Another idea would be to just use an ip binding to 127.0.0.1 for port 8000: https://httpd.apache.org/docs/2.4/bind.html, I think this would be the cleaner way

szaimen: > Another idea would be to just use an ip binding to 127.0.0.1 for port 8000: https://httpd.apache.org/docs/2.4/bind.html

Lets do it like that :)

Zoey2936: see #7000

szaimen: https://github.com/nextcloud/all-in-one/pull/7000 was merged so @joshtrichards I guess the note regarding port 8000 can be removed now :)

Containers/mastercontainer/README.md
```diff
@@ -0,0 +1,70 @@
...
+of all other containers in the Nextcloud All-in-One stack. It hosts:
+
+- A dedicated PHP SAPI/backend (php-fpm) for AIO itself (not Nextcloud Server)
+- An Apache service for accessing the AIO interface via a self-signed HTTPS VirtualHost on 8080/tcp
+- A Caddy reverse proxy service enabling HTTPS access to the AIO frontend on port 8443/tcp.
+  - This uses a self-signed certificate by default.
+  - The self-signed certificate can be automatically replaced by a Let's Encrypt issued certificate if port 80
+    is open/forwarded and a domain pointer is in place; then, simply open the Nextcloud AIO interface using the
+    domain (`https://your-domain-that-points-to-this-server.tld:8443`). The Let's Encrypt certificate request will
+    use an [ACME HTTP-01](https://letsencrypt.org/docs/challenge-types/#http-01-challenge) challenge.
```
szaimen: I just tested this again and it will simply not open the page if the setup is not correct. It will either show a `ERR_SSL_PROTOCOL_ERROR` or `ERR_TIMED_OUT`.

```suggestion
  - Caddy will automatically issue a Let's Encrypt issued certificate if port 80 and 8443
    is open/forwarded and a domain pointer is in place; then, simply open the Nextcloud AIO interface using the
    domain (`https://your-domain-that-points-to-this-server.tld:8443`). The Let's Encrypt certificate request will
    use an [ACME HTTP-01](https://letsencrypt.org/docs/challenge-types/#http-01-challenge) challenge.
```

Issue #6964, opened 2025-10-12
Collabora container should not require extra capabilities/rootful Docker anymore


Fs00:
### Is your feature request related to a problem? Please describe. 
I was looking into how to setup Nextcloud AIO to run with Docker rootless and noticed the heads-up regarding Collabora not working in such mode in the [Docker rootless documentation](https://github.com/nextcloud/all-in-one/blob/main/docker-rootless.md).
However, the upstream Collabora issue linked in the document (https://github.com/CollaboraOnline/online/issues/2800) has seen a couple of responses from Collabora devs recently. They have stated that the container does not need root permissions or extra capabilities anymore, but only a custom [seccomp security profile](https://docs.docker.com/engine/security/seccomp/) which enables a few extra syscalls required for their "jail" feature.
Indeed, [their official docs for the Docker image](https://sdk.collaboraonline.com/docs/installation/CODE_Docker_image.html#container-security-and-capability-management) recommend to either grant only the required capabilities or use the custom seccomp profile.

### Describe the solution you'd like
Ideally, Nextcloud AIO should employ the Collabora-provided custom seccomp profile to reduce the amount of privileges given to the Collabora container.
If the latter is now able to work without root privileges, the Docker rootless docs should be updated (haven't tested yet so I cannot confirm).

szaimen:
Thanks for the report! I'll try to have a look in the coming days.

Fs00:
Thank you!
By the way, I've also noticed that [the Collabora container is given the `--o:mount_jail_tree=false`](https://github.com/nextcloud/all-in-one/blob/26bb93df914e70f83f8faa8b19f0bdf51293a980/php/containers.json#L383C69-L383C83) parameter which in theory disables the jail feature that requires the extra privileges. 🤔 
If the custom seccomp profile works as intended in rootless mode, I think that jails could (and should) be re-enabled since they are a [nice feature that improves isolation](https://sdk.collaboraonline.com/docs/architecture.html).

szaimen:
Hey @Fs00 If I understand the docs correctly, the seccomp profile needs to be downloaded onto the server where the docker deaemon runs and then specified as option for the collabora container which then uses the seccomp profile that must be exisitng on the host for the collabora container, correct?

Fs00:
[...]

PR #6995, opened 2025-10-16
collabora: load seccomp profile on the fly

szaimen:
- Close https://github.com/nextcloud/all-in-one/issues/6964
- Close https://github.com/nextcloud/all-in-one/issues/7056

### DIFF
--- a/php/src/Docker/DockerActionManager.php
+++ b/php/src/Docker/DockerActionManager.php
@@ -48,7 +48,7 @@ readonly class DockerActionManager {
             throw $e;
         }
 
-        $responseBody = json_decode((string)$response->getBody(), true);
+        $responseBody = json_decode((string)$response->getBody(), true, 512, JSON_THROW_ON_ERROR);
 
         if ($responseBody['State']['Running'] === true) {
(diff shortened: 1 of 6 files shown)


php/src/Docker/DockerActionManager.php
```diff
@@ -395,11 +396,21 @@ public function CreateContainer(Container $container): void {
...
+
+        // Special things for the caddy community container
         } elseif ($container->GetIdentifier() === 'nextcloud-aio-caddy') {
             $requestBody['HostConfig']['ExtraHosts'] = ['host.docker.internal:host-gateway'];
-            // Special things for the collabora container which should not be exposed in the containers.json
+
+        // Special things for the collabora container which should not be exposed in the containers.json
         } elseif ($container->GetIdentifier() === 'nextcloud-aio-collabora') {
+            // Load reference seccomp profile for collabora
+            $seccompProfile = file_get_contents(__DIR__ . '/../cool-seccomp-profile.json');
```
szaimen: Should probably retrieve the path from dataconst...

PR #7001, opened 2025-10-17
docs: Revise "how to use this" section of README

joshtrichards:
Reorganized and updated instructions for clarity and approachability.

No content was removed[^1], but I changed wording, added a couple extra details, and moved around a few things in a way that I thought made sense for clarity or approachability.

[^1]: Except for the manual Docker install via curl command; I linked to the full Docker install docs instead. People can decide see that there.

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -91,6 +91,7 @@ The steps below are written for Linux. For platform-specific guidance see:
 - macOS: [How to run AIO on macOS](#how-to-run-aio-on-macos)
 - Windows: [How to run AIO on Windows](#how-to-run-aio-on-windows)
 - Synology DSM: [How to run AIO on Synology DSM](#how-to-run-aio-on-synology-dsm)
+- TrueNAS SCALE: [Can I run AIO on TrueNAS SCALE?](#can-i-run-aio-on-truenas-scale)
 
 > [!IMPORTANT]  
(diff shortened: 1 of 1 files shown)


readme.md
```diff
@@ -86,66 +86,99 @@ Included are:
...
+> ```sh
+> sudo snap remove docker
+> ```
+
+2. If you need IPv6 support, enable it by following: [Docker IPv6 support for AIO](https://github.com/nextcloud/all-in-one/blob/main/docker-ipv6-support.md)
+
+3. AIO uses a special `mastercontainer` to orchestrate the various pieces of the Nextcloud stack. To start AIO, launch the `mastercontainer` with the command below:
 
 ```sh
-curl -fsSL https://get.docker.com | sudo sh
```
szaimen, line 103: What is the reason you removed this?

joshtrichards: I removed the line, along with old lines 96-97 and 99-100, because they all related to the Docker convenience script. 

Our existing warning for the script -- which essentially advises "use this, but don't use it" -- is confusing for readers, even if it is reasonable from a security perspective. 

Since Step 1 already links directly to the official Docker docs (which covers multiple installation methods as well as the convenience script in context), those lines were redundant.

szaimen: Ok makes sense

szaimen (approved):
LGTM, thanks! 😊

PR #7022, opened 2025-10-22
docs: restructure and clarify reverse proxy documentation (part one)

joshtrichards:
Updated the reverse proxy documentation for Nextcloud AIO, enhancing clarity and structure:

- Reorganized the top sections.
- Added sections to provide context on integrated versus external reverse proxies, as well as secure tunnels/proxy platforms.
- Deduplicated some content.
- General proofreading and other improvements.

No substantive content was dropped, but some was moved. If you notice a large block missing during review, please look for it elsewhere in the diff.

This was just a first pass, focused mostly on the top portion. Other areas were left unchanged to keep the work (and PR reviews) to a reasonable size.

### DIFF
--- a/reverse-proxy.md
+++ b/reverse-proxy.md
@@ -2,31 +2,54 @@
 
 ## Introduction
 
-This guide explains how to connect to Nextcloud AIO securely via HTTPS (TLS) using a reverse proxy or a secure tunnel/proxying platform. It covers:
+This guide explains how to connect to Nextcloud AIO securely via HTTPS (TLS) using a reverse proxy or a secure tunneling platform. It covers several potential scenarios:
 
-- Integrated: AIO's internal reverse proxy
-- External: an external reverse proxy such as Caddy or Nginx
-- Secure tunnel: a Zero Trust Network Access platform such as Tailscale or Cloudflare Tunnel
+- **Integrated**: AIO's built-in reverse proxy with automatic HTTPS
+- **External**: An external reverse proxy (such as Caddy or Nginx or Cloudflare Proxy)
+- **Secure tunnel**: Tunneling services for private network access or public access without port forwarding (such as Tailscale Serve or Cloudflare Tunnel)
+
+## Choosing Your Approach
 
 > [!TIP]
 > If AIO's internal reverse proxy meets your needs, you may not need to set up your own reverse proxy. See the next section to assess whether this is the case.
 
 > [!NOTE]
-> If your goal is to use AIO purely locally, refer to the [Local instance documentation](https://github.com/nextcloud/all-in-one/blob/main/local-instance.md).
+> If your goal is to use AIO purely locally, refer to the [Local instance documentation](https://github.com/nextcloud/all-in-one/blob/main/local-instance.md). Local instance setups don't require domain validation.
+
+### When to use each approach
+
+| Approach | Best for | Requirements | Inbound Ports Required |
+|----------|----------|--------------|---------------|
+| **Integrated** | Simple setups, single service on port 443 | Public IP, dedicated port 443 | Yes (443) |
+| **External Reverse Proxy** (including Cloudflare Proxy) | Multiple services, existing web server, or users wanting DDoS protection | Existing reverse proxy, willingness to set one up, or Cloudflare account | Yes (443) |
+| **Cloudflare Tunnel** | No port forwarding possible/desired, public access | Cloudflare account | No |
+| **Tailscale Serve** | Private access (tailnet only) | Tailscale account | No |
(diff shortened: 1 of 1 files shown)


reverse-proxy.md
```diff
@@ -1,67 +1,71 @@
...
 > [!NOTE]
-> Since the Apache container gets created by the mastercontainer, there is **NO** way to provide custom docker labels or custom environmental variables for the Apache container. So please do not attempt to do this because it will fail!
+> An external reverse proxy can also facilitate other routing approaches (for example shared-hostname / subfolder-based routing), but Nextcloud AIO only supports webroot-based (non-shared-hostname) access, so those scenarios are not applicable here.[^shared]
+
+### Secure tunnel: Using AIO with a Zero Trust Network Access platform (*Tailscale, Cloudflare*)
+
+Cloudflare and Tailscale provide Zero Trust Network Access services. For AIO we are primarily concerned with:
 
-## Content
+- Cloudflare Tunnel / Cloudflare Proxy
```
szaimen: Maybe we should point out that Cloudlare Tunel usually still exposes things externally. So I am not completely sure if it fits the category `Zero Trust Network Access platform`. Also Cloudflare Proxy does not fit in this category IMO

szaimen (commented):
Hey 👋

Thanks a lot for working on this! ☺️

In general, I am still wondering if we provide too much and too verbose information in this file but probably your are right and this is needed for people to get an overview first in oder to understand what to do with it. I just hope that people actually read it and not skip it if too much info is provided. But I also don't see a better way currently.

Apart from one point below, the restructuring looks good to me. Thanks again! ☺️

reverse-proxy.md
```diff
@@ -72,9 +76,19 @@ The process to run Nextcloud behind a reverse proxy consists of at least steps 1
...
+> Since the Apache container gets created by the mastercontainer, there is **NO** way to provide custom docker labels or custom environmental variables for the Apache container. So please do not attempt to do this because it will fail!
+
+### 1. Configure the reverse proxy
+
+#### Adapting the sample web server configurations below
 
-### Adapting the sample web server configurations below
 1. Replace `<your-nc-domain>` with the domain on which you want to run Nextcloud.
 1. Adjust the port `11000` to match your chosen `APACHE_PORT`.
 1. Adjust `localhost` or `127.0.0.1` to point to the Nextcloud server IP or domain depending on where the reverse proxy is running. See the following options.
```
szaimen, line 156: btw, I always have the feeling that people skip this part and directly head over to the example config. Maybe we should document the different possibilities in the config examples directly as comments? WDYT?

szaimen:
@joshtrichards I just found this. It seems AI-generated but WDYT about the general structure? See https://deepwiki.com/nextcloud/all-in-one/3.3-reverse-proxy-setup

joshtrichards:
> In general, I am still wondering if we provide too much and too verbose information in this file but probably your are right and this is needed for people to get an overview first in order to understand what to do with it.

Seems to be. Though I imagine we'll get better at deciding what to include/exclude over time -- or maybe get to a point where splitting things up further (i.e. into separate, more narrowly focused documents) makes sense for greater clarity or to avoid overwhelming the reader. Constant battle. Continuous learning and improvement I guess? ;-)

> I just hope that people actually read it and not skip it if too much info is provided. But I also don't see a better way currently.

Me neither.

> I just found this. It seems AI-generated but WDYT about the general structure? See https://deepwiki.com/nextcloud/all-in-one/3.3-reverse-proxy-setup

Looks interesting. I'll try to give it a closer look. Also reminds me that I do think some visual aids could be helpful (though also means more judgement about what to include in it hah).

P.S. Gave this PR another round of updates to try to address Cloudflare Tunnel versus Proxy matters. Let me know.

reverse-proxy.md
joshtrichards: >  So I am not completely sure if it fits the category Zero Trust Network Access platform. Also Cloudflare Proxy does not fit in this category IMO

Fair enough. I struggled with what to overly simplify versus what to get into for technical accuracy. Honestly I also get confused myself at times, particularly with Cloudflare's offerings! I took a another pass today to try to address these today. 

> Maybe we should point out that Cloudlare Tunel usually still exposes things externally. 

Hmm. Yeah at some point we've got to draw the line, assume the reader needs to take some responsibility for adjusting to their specific use case, and/or punt to Cloudflare (and/or Docker/etc) docs. But maybe there's still a spot where some additional warnings are warranted. Let me know if you see one after checking out the latest version.

szaimen (approved):
LGTM. Thank you very much! ☺️

PR #7032, opened 2025-10-24
clamav: change clamav-milter OnInfected Action to Reject

winterrific:
During testing the Beta with eicar virus test files, I noticed that Stalwart is not reliable respecting the Quarantine from clamav-milter. Even though the Clamav scanner was able to detect the test virus every time. 
Therefore, I recommend changing the OnInfected Action to Reject, what is in line with the configuration of Nextcloud Antivirus plugin which also rejects infected uploads.


szaimen:

PR #7098, opened 2025-11-11
increase API version to 1.44

szaimen:
- Close https://github.com/nextcloud/all-in-one/issues/7096

### DIFF
--- a/php/src/Docker/DockerActionManager.php
+++ b/php/src/Docker/DockerActionManager.php
@@ -13,7 +13,7 @@ use GuzzleHttp\Exception\RequestException;
 use http\Env\Response;
 
 readonly class DockerActionManager {
-    private const string API_VERSION = 'v1.41';
+    private const string API_VERSION = 'v1.44';
     private Client $guzzleClient;
 
     public function __construct(


Zoey2936:
I think that at least some changes mentioned here need to be changed: https://github.com/nextcloud/all-in-one/issues/7096#issuecomment-3515509772

szaimen:
> I think that at least some changes mentioned here need to be changed: [#7096 (comment)](https://github.com/nextcloud/all-in-one/issues/7096#issuecomment-3515509772)

I just tried it on my test server and it seems to work without any changes. Though I will run a more in-detail testing later

Zoey2936:
> I think that at least some changes mentioned here need to be changed: [#7096 (comment)](https://github.com/nextcloud/all-in-one/issues/7096#issuecomment-3515509772)

I have currently no time to check further

Zoey2936:
-  POST /networks/create with a duplicated name now fails systematically. As such, the CheckDuplicate field is now deprecated. Note that this change is unversioned and applied to all API versions on daemon that support version 1.44.

this change should be applied in any case as an example, not sure baout others

szaimen:
Thanks Zoey, will be done in https://github.com/nextcloud/all-in-one/pull/7100. I also had a fast look and I think we are not going to run into the mentioned problems. But will run extensive testing later on 👍

szaimen:

jparr721:
@szaimen confirming that this resolved the issue, at least on my setup.

DerKnerd:
@szaimen shouldn't this be a hotfix for 11 as well?

PR #7126, opened 2025-11-14
`ConnectContainerIdToNetwork`: switch from `Aliases` to  `DNSNames`

szaimen:
As mentioned here: https://github.com/nextcloud/app_api/pull/699#issue-3624297168

### DIFF
--- a/php/src/Docker/DockerActionManager.php
+++ b/php/src/Docker/DockerActionManager.php
@@ -860,7 +860,7 @@ readonly class DockerActionManager {
         );
         $jsonPayload = ['Container' => $id];
         if ($alias !== '') {
-            $jsonPayload['EndpointConfig'] = ['Aliases' => [$alias]];
+            $jsonPayload['EndpointConfig'] = ['DNSNames ' => [$alias]];
         }
 
         try {


apparle:
Is this really necessary? 
From what I read in changelog for v1.44 and v1.45, only the GET method's value for `Aliases` in the returned payload is changing.
We use the POST method, and I believe there's no change for that. I see even [API version v1.52](https://docs.docker.com/reference/api/engine/version/v1.52/#tag/Network/operation/NetworkConnect) supports `Aliases` in the payload

My key concern is that this may break podman compatibility; hard to locate the exact documentation on exact version compatibility.. Might be easier to stay on older `Aliases` field for maximum backward compatibility if it's still valid in newer docker engine versions.

szaimen:
> Is this really necessary? From what I read in changelog for v1.44 and v1.45, only the GET method's value for `Aliases` in the returned payload is changing. We use the POST method, and I believe there's no change for that. I see even [API version v1.52](https://docs.docker.com/reference/api/engine/version/v1.52/#tag/Network/operation/NetworkConnect) supports `Aliases` in the payload
> 
> My key concern is that this may break podman compatibility; hard to locate the exact documentation on exact version compatibility.. Might be easier to stay on older `Aliases` field for maximum backward compatibility if it's still valid in newer docker engine versions.

Thanks for the feedback! I agree to close this then...

PR #7127, opened 2025-11-14
allow to specify the `DOCKER_API_VERSION`

szaimen:
- Fix problems like https://github.com/nextcloud/all-in-one/issues/7096 systematically in the future
- This is best reviewed like this: https://github.com/nextcloud/all-in-one/pull/7127/files?w=1


Zoey2936:
I think there should be at least mentioned somehwere that this env doesn't make AIO magically work with all api versions and there may are incompatibilities

szaimen:
> I think there should be at least mentioned somehwere that this env doesn't make AIO magically work with all api versions and there may are incompatibilities

Agreed. I mentioned it already in the docs, I added. Feel free to review the files. Or should we somehow make this even more explicit?

szaimen:
The main problem that I want to fix is that apparently on Synology only v24 is currently supported which only supports up to 1.43 of the docker api. To not break it for all synology users without any workaround available, we need to somehow be able to change the api version even though we don't officially support or test any other than our default, so this is the best way I see going forward.

Zoey2936:
but would an older version then not break becuase of #7100 ?

szaimen:
> but would an older version then not break becuase of #7100 ?

Might be possible but I think it also works on older version without... Not sure why this setting was added anyway 😅

readme.md
```diff
@@ -427,6 +428,9 @@ Yes. If SELinux is enabled, you might need to add the `--security-opt label:disa
 
 ## Customization
 
+### How to adjust the internally used docker api version?
+If you run an outdated or too new docker version, you might run into problems with the by AIO internally used docker version. To fix this, you can specify the api version manually. You can do so by adding `--env DOCKER_API_VERSION=1.44` to the docker run command of the mastercontainer (but before the last line `ghcr.io/nextcloud-releases/all-in-one:latest`! If it was started already, you will need to stop the mastercontainer, remove it (no data will be lost) and recreate it using the docker run command that you initially used). This variable excepts a string based on the pattern `[0-9].[0-9]+`, so e.g. `1.44`. ⚠️ However please note that only the default api version (unset this variable) is supported and tested by the maintainers of Nextcloud AIO. So use this on your own risk and things might break without warning.
```
```suggestion
If you run an outdated or too new docker version, you might run into problems with the by AIO internally used docker api version. To fix this, you can specify the api version manually. You can do so by adding `--env DOCKER_API_VERSION=1.44` to the docker run command of the mastercontainer (but before the last line `ghcr.io/nextcloud-releases/all-in-one:latest`! If it was started already, you will need to stop the mastercontainer, remove it (no data will be lost) and recreate it using the docker run command that you initially used). This variable excepts a string based on the pattern `[0-9].[0-9]+`, so e.g. `1.44`. ⚠️ However please note that only the default api version (unset this variable) is supported and tested by the maintainers of Nextcloud AIO. So use this on your own risk and things might break without warning.
```

szaimen:
This is now released with v12.1.0 Beta. Testing and feedback is welcome! See https://github.com/nextcloud/all-in-one#how-to-switch-the-channel

PR #7128, opened 2025-11-14
fix whiteboard recording chrome #3

hweihwang:
- running `apk upgrade --no-cache -a` and `chmod 777 -R /tmp` after pulling the upstream image that wipes out the secure `/tmp/chromium-runtime`/`chrome-crashpad` directories the upstream build prepares for Puppeteer, so Chromium never exposes its WS endpoint and all recording checks time out (10 s error dialog).
- only install `bash`, leave `/tmp` permissions alone, and only replace the crashpad handler when the upstream image hasn’t already patched it. Also keep the upstream working directory (`/app`), since the base image expects that.

Please have it a look and test @szaimen, thank you!

### DIFF
--- a/Containers/whiteboard/Dockerfile
+++ b/Containers/whiteboard/Dockerfile
@@ -5,6 +5,7 @@ FROM ghcr.io/nextcloud-releases/whiteboard:v1.4.1
 USER root
 RUN set -ex; \
     apk add --no-cache bash; \
+    chmod 777 -R /tmp; \
     if [ -f /usr/lib/chromium/chrome_crashpad_handler ] && [ ! -f /usr/lib/chromium/chrome_crashpad_handler.real ]; then \
         mv /usr/lib/chromium/chrome_crashpad_handler /usr/lib/chromium/chrome_crashpad_handler.real; \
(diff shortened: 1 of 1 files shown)


Containers/whiteboard/Dockerfile
```diff
@@ -20,7 +17,7 @@ COPY --chmod=775 healthcheck.sh /healthcheck.sh
 
 HEALTHCHECK CMD /healthcheck.sh
 
-WORKDIR /tmp
+WORKDIR /app
```
szaimen:
```suggestion
WORKDIR /tmp
```

szaimen (dismissed):
LGTM, thanks!

szaimen:
> * only install `bash`, leave `/tmp` permissions alone, and only replace the crashpad handler when the upstream image hasn’t already patched it. Also keep the upstream working directory (`/app`), since the base image expects that.

@hweihwang unfortunately there is a problem with that. If you look at https://github.com/nextcloud/all-in-one/blob/f75562992a0023fd4e3662ff0037a3d137154201/manual-install/latest.yml#L458-L486, we run the container with read-only root fs. Only readable is the /tmp directory where we mount a tmpfs. Maybe we need to adjust this somehow?

PR #7153, opened 2025-11-20
downgrade internal docker to fix incompatibility with synology DSM

szaimen:
- Fix https://github.com/nextcloud/all-in-one/issues/7096#issuecomment-3558413666


Zoey2936:
in theory this could also be changed via env: https://docs.docker.com/reference/cli/docker/version/#api-version-and-version-negotiation

szaimen:
> in theory this could also be changed via env: https://docs.docker.com/reference/cli/docker/version/#api-version-and-version-negotiation

You are right. I think this did not work because sudo does not keep the envs. See https://github.com/nextcloud/all-in-one/pull/7155

PR #7179, opened 2025-11-24
whiteboard: create required chrome directories on the fly

szaimen:
- Follow-up to https://github.com/nextcloud/all-in-one/pull/7128


szaimen:
Seems like it works now

PR #7187, opened 2025-11-24
apache: enable proxy protocol

szaimen:
- Close https://github.com/nextcloud/all-in-one/discussions/4438

### DIFF
--- a/Containers/apache/Caddyfile
+++ b/Containers/apache/Caddyfile
@@ -7,6 +7,16 @@
 
     servers {
         # trusted_proxies placeholder
+
+        # Enable proxy protocol for caddy
+        listener_wrappers {
+            proxy_protocol {
+                timeout 5s
+                allow private_ranges 100.64.0.0/10
+                fallback_policy skip
+            }
+            tls
+        }
     }
 
     log {


Containers/apache/Caddyfile
```diff
@@ -7,6 +7,16 @@
 
     servers {
         # trusted_proxies placeholder
+
+        # Enable proxy protocol for caddy
+        listener_wrappers {
+            proxy_protocol {
+                timeout 5s
+                allow private_ranges 100.64.0.0/10
```
szaimen, line 15: need to check if `private_ranges` works here.

Containers/apache/Caddyfile
```diff
@@ -7,6 +7,16 @@
...
+                fallback_policy skip
+            }
+            tls
+        }
```
szaimen, line 19: Maybe we should only enable this when AIO runs in reverse proxy mode?

Zoey2936:
Is there a reason for this? proxy_protcoll makes sense for protocols which don't support things like X-Forwarded headers, but http(s) does this, I don't see any reason for this

Containers/apache/Caddyfile
Zoey2936: it does not:
`Error: sending configuration to instance: caddy responded with error: HTTP 400: {"error":"loading config: loading new config: loading http app module: provision http: loading listener wrapper modules: position 0: loading module 'proxy_protocol': provision caddy.listeners.proxy_protocol: netip.ParsePrefix(\"private_ranges\"): no '/'"}`

Containers/apache/Caddyfile
Zoey2936: I think it should be reversed, so this is only enabled if AIO is **not** in reverse proxy mode, so AIO can still handle tls/acme even if a proxy runs in front of it

szaimen:
> Is there a reason for this? proxy_protcoll makes sense for protocols which don't support things like X-Forwarded headers, but http(s) does this, I don't see any reason for this

The main reason is this request: https://github.com/nextcloud/all-in-one/discussions/4438

Zoey2936:
Yes, but what are use cases?
reverse proxy mode currently:
Users <---https---> proxy <---http with X-Forwarded---> nextcloud
with proxy protocol in reversr proxy mode:
Users <---https---> proxy <---proxy protocol packets containing http---> nextcloud
It has no real advantage, since the proxy still needs to do tls (even if the proxy doesn't work with the application level)

without reverse proxy currently:
Users <---https--> nextcloud
without reverse proxy but with proxy_procol:
Users <---https---> proxy <---proxy protocol packets containing https---> nextcloud
In this case the proxy just passes the raw https packages to nextcloud and nextcloud responses to the users, so tls is fully handled by caddy, which would be a user case and it would also allow the proxy to route the traffic based on the SNI (as long as AIO doesn't use ECH)

winterrific:
Indeed, routing will be possible, but the original IP will be lost with a l4 proxy. That's where proxy_protocol steps in. In the case of @redyolocraft HA_Proxy is forwarding the tls packages on layer4, without termination of tls, tough not able to add any additional headers. So the user's IP got lost on the way to nextcloud-container.

So, you're right that this is a special use case where caddy shall handle the tls (incl. acme and certs) and there is another (l4) proxy in front of it. 

How about introducing an ENV to configure the allowed IP source range? This ENV could be mapped on `allow private_ranges {$PROXY_IP_RANGE}` and only if this ENV is available proxy_protcol listener is enabled. 

```
if [ "${PROXY_IP_RANGE}" ]; then
    cat << PROXY > /tmp/proxy.config
    listener_wrappers {
            proxy_protocol {
                timeout 5s
                allow private_ranges {$PROXY_IP_RANGE}
                fallback_policy skip
            }
          tls
    }
PROXY
    CADDYFILE="$(sed "/proxy_listener/r /tmp/proxy.config" /tmp/Caddyfile)"
    echo "$CADDYFILE" > /tmp/Caddyfile
fi

```

Zoey2936:
I think this would be the only use case, if Nextcloud/Caddy fully handles TLS and the proxy just passes the encrypted data through via proxy protocol without terminating TLS itself. It would keep the Client IP, routing based on SNI (if ECH is not used) is possible and TLS could still be handled by Nextcloud/Caddy while running a proxy in front of it. This use case should be rare, so I think an ENV like @winterrific mentioned would be the best solution for this, but it only makes sense if AIO does **not** run in reverse proxy mode, and the ENV should also be required since `private_ranges` doesn't work in this case. (It would also work with in reverse proxy mode, but it doesn't make much sense, since http+headers or proxy protocol+http has the same result)

szaimen:
Thanks foe all the feedback @Zoey2936 and @winterrific! 😊

The more I think about this, I get to the point where this probably would make most sense to be implemented in the caddy community container. WDYT about it? See https://github.com/szaimen/aio-caddy

szaimen:
Superceded by https://github.com/szaimen/aio-caddy/pull/100

PR #7196, opened 2025-11-25
helm: adjust naming of variables

szaimen:
- See https://github.com/nextcloud/all-in-one/discussions/7628


szaimen:
- [ ] TODO: check if we can add fall-backs to old variables
- [ ] TODO: automated the renaming of vairables

Maximilian-Maag:
Perhaps a simple bash script could solve this problem? It could read the current key and rewrite it with the new syntax as some kind of migration script for exisiting intances?

PR #7248, opened 2025-12-02
Update psalm.yml

szaimen:


.github/workflows/psalm.yml
```diff
@@ -10,8 +10,6 @@ name: Static analysis
 
 on:
   pull_request:
-    paths:
-      - 'php/**'
```
szaimen, line 14: Revert

szaimen:
Should be obsolete with https://blog.packagist.com/composer-2-9/

PR #7257, opened 2025-12-03
replace int cast with intval function

szaimen:


szaimen:
I just read about casting vs using the function a bit more and it looks like casting is generally faster so keep using that.

PR #7285, opened 2025-12-08
make redis port configurable

szaimen:


lukbukkit:
I think this PR broke my existing Nextcloud AIO setup in some way. This night, my Nextcloud went offline. The containers Nextcloud, Nextcloud Push, and Whiteboard were stuck in a starting state trying to connect with Redis (see the following log) while the Redis container was up and healthy.
```
2025-12-19T08:22:54.853408915Z nc: service "" unknown
2025-12-19T08:22:54.853540693Z Waiting for Redis to start...
```
The solution was to stop all Nextcloud AIO containers manually via the command line and let the containers update. Now everything seems to be working again.

PR #7286, opened 2025-12-08
rp-docs: add encoded characters config for Traefik v3.6.4+

stefano99:
Add HTTP settings for encoded URL characters in Traefik. This fixes an issue with Collabora not being able to open files, specifically giving the error:

> Failed to establish socket connection or socket connection closed unexpectedly. The reverse proxy might be misconfigured, please contact the administrator. For more info on proxy configuration please checkout https://sdk.collaboraonline.com/docs/installation/Proxy_settings.html


stefano99:
I'll edit the PR and swap the tabs for the spaces.

Regarding the http key, it's used to define http options, so I don't think it exists a https key.
I tried substituting it with https in my config and traefik didn't start.
In my config, I also have _tls: true_ added under the http key for the https entrypoint (also called websecure in some guides), to enable tls on every router added to that entrypoint. In the docs it's added in the router (lines 902 in main or 907 in pr), so I believe it's the same: in both cases, the connection is https.

stefano99:
Ok I should've removed all the tabs and used only spaces.
Sorry for the mess, but it's my first contribution and i'm not used to github!

szaimen (approved):
LGTM, thanks! ☺️

PR #7362, opened 2025-12-30
Add COLLABORA_EXTRA_HOSTS option

kw217:
This PR adds the ability to add extra entries to /etc/hosts on the nextcloud-aio-collabora container.

I need this to work around an issue that I haven't yet fully resolved, to do with Collabora (in response to a request coming via coolwsd) attempting to retrieve config from my instance's external domain, rather than the internal domain. (This may be https://github.com/nextcloud/richdocuments/issues/4906.) I intend to properly debug this and provide a PR later. But this functionality could be useful to others, e.g., https://github.com/nextcloud/all-in-one/discussions/2395.

I didn't understand the function of `update-yaml.sh`, nor do I understand what the `configName` is for, so please let me know if I have missed anything.

This would probably be better if it was generalised to allow specifying for any container rather than only nextcloud-aio-collabora. But I didn't know the best way to represent this in config.

Thanks!


kw217:
Thanks for the nudge. Going through that again and thinking some more means I may have got a step closer. See my original post and my update in this thread: https://github.com/nextcloud/richdocuments/issues/4906#issuecomment-3700504607 . Thanks.

kw217:
@szaimen if you would prefer to avoid the `COLLABORA_EXTRA_HOSTS` approach, then I'd like to have a way to set the `net.proto=IPv6` for Collabora under AIO.

Looking at containers.json in AIO, it looks like there's no way to pass arbitrary options from compose.yaml to Collabora - the only configurable option is currently the seccomp one.

I can think of a couple of approaches, both of which could be done as modifications of this present PR:

1. Provide a general `COLLABORA_OPTIONS` env var.
2. Provide a specific `COLLABORA_NET_PROTO` which is enforced to be one of all/IPv4/IPv6.

Or perhaps there is some way I can get my system working without any of these changes.

Happy to proceed on whichever of these paths would be preferable to you.

szaimen:
There is https://github.com/nextcloud/all-in-one/blob/e8d1bce4f68702de1a569204c36171c16c1df5fe/php/templates/includes/optional-containers.twig#L180-L191. Maybe that works for you?

kw217:
It didn't help - I think the `net.proto=IPv6` thing was just coincidence, and did not fix it. So I'm back to needing the extra host added. For now I've set up a cron job on the host which does the following, but it's not a very pretty solution. Long term I would like to fix the code so it doesn't try to retrieve the config from the external domain, but I'm not sure if I'll be able to. In the short term, my original request would help - but for now I'll just keep doing this.
```
docker exec -u root nextcloud-aio-collabora bash -c 'grep cloud.example.org /etc/hosts || echo 2001:db8::1 cloud.example.org >> /etc/hosts'
```
(IP and hostname redacted obviously).

kw217:
Cool, I'll try again once I have #7481  - thank you!

szaimen:
> Cool, I'll try again once I have #7481 - thank you!

PR #7388, opened 2026-01-06
Trigger config saving manually and add more setter methods

pabzm:
This requires some additional code lines but allows the calling code to group calls to setter methods and only save them at once.

`ConfigurationManager->set()` is a public method because `DockerActionManager->CreateContainer()` acts on a dynamic list of variables, which I can't (or don't understand how to) map to setter methods.

I only refactored code I had to touch anyways, there's more low-hanging fruits I'd post in a later PR (e.g. lower-casing method names, or using a `get()` helper method in `ConfigurationManager`).

Closes #7239

PR #7391, opened 2026-01-07
Private GetConfig and WriteConfig by providing generic set() method

pabzm:
This is an alternative to #7388 

The first commit is the minimum required change to allow privatizing `GetConfig()` and `WriteConfig()`, the second one exemplifies how some setter methods could be removed and replaced by calls to `set()`.


pabzm:
Made this a draft so it doesn't get accidentally merged until it would be actually ready.

pabzm:
Actually we could introduce a list of allowed config keys and use it as a filter to also check the expected type.

```
    private array $configKeys = [
        'isDockerSocketProxyEnabled' => 'boolean',
    ];
```

That in a way is rebuilding language-features manually, but it would work.

pabzm:
Closing in favour of https://github.com/nextcloud/all-in-one/pull/7392

Issue #7239, opened 2025-12-01
make `WriteConfig` and `GetConfig` private


szaimen:
The `WriteConfig` and `GetConfig` functions are currently accessed in some files outside ConfigurationManager. In best case the functions should only be used by class internal functions. 

See 
- https://github.com/search?q=repo%3Anextcloud%2Fall-in-one+WriteConfig+-path%3A%2F%5Ephp%5C%2Fsrc%5C%2FData%5C%2F%2F&type=code
- https://github.com/search?q=repo%3Anextcloud%2Fall-in-one+GetConfig+-path%3A%2F%5Ephp%5C%2Fsrc%5C%2FData%5C%2F%2F&type=code

PR #7392, opened 2026-01-07
Use property hooks to replace classic getter and setter methods.

pabzm:
- Close https://github.com/nextcloud/all-in-one/issues/7239

### DIFF
--- a/php/get-configurable-aio-variables.sh
+++ b/php/get-configurable-aio-variables.sh
@@ -1,3 +1,3 @@
 #!/usr/bin/env bash
 
-awk '/^    public [^f][^u][^n]/ { sub(/\$/, "", $3); print $3 }' php/src/Data/ConfigurationManager.php | sort
+awk '/^    public [^f][^u][^n]/ { sub(/\$/, "", $3); print $3 }' src/Data/ConfigurationManager.php | sort


pabzm:
Actually we can combine the property hooks style of this PR with the generic `setMultiple()` method from #7391 – but we would loose the power to restrict which variables can be written to the config file.

pabzm:
Made this a draft so it doesn't get accidentally merged until it would be actually ready.

pabzm:
This is now finally ready for review. I made some effort to produce comprehensible and atomic commits, so you can review those step by step if the big picture is too crowded.

php/src/Data/ConfigurationManager.php
```diff
@@ -654,23 +654,21 @@ public function WriteConfig(?array $config) : void {
...
-            if (!isset($config[$configName])) {
-                $config[$configName] = '';
-            }
-            if ($envVariableOutput !== $config[$configName]) {
-                $config[$configName] = $envVariableOutput;
-                $this->WriteConfig($config);
+
+        if (file_exists(DataConst::GetConfigFile())) {
+            if ($envVariableOutput !== $configValue) {
+                $this->set($configName, $envVariableOutput);
```
szaimen, line 704: I am not sure about this part, I think this changes the behaviour

php/src/Data/ConfigurationManager.php
```diff
@@ -13,6 +13,11 @@ class ConfigurationManager
 
     private bool $noWrite = false;
 
+    public string $AIO_TOKEN {
```
szaimen: can we please use camel case?
```suggestion
    public string $AioToken {
```

php/src/Data/ConfigurationManager.php
```diff
@@ -23,6 +23,11 @@ class ConfigurationManager
         set { $this->set('password', $value); }
     }
 
+    public bool $wasStartButtonClicked {
+        get => $this->get('wasStartButtonClicked', false);
```
szaimen: What if wasStartButtonClicked is currently set to `1`? Will this fail because it does not return a bool but an int?

php/src/Data/ConfigurationManager.php
```diff
@@ -23,6 +23,11 @@ class ConfigurationManager
         set { $this->set('password', $value); }
     }
 
+    public string $AIO_URL {
```
szaimen: can we please use camel case?
```suggestion
    public string $AioUrl {
```

php/src/Controller/DockerController.php
```diff
@@ -181,16 +181,7 @@ public function StartContainer(Request $request, Response $response, array $args
...
-            $installLatestMajor = 32;
-        } else {
-            $installLatestMajor = "";
-        }
-
-        $config = $this->configurationManager->GetConfig();
-        // set install_latest_major
-        $config['install_latest_major'] = $installLatestMajor;
-        $this->configurationManager->WriteConfig($config);
+        $this->configurationManager->install_latest_major = isset($request->getParsedBody()['install_latest_major']);
```
szaimen: Can we please keep writing the version number to the config file? We might want to use that later on in the UI... it would probably make most sense to set the actual value in the web interface? See https://github.com/nextcloud/all-in-one/blob/4d51f11282b7dafe871af32e6d66585b3b2b6ff8/php/templates/containers.twig#L346

php/src/Data/ConfigurationManager.php
```diff
@@ -33,6 +33,11 @@ class ConfigurationManager
         set { $this->set('wasStartButtonClicked', $value); }
     }
 
+    public bool $install_latest_major {
```
szaimen: can we please use camel case?
```suggestion
    public bool $installLatestMajor {
```

php/src/Data/ConfigurationManager.php
```diff
@@ -33,6 +33,11 @@ class ConfigurationManager
...
+        get => $this->get('install_latest_major', false);
```
szaimen: Additionally what happens if we return an int or empty string here like how it is currently set on exisitng instances?

szaimen: Btw, this is a behavioural change. Please check the logic in https://github.com/nextcloud/all-in-one/blob/4d51f11282b7dafe871af32e6d66585b3b2b6ff8/php/src/Data/ConfigurationManager.php#L865-L871

php/src/Data/ConfigurationManager.php
```diff
@@ -23,6 +23,11 @@ class ConfigurationManager
         set { $this->set('password', $value); }
     }
 
+    public bool $restoreExcludePreviews {
+        get => $this->get('restore-exclude-previews', false);
```
szaimen: What happens if an int or empty string is stored on the backend for existing instances?

php/src/Data/ConfigurationManager.php
```diff
@@ -23,6 +23,11 @@ class ConfigurationManager
         set { $this->set('password', $value); }
     }
 
+    public bool $isWhiteboardEnabled {
+        get => $this->get('isWhiteboardEnabled', true);
```
szaimen: what happens with existing instances where an int is set?

php/src/Data/ConfigurationManager.php
```diff
@@ -23,6 +23,11 @@ class ConfigurationManager
         set { $this->set('password', $value); }
     }
 
+    public bool $isDockerSocketProxyEnabled {
+        get => $this->get('isDockerSocketProxyEnabled', false);
```
szaimen: what happens with existing instances where an int is set?

php/src/Docker/DockerActionManager.php
```suggestion
            'BORGBACKUP_MODE' => $this->configurationManager->backupMode,

```

php/src/Data/ConfigurationManager.php
```diff
@@ -48,6 +48,11 @@ class ConfigurationManager
         set { $this->set('backup-mode', $value); }
     }
 
+    public bool $instance_restore_attempt {
+        get => $this->get('instance_restore_attempt', false);
```
szaimen: What happens if an int is returned?

php/src/Data/ConfigurationManager.php
```diff
@@ -48,6 +48,11 @@ class ConfigurationManager
         set { $this->set('backup-mode', $value); }
     }
 
+    public bool $instance_restore_attempt {
```
szaimen: can we please use camel case?
```suggestion
    public bool $instanceRestoreAttempt {
```

php/src/Data/ConfigurationManager.php
```diff
@@ -68,6 +68,11 @@ class ConfigurationManager
         set { $this->set('install_latest_major', $value); }
     }
 
+    public bool $isClamavEnabled {
+        get => $this->get('isClamavEnabled', false);
```
szaimen: What happens if an int is returned?

php/src/Data/ConfigurationManager.php
```diff
@@ -73,6 +73,11 @@ class ConfigurationManager
         set { $this->set('isClamavEnabled', $value); }
     }
 
+    public bool $isOnlyofficeEnabled {
+        get => $this->get('isOnlyofficeEnabled', false);
```
szaimen: What happens if an int is returned?

php/src/Data/ConfigurationManager.php
```diff
@@ -78,6 +78,11 @@ class ConfigurationManager
         set { $this->set('isOnlyofficeEnabled', $value); }
     }
 
+    public bool $isCollaboraEnabled {
+        get => $this->get('isCollaboraEnabled', true);
```
szaimen: What happens if an int is returned?

php/src/Data/ConfigurationManager.php
```diff
@@ -83,6 +83,11 @@ class ConfigurationManager
         set { $this->set('isCollaboraEnabled', $value); }
     }
 
+    public bool $isTalkEnabled {
+        get => $this->get('isTalkEnabled', true);
```
szaimen: What happens if an int is returned?

php/src/Data/ConfigurationManager.php
```diff
@@ -88,6 +88,11 @@ class ConfigurationManager
         set { $this->set('isTalkEnabled', $value); }
     }
 
+    public bool $isTalkRecordingEnabled {
+        get => $this->isTalkEnabled && $this->get('isTalkRecordingEnabled', false);
```
szaimen: What happens if an int is returned?

php/src/Data/ConfigurationManager.php
```diff
@@ -93,6 +93,11 @@ class ConfigurationManager
         set { $this->set('isTalkRecordingEnabled', $this->isTalkEnabled && $value); }
     }
 
+    public bool $isImaginaryEnabled {
+        get => $this->get('isImaginaryEnabled', true);
```
szaimen: What happens if an int is returned?

php/src/Data/ConfigurationManager.php
```diff
@@ -98,6 +98,12 @@ class ConfigurationManager
         set { $this->set('isImaginaryEnabled', $value); }
     }
 
+    public bool $isFulltextsearchEnabled {
+        get => $this->get('isFulltextsearchEnabled', false);
```
szaimen: What happens if an int is returned?

php/src/Data/ConfigurationManager.php
szaimen, line 668: Why cant    we use `$this->password = $newPassword` here?

php/src/Data/ConfigurationManager.php
```diff
@@ -109,6 +109,11 @@ class ConfigurationManager
         set { $this->SetDomain($value); }
     }
 
+    public string $borg_backup_host_location {
```
szaimen: can we please use camel case?
```suggestion
   public string $borgBackupHostLocation {
```

php/src/Data/ConfigurationManager.php
```diff
@@ -114,6 +114,11 @@ class ConfigurationManager
         set { $this->set('borg_backup_host_location', $value); }
     }
 
+    public string $borg_remote_repo {
```
szaimen: can we please use camel case?
```suggestion
    public string $borgRemoteRepo {
```

php/src/Data/ConfigurationManager.php
```diff
@@ -119,6 +119,11 @@ class ConfigurationManager
         set { $this->set('borg_remote_repo', $value); }
     }
 
+    public string $borg_restore_password {
```
szaimen: can we please use camel case?
```suggestion
    public string $borgRestorePassword {
```

php/src/Data/ConfigurationManager.php
```diff
@@ -465,12 +470,11 @@ public function SetBorgRestoreLocationVarsAndPassword(string $location, string $
         }
 
         $config = $this->GetConfig();
-        $config['borg_restore_password'] = $password;
         $this->WriteConfig($config);
```
szaimen: I guess we can remove this?

php/src/Data/ConfigurationManager.php
```diff
@@ -136,6 +136,18 @@ class ConfigurationManager
         }
     }
 
+    /**
+     * @throws InvalidSettingConfigurationException
+     */
+    public string $collabora_dictionaries {
```
szaimen: can we please use camel case?
```suggestion
    public string $collaboraDictionaries {
```

php/src/Data/ConfigurationManager.php
```diff
@@ -148,6 +148,18 @@ class ConfigurationManager
         }
     }
 
+    /**
+     * @throws InvalidSettingConfigurationException
+     */
+    public string $collabora_additional_options {
```
szaimen: can we please use camel case?
```suggestion
    public string $collaboraAdditionalOptions {
```

php/src/Data/ConfigurationManager.php
```diff
@@ -160,6 +160,11 @@ class ConfigurationManager
         }
     }
 
+    public array $aio_community_containers {
```
szaimen: can we please use camel case?
```suggestion
    public array $AioCommunityContainers {
```

php/src/Data/ConfigurationManager.php
```diff
@@ -165,6 +165,11 @@ class ConfigurationManager
         set { $this->set('aio_community_containers', implode(' ', $value)); }
     }
 
+    public string $turn_domain {
```
szaimen: can we please use camel case?
```suggestion
    public string $turnDomain {
```

php/src/Data/ConfigurationManager.php
```diff
@@ -124,6 +124,11 @@ class ConfigurationManager
         set { $this->set('borg_restore_password', $value); }
     }
 
+    public string $apache_ip_binding {
+        get => $this->GetEnvironmentalVariableOrConfig('APACHE_IP_BINDING', 'apache_ip_binding', '');
+        set { $this->set('apache_ip_binding', $value); }
```
szaimen, line 142: do we need the set function here?

php/src/Data/ConfigurationManager.php
```diff
@@ -124,6 +124,11 @@ class ConfigurationManager
         set { $this->set('borg_restore_password', $value); }
     }
 
+    public string $apache_ip_binding {
```
szaimen: can we please use camel case?
```suggestion
    public string $apacheIpBinding{
```

php/src/Data/ConfigurationManager.php
```diff
@@ -558,11 +558,9 @@ public function ChangeMasterPassword(string $currentPassword, string $newPasswor
...
     }
 
-    public function GetApachePort() : string {
-        $envVariableName = 'APACHE_PORT';
-        $configName = 'apache_port';
-        $defaultValue = '443';
-        return $this->GetEnvironmentalVariableOrConfig($envVariableName, $configName, $defaultValue);
+    public string $apache_port {
+        get => $this->GetEnvironmentalVariableOrConfig('APACHE_PORT', 'apache_port', '443');
+        set { $this->set('apache_port', $value); }
```
szaimen: Do we need the set here?

php/src/Data/ConfigurationManager.php
```diff
@@ -558,11 +558,9 @@ public function ChangeMasterPassword(string $currentPassword, string $newPasswor
         $this->set('password', $newPassword);
     }
 
-    public function GetApachePort() : string {
-        $envVariableName = 'APACHE_PORT';
-        $configName = 'apache_port';
-        $defaultValue = '443';
-        return $this->GetEnvironmentalVariableOrConfig($envVariableName, $configName, $defaultValue);
+    public string $apache_port {
```
szaimen: can we please use camel case?
```suggestion
    public string $apachePort {
```

php/src/Data/ConfigurationManager.php
```diff
@@ -562,12 +562,10 @@ public function ChangeMasterPassword(string $currentPassword, string $newPasswor
...
         set { $this->set('apache_port', $value); }
     }
-
-    public function GetTalkPort() : string {
-        $envVariableName = 'TALK_PORT';
-        $configName = 'talk_port';
-        $defaultValue = '3478';
-        return $this->GetEnvironmentalVariableOrConfig($envVariableName, $configName, $defaultValue);
+        
+    public string $talk_port {
```
szaimen: can we please use camel case?
```suggestion
    public string $talkPort {
```

php/src/Data/ConfigurationManager.php
```diff
@@ -634,13 +634,10 @@ public function GetNextcloudDatadirMount() : string {
         return $this->GetEnvironmentalVariableOrConfig($envVariableName, $configName, $defaultValue);
     }
 
-    public function GetNextcloudUploadLimit() : string {
-        $envVariableName = 'NEXTCLOUD_UPLOAD_LIMIT';
-        $configName = 'nextcloud_upload_limit';
-        $defaultValue = '16G';
-        return $this->GetEnvironmentalVariableOrConfig($envVariableName, $configName, $defaultValue);
+    public string $nextcloud_upload_limit {
```
szaimen: can we please use camel case?
```suggestion
    public string $nextcloudUploadLimit {
```

php/src/Data/ConfigurationManager.php
```diff
@@ -634,13 +634,10 @@ public function GetNextcloudDatadirMount() : string {
...
+        get => $this->GetEnvironmentalVariableOrConfig('NEXTCLOUD_UPLOAD_LIMIT', 'nextcloud_upload_limit', '16G');
+        set { $this->set('nextcloud_upload_limit', $value); }
```
szaimen: do we need the set method here?

php/src/Data/ConfigurationManager.php
```diff
@@ -620,11 +620,9 @@ public function GetBorgPublicKey() : string {
         return trim((string)file_get_contents(DataConst::GetBackupPublicKey()));
     }
 
-    public function GetNextcloudMount() : string {
-        $envVariableName = 'NEXTCLOUD_MOUNT';
-        $configName = 'nextcloud_mount';
-        $defaultValue = '';
-        return $this->GetEnvironmentalVariableOrConfig($envVariableName, $configName, $defaultValue);
+    public string $nextcloud_mount {
```
szaimen: can we please use camel case?
```suggestion
    public string $nextcloudMount {
```

php/src/Data/ConfigurationManager.php
```diff
@@ -625,11 +625,10 @@ public function GetBorgPublicKey() : string {
         set { $this->set('nextcloud_mount', $value); }
     }
 
-    public function GetNextcloudDatadirMount() : string {
-        $envVariableName = 'NEXTCLOUD_DATADIR';
-        $configName = 'nextcloud_datadir';
-        $defaultValue = 'nextcloud_aio_nextcloud_data';
-        return $this->GetEnvironmentalVariableOrConfig($envVariableName, $configName, $defaultValue);
+
+    public string $nextcloud_datadir_mount {
```
szaimen: can we please use camel case?
```suggestion
    public string $nextcloudDatadirMount {
```

php/src/Data/ConfigurationManager.php
```diff
@@ -635,11 +635,10 @@ public function GetBorgPublicKey() : string {
         get => $this->GetEnvironmentalVariableOrConfig('NEXTCLOUD_UPLOAD_LIMIT', 'nextcloud_upload_limit', '16G');
         set { $this->set('nextcloud_upload_limit', $value); }
     }
-    public function GetNextcloudMemoryLimit() : string {
-        $envVariableName = 'NEXTCLOUD_MEMORY_LIMIT';
-        $configName = 'nextcloud_memory_limit';
-        $defaultValue = '512M';
-        return $this->GetEnvironmentalVariableOrConfig($envVariableName, $configName, $defaultValue);
+    
+    public string $nextcloud_memory_limit {
```
szaimen: can we please use camel case?
```suggestion
    public string $nextcloudMemoryLimit {
```

php/src/Data/ConfigurationManager.php
```diff
@@ -646,11 +646,9 @@ public function GetApacheMaxSize() : int {
         return $uploadLimit * 1024 * 1024 * 1024;
     }
 
-    public function GetNextcloudMaxTime() : string {
-        $envVariableName = 'NEXTCLOUD_MAX_TIME';
-        $configName = 'nextcloud_max_time';
-        $defaultValue = '3600';
-        return $this->GetEnvironmentalVariableOrConfig($envVariableName, $configName, $defaultValue);
+    public string $nextcloud_max_time {
```
szaimen: can we please use camel case?
```suggestion
    public string $nextcloudMaxTime {
```

php/get-configurable-aio-variables.sh
```diff
@@ -0,0 +1,3 @@
+#!/usr/bin/env bash
+
+awk '/^    public [^f][^u][^n]/ { sub(/\$/, "", $3); print $3 }' php/src/Data/ConfigurationManager.php | sort
```
szaimen: Can we please put the script in into the php subfolder? https://github.com/nextcloud/all-in-one/tree/main/php

szaimen (changes requested):
First of all, thank you very much! This is a tremendous improvement to the code base imho ☺️

I just went through all the comments and added some comments:

szaimen (changes requested):
First of all, thank you very much! This is a tremendous improvement to the code base imho ☺️

I just went through all the comments and added some comments.

szaimen:
Btw, we have some conflicts unfortunately

szaimen:
Btw @pabzm we still have some config keys that could possibly get migrated? See https://github.com/nextcloud/all-in-one/blob/3dde66f2e367a35a777417c939133575bac03f6d/php/src/Data/ConfigurationManager.php#L642-L701 and https://github.com/nextcloud/all-in-one/blob/3dde66f2e367a35a777417c939133575bac03f6d/php/src/Data/ConfigurationManager.php#L877-L897 and https://github.com/nextcloud/all-in-one/blob/3dde66f2e367a35a777417c939133575bac03f6d/php/src/Data/ConfigurationManager.php#L934-L973

php/src/Controller/DockerController.php
pabzm: Yes, good point. I chose to make it a `string`, so we have more freedom what to write into the option.

php/src/Data/ConfigurationManager.php
pabzm: Implemented as `aioUrl`

php/src/Data/ConfigurationManager.php
pabzm: Implemented as `aioToken`

pabzm:
I camelized the property names, replaced `setMultiple()` by `startTransaction()` + `commitTransaction()`, fixed the behaviour bugs you mentioned and resolved the conflicts.

Turning the other env-based attributes into properties will have to wait until Monday.

pabzm:
I finished camel-casing all properties, converting all the ENV-based methods into properties, and rebased onto the latest of "main".

php/src/Data/ConfigurationManager.php
```diff
@@ -1080,7 +1081,7 @@ private function getPlaceholderValue(string $placeholder) : string {
             'NEXTCLOUD_STARTUP_APPS' => $this->GetNextcloudStartupApps(),
             'NEXTCLOUD_ADDITIONAL_APKS' => $this->GetNextcloudAdditionalApks(),
             'NEXTCLOUD_ADDITIONAL_PHP_EXTENSIONS' => $this->GetNextcloudAdditionalPhpExtensions(),
-            'INSTALL_LATEST_MAJOR' => $this->install_latest_major ? 'yes' : '',
+            'INSTALL_LATEST_MAJOR' => $this->install_latest_major,
```
szaimen: I think this is a change in behaviour. We need to keep this

szaimen:
@pabzm please also fix the failing CI. Then this is good to go afaics ☺️

php/src/Data/ConfigurationManager.php
```diff
@@ -9,47 +9,345 @@ class ConfigurationManager
...
+        set { $this->set('nextcloud_addtional_apks', $value); }
+    }
+
+    public string $nextcloudAdditionalPhpExtensions {
+        get => trim($this->GetEnvironmentalVariableOrConfig('NEXTCLOUD_ADDITIONAL_PHP_EXTENSIONS', 'nextcloud_additional_php_extensions', 'imagick'));
+        set { $this->set('nextcloud_additional_php_extensions', $value); }
+    }
+
+    public bool $collaboraSeccompDisabled {
+        get => booleanize($this->GetEnvironmentalVariableOrConfig('COLLABORA_SECCOMP_DISABLED', 'collabora_seccomp_disabled', ''));
```
szaimen: Shouldnt it be `$this->booleanize` ?
```suggestion
        get => $this->booleanize($this->GetEnvironmentalVariableOrConfig('COLLABORA_SECCOMP_DISABLED', 'collabora_seccomp_disabled', ''));
```

php/src/Data/ConfigurationManager.php
```diff
@@ -1083,55 +933,121 @@ public function listAvailableCommunityContainers() : array {
...
+            // Pad the result with nulls so psalm is happy (and we don't risk to run into warnings in case
+            // the check for an equal sign from above gets changed).
+            [$key, $value] = explode('=', $keyWithValue, 2) + [null, null];
+            $key = $this->camelize($key);
+            if ($value === null) {
+                error_log("Invalid input: '$keyWithValue' has no value after the equal sign");
+            } else if (!property_exists($confManager, $key)) {
+                error_log("Error: '$key' is not a valid configuration key (in '$keyWithValue')");
+            } else {
+                $confManager->$key = $value;
```
szaimen:
```suggestion
            $keyWithValue = $this->replaceEnvPlaceholders($variable);
            // Pad the result with nulls so psalm is happy (and we don't risk to run into warnings in case
            // the check for an equal sign from above gets changed).
            [$key, $value] = explode('=', $keyWithValue, 2) + [null, null];
            $key = $this->camelize($key);
            if ($value === null) {
                error_log("Invalid input: '$keyWithValue' has no value after the equal sign");
            } else if (!property_exists($this, $key)) {
                error_log("Error: '$key' is not a valid configuration key (in '$keyWithValue')");
            } else {
                $this->$key = $value;
```

php/src/Data/ConfigurationManager.php
pabzm: Yes, of course! 🙈

php/src/Data/ConfigurationManager.php
pabzm: I had understood that we want the actual major version in there, that's why I changed the assignment. But I would overlooked to also change the check in `Containers/nextcloud/entrypoint.sh:395`.

Now changed back to the original assignment.

php/src/Controller/ConfigurationController.php
szaimen, line 19: I think we should be able to start a transaction here

szaimen, line 135: Then we should also commit the transaction here

php/src/Docker/DockerActionManager.php
```diff
@@ -408,7 +400,7 @@ public function CreateContainer(Container $container): void {
             // // Special things for the nextcloud container which should not be exposed in the containers.json
             // } elseif ($container->identifier === 'nextcloud-aio-nextcloud') {
             //     foreach ($container->volumes->GetVolumes() as $volume) {
-            //         if ($volume->name !== $this->configurationManager->GetNextcloudMount()) {
+            //         if ($volume->name !== $this->configurationManager->nextcloud_mount) {
```
```suggestion
            //         if ($volume->name !== $this->configurationManager->nextcloudMount) {
```

szaimen:
Conflicts unfortunately

pabzm:
Rebased to resolve the conflicts, and added an addition to renaming `nextcloud_mount` to `nextcloudMount` in commented code.

php/src/Controller/ConfigurationController.php
pabzm: We can technically do that, but it would encapsulate quite a lot of code, and we would need to be really sure that no other process or code tries to write anything to the config while our "transaction" is "open", or we create ourselves race conditions.

I'd like to think about this for a day, and would suggest to merge this PR as it is, since it's big and old enough already, and to create a new PR for this idea.

szaimen: fine by me :)

szaimen: See https://github.com/nextcloud/all-in-one/pull/7392#discussion_r2736145005

php/get-configurable-aio-variables.sh
szaimen:
```suggestion
awk '/^    public [^f][^u][^n]/ { sub(/\$/, "", $3); print $3 }' src/Data/ConfigurationManager.php | sort
```

szaimen:
🎉🎉🎉🎉🎉

szaimen:

PR #7402, opened 2026-01-09
allow to configure redis cluster

szaimen:


Containers/nextcloud/config/redis.config.php
```diff
@@ -1,23 +1,48 @@
...
-if (getenv('REDIS_HOST')) {
-  $CONFIG = array(
-    'memcache.distributed' => '\OC\Memcache\Redis',
-    'memcache.locking' => '\OC\Memcache\Redis',
-    'redis' => array(
-      'host' => getenv('REDIS_HOST'),
-      'password' => (string) getenv('REDIS_HOST_PASSWORD'),
-    ),
-  );
+if (!getenv('REDIS_CLUSTER')) {
```
szaimen: alternatively check if a second host is provided and then set cluster automatically

szaimen: done

Containers/nextcloud/Dockerfile
```diff
@@ -138,8 +138,8 @@ RUN set -ex; \
     } > /usr/local/etc/php/conf.d/nextcloud.ini; \
     \
     { \
-        echo 'session.save_handler = redis'; \
-        echo 'session.save_path = "tcp://${REDIS_HOST}:${REDIS_PORT}?database=${REDIS_DB_INDEX}${REDIS_USER_AUTH}&auth[]=${REDIS_HOST_PASSWORD}"'; \
+        echo 'session.save_handler = rediscluster'; \
```
szaimen: this needs to be conditionally toggled. So probably actually need a variable for this

Containers/nextcloud/Dockerfile
```diff
@@ -138,8 +138,8 @@ RUN set -ex; \
...
+        echo 'session.save_path = "seed[]=${REDIS_HOST}:${REDIS_PORT}&seed[]=${REDIS_HOST_2}:${REDIS_PORT_2}&seed[]=${REDIS_HOST_3}:${REDIS_PORT_3}&seed[]=${REDIS_HOST_4}:${REDIS_PORT_4}&seed[]=${REDIS_HOST_5}:${REDIS_PORT_5}&seed[]=${REDIS_HOST_6}:${REDIS_PORT_6}&seed[]=${REDIS_HOST_7}:${REDIS_PORT_7}&seed[]=${REDIS_HOST_8}:${REDIS_PORT_8}&seed[]=${REDIS_HOST_9}:${REDIS_PORT_9}&timeout=0&read_timeout=0&failover=error&persistent=1&auth=${REDIS_HOST_PASSWORD}&stream[verify_peer]=0"'; \
```
szaimen: Not sure how we can handle this. Maybe one seed is enough?

Containers/nextcloud/config/redis.config.php
```diff
@@ -1,5 +1,5 @@
 <?php
-if (getenv('REDIS_HOST')) {
+if (!getenv('REDIS_HOST_2')) {
```
szaimen: if we have a vaiable for cluster we can come back to using that

Containers/nextcloud/config/redis.config.php
```diff
@@ -17,7 +17,33 @@
...
+      'seeds' => array(
+        getenv('REDIS_HOST') . ':' . (string)getenv('REDIS_PORT'),
+        getenv('REDIS_HOST_2') . ':' . (string)getenv('REDIS_PORT_2'),
+        getenv('REDIS_HOST_3') . ':' . (string)getenv('REDIS_PORT_3'),
+        getenv('REDIS_HOST_4') . ':' . (string)getenv('REDIS_PORT_4'),
+        getenv('REDIS_HOST_5') . ':' . (string)getenv('REDIS_PORT_5'),
+        getenv('REDIS_HOST_6') . ':' . (string)getenv('REDIS_PORT_6'),
+        getenv('REDIS_HOST_7') . ':' . (string)getenv('REDIS_PORT_7'),
+        getenv('REDIS_HOST_8') . ':' . (string)getenv('REDIS_PORT_8'),
+        getenv('REDIS_HOST_9') . ':' . (string)getenv('REDIS_PORT_9'),
```
szaimen: not sure if this is going to work. What happens if not all variables are set?

szaimen:
Another question are whiteboard and notify-push integrations...

Containers/nextcloud/Dockerfile
szaimen: We need to assamble the stribg from outside

szaimen:
> Another question are whiteboard and notify-push integrations...

Whiteboard doesnt sipport it, notify push does

Containers/notify-push/start.sh
```diff
@@ -83,6 +83,7 @@ elif [ -f "/nextcloud/data/certificates/MYSQL" ]; then
 fi
 
 # Set sensitive values as env
+# todo
```
szaimen: Todo

Containers/nextcloud/Dockerfile
szaimen: Redis_mode

Containers/nextcloud/config/redis.config.php
szaimen: Maybe
```suggestion
if (getenv('REDIS_HOST_9') && getenv('REDIS_PORT_9')) {
        getenv('REDIS_HOST_9') . ':' . (string)getenv('REDIS_PORT_9'),
}
```

szaimen: found a solution

Containers/nextcloud/config/redis.config.php
szaimen: done

Containers/notify-push/start.sh
```diff
@@ -44,52 +38,9 @@ fi
...
-    export POSTGRES_USER
-fi
-
-# URL-encode passwords
-POSTGRES_PASSWORD="$(jq -rn --arg v "$POSTGRES_PASSWORD" '$v|@uri')"
-REDIS_HOST_PASSWORD="$(jq -rn --arg v "$REDIS_HOST_PASSWORD" '$v|@uri')"
-
-# Postgres root cert
-if [ -f "/nextcloud/data/certificates/POSTGRES" ]; then
-    CERT_OPTIONS="?sslmode=verify-ca&sslrootcert=/nextcloud/data/certificates/ca-bundle.crt"
```
szaimen, line 79: This is probably going to break notify push. It only might work if we mount thevolume in the same dir like in the nextcloud container

szaimen:
Failing playwright test is expected at this point as we did not re-built the images yet.

Containers/notify-push/start.sh
```diff
@@ -44,52 +38,9 @@ fi
...
-export DATABASE_URL="$DATABASE_TYPE://$POSTGRES_USER:$POSTGRES_PASSWORD@$POSTGRES_HOST:$POSTGRES_PORT/$POSTGRES_DB$CERT_OPTIONS"
-export REDIS_URL="redis://$REDIS_USER:$REDIS_HOST_PASSWORD@$REDIS_HOST:$REDIS_PORT/$REDIS_DB_INDEX"
-
 # Run it
-/nextcloud/custom_apps/notify_push/bin/"$CPU_ARCH"/notify_push \
-    --database-prefix="oc_" \
-    --nextcloud-url "https://$NC_DOMAIN" \
-    --port 7867
+/var/www/html/custom_apps/notify_push/bin/"$CPU_ARCH"/notify_push \
+    --port 7867 \
```
pabzm, line 43: Port 7867 is the default, this line could be removed.

php/containers.json
```diff
@@ -298,22 +298,13 @@
       "volumes": [
         {
           "source": "nextcloud_aio_nextcloud",
-          "destination": "/nextcloud",
+          "destination": "/var/www/html",
```
pabzm, line 301: Why do you change the volume mount point in this PR?

Containers/notify-push/start.sh
szaimen: Because we do not version the binary (it is loaded from app-store) I would like to make sure that it stays like that this way :)

php/containers.json
szaimen: We also load any config for certificates here, so we probably need to make sure that it is mounted at the same place so that the path is the same :)

szaimen:

PR #7410, opened 2026-01-12
s3.config.php: allow to configure num_buckets

szaimen:


Containers/nextcloud/config/s3.config.php
```diff
@@ -10,6 +10,7 @@
       'class' => '\OC\Files\ObjectStore\S3',
       'arguments' => array(
         'multibucket' => $multibucket === 'true',
+        'num_buckets' => (int)getenv('OBJECTSTORE_S3_NUM_BUCKETS') ?: 64,
```
provokateurin, line 13:
```suggestion
        'num_buckets' => (int)(getenv('OBJECTSTORE_S3_NUM_BUCKETS') ?: 64),
```
Did you test what happens if you put in no value? I suspect null would be casted to an integer (which throws an error), before the elvis operator comes into play.

szaimen: > Did you test what happens if you put in no value?

I tested it:

image

Issue #7308, opened 2025-12-13
Restore with BorgBackup fail to set permissions


BlackDuck888:
### Steps to reproduce
1. Install an environment from scratch.
2. Take a backup
3. Remove the entire environment with volume 
4. Start from scratch but restore from the backup instead of performing a new setup.

### Expected behavior 
I expect that all files, when restored from the backup, will have the same permissions as they had when they were saved in the backup.

### Actual behavior 
The system can be restored and is functioning, but the UI indicates that the last restoration failed. There are numerous entries in the log regarding unset file permissions.

#### Screenshot
<img width="699" height="778" alt="Image" src="https://github.com/user-attachments/assets/9852e08a-6866-4625-bcad-7e1f28fc2d73" />

#### Log extract
```
2025-12-11T13:20:00.949627669Z Restoring '20251211_115554-nextcloud-aio'...
2025-12-11T13:20:03.316526356Z sending incremental file list
2025-12-11T13:20:03.369280130Z [sender] expand file_list pointer array to 1024 bytes, did move
2025-12-11T13:20:03.411225744Z [sender] expand file_list pointer array to 4096 bytes, did move
2025-12-11T13:20:03.499045942Z [sender] expand file_list pointer array to 1024 bytes, did move
2025-12-11T13:20:03.532722999Z [sender] expand file_list pointer array to 4096 bytes, did move
2025-12-11T13:20:03.837094135Z [sender] expand file_list pointer array to 16384 bytes, did move
2025-12-11T13:20:04.343513711Z delta-transmission disabled for local transfer or --whole-file
2025-12-11T13:20:04.343544970Z ./
2025-12-11T13:20:04.345614561Z rsync: [generator] failed to set permissions on "/nextcloud_aio_volumes/nextcloud_aio_apache/caddy": Bad address (14)
2025-12-11T13:20:04.345650639Z rsync: [generator] failed to set permissions on "/nextcloud_aio_volumes/nextcloud_aio_database": Bad address (14)
2025-12-11T13:20:04.350890936Z nextcloud_aio_apache/
2025-12-11T13:20:04.350918257Z nextcloud_aio_apache/caddy/
2025-12-11T13:20:04.350924830Z nextcloud_aio_database/
2025-12-11T13:20:04.351228910Z nextcloud_aio_database/PG_VERSION
2025-12-11T13:20:04.370109919Z nextcloud_aio_database/pg_hba.conf
...
[...]

PR #7414, opened 2026-01-13
DockerActionManager: disable seccomp policy for borgbackup container

szaimen:
- Fix https://github.com/nextcloud/all-in-one/issues/7308

### DIFF
--- a/php/src/Docker/DockerActionManager.php
+++ b/php/src/Docker/DockerActionManager.php
@@ -376,6 +376,11 @@ readonly class DockerActionManager {
 
         // Special things for the backup container which should not be exposed in the containers.json
         if (str_starts_with($container->GetIdentifier(), 'nextcloud-aio-borgbackup')) {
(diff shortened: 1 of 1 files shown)


BlackDuck888:
My tests showed that, as I wrote, `seccomp:unconfined` has no effect. The only thing that works for me is setting my own Seccomp rule or using Alpine version 3.18. It would help me and all QNAP users if we could specify the Alpine version in the dock container, for example.

szaimen:
> My tests showed that, as I wrote, `seccomp:unconfined` has no effect. The only thing that works for me is setting my own Seccomp rule or using Alpine version 3.18. It would help me and all QNAP users if we could specify the Alpine version in the dock container, for example.

Ah wait so you disabled the seccomp policy for the test container and it did not make it rsync work?

BlackDuck888:
That is correct!

szaimen:
> That is correct!

Ok, then I misunderstood your comment in thr issue. Need to follow-up on this then.

szaimen:
Follow-up here: https://github.com/nextcloud/all-in-one/pull/7426

PR #7419, opened 2026-01-14
Feature #7351. Adding APACHE_LABEL to apply custom labels

omnizach:
This applies custom labels to the Apache and Domaincheck containers based on the APACHE_LABEL environment label when composing the master container. This allows for configuration of label-based reverse proxies such as TSDProxy and Traefik, making reverse proxy configuration much easier and provide more options.


omnizach:
I missed that the environment variables need to be passed through container.json.

I see what you mean about not doing this in the DockerFile, but in the php. I'm working on moving the code there. However, I see in your example the `Labels` entry is a dictionary. According to the Docker [Label](https://docs.docker.com/reference/dockerfile/#label) command, you can specify multiple labels with spaces delimited, so it could just be passing the value provided through as a string and allow Docker to handle the parsing. This seems safer than doing the parsing in php since the rules are more complicated than simply split-on-space.

After adding the environment variables to container.json, will the direct DockerFile solution not work at all? Would adding these labels as a simple string in php work (as it will eventually add this to the generated DockerFile in the same syntax)?

Apologies for the rudimentary questions here. I looked for documentation for setting this up in a development environment, but I have to admit that while I can do this for a simple DockerFile, this is a bit beyond my skills at the moment. If I can get a dev environment going, I can answer these questions for myself and get it right.

szaimen:
Hi,

> I looked for documentation for setting this up in a development environment, but I have to admit that while I can do this for a simple DockerFile, this is a bit beyond my skills at the moment. If I can get a dev environment going, I can answer these questions for myself and get it right.

See https://github.com/nextcloud/all-in-one/blob/main/develop.md#how-to-locally-build-and-test-changes-to-mastercontainer for how you can spin up a dev instance

osmal-dev:
for traefik there already a solution that can be used until this feature is ready, here a compose example:
```yaml
services:
    ...
    environment:
      ...
      - APACHE_ADDITIONAL_NETWORK=proxy
    labels:
      ...
      - "traefik.docker.network=proxy"
      - "traefik.http.services.nextcloud.loadbalancer.server.url=http://nextcloud-aio-apache:PORT"
```

Issue #7198, opened 2025-11-25
Easier office switcher and UI polishing


juliusknorr:
### Is your feature request related to a problem? Please describe. 

We want to improve visibility of different office solutions and make switching easier to admins. During planning meetings we discussed ideas and the first logical step seems to be to improve the AIO interface. 

### Describe the solution you'd like

The idea would be to have a feature matrix like comparison advertising the different options and highlighting Nextcloud Office.

@marcoambrosini can probably come up with a design proposal.

Also some general design review of the UI might be nice to polish a bit (e.g. grouping of settings)

### Describe alternatives you've considered

Do this in Nextcloud but there we lack automatic setup and configuration that AIO can already provide, we may still implement this but as a fallback for non-AIO instances.

### Additional context

Current UI screenshot:

<img width="351" height="635" alt="Image" src="https://github.com/user-attachments/assets/74b69195-fa72-4f6c-a7ca-21d3fc80e2b9" />

marcoambrosini:
Current state

https://github.com/user-attachments/assets/c4171fc4-5e6e-411b-b074-d9b9bbcc7ada

PR #7421, opened 2026-01-14
feat: Add office switcher with feature comparison

juliusknorr:
Implements the office switcher from #7198 

Screenshot 2026-01-14 at 22 21 40

I'm unable to test with the actual containers, only managed to get the master container itself running locally. @szaimen Can you help with that?

[Patch for myself to remember how to develop locally](https://gist.github.com/juliusknorr/4c3d0cd38476dcce45d095f7701c87a5)

### DIFF
--- a/php/public/style.css
+++ b/php/public/style.css
@@ -576,6 +576,7 @@ input[type="checkbox"]:disabled:not(:checked) + label {
 
 .office-card-disabled {
     opacity: 50%;
+    pointer-events: none;
 }
 
 .office-card:hover {
--- a/php/templates/includes/optional-containers.twig
+++ b/php/templates/includes/optional-containers.twig
@@ -40,12 +40,14 @@
                 <li>Best ODF compatibility</li>
                 <li>Best support for legacy files</li>
             </ul>
-            <a href="https://www.collaboraoffice.com/code/" target="_blank" class="office-learn-more" onclick="event.stopPropagation();">
-                Learn more
-                <svg width="16" height="16" viewBox="0 0 16 16" fill="none" xmlns="http://www.w3.org/2000/svg" style="vertical-align: middle; margin-left: 4px;">
-                    <path d="M6 12L10 8L6 4" stroke="currentColor" stroke-width="2" stroke-linecap="round" stroke-linejoin="round"/>
-                </svg>
-            </a>
+            {% if isAnyRunning == false %}
+                <a href="https://www.collaboraoffice.com/code/" target="_blank" class="office-learn-more" onclick="event.stopPropagation();">
+                    Learn more
+                    <svg width="16" height="16" viewBox="0 0 16 16" fill="none" xmlns="http://www.w3.org/2000/svg" style="vertical-align: middle; margin-left: 4px;">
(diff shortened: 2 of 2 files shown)


php/templates/containers.twig
```diff
@@ -32,7 +32,7 @@
             {% set isAnyRunning = false %}
             {% set isAnyRestarting = false %}
             {% set isWatchtowerRunning = false %}
-            {% set isDomaincheckRunning = false %}
+            {% set isDomaincheckRunning = true %}
```
juliusknorr: Sorry that was unintended and just for easier testing, will remove

marcoambrosini (commented):
Very nice @juliusknorr. I actually like the styling of the cards here better. would it be possible to apply it also to https://github.com/nextcloud/server/pull/57555 ?

szaimen:
Hey @juliusknorr , thanks a lot for the PR! ☺️

> I'm unable to test with the actual containers, only managed to get the master container itself running locally. @szaimen Can you help with that?

Usually I would recommend to follow https://github.com/nextcloud/all-in-one/blob/main/develop.md#how-to-locally-build-and-test-changes-to-mastercontainer. You might be able to also combine this with https://github.com/nextcloud/all-in-one/discussions/6817.

szaimen:
BTW @juliusknorr can you please fix the Playwright test?

szaimen:
> * In general it would be nice to widen the main container that wraps all the AIO interface a bit. @szaimen what do you think?

yeah would be possible to widen it a bit. In general we limited the width for best readability

marcoambrosini:
Just tested with the `--max-width` variable set to 580px and it looks pretty good. We're still below 80 characters per line and I think it looks a bit better. These cards would also benefit from it.

php/public/style.css
```diff
@@ -28,7 +28,7 @@
     --border-radius-large: 12px;
     --default-font-size: 13px;
     --checkbox-size: 16px;
-    --max-width: 500px;
+    --max-width: 800px;
```
marcoambrosini:
```suggestion
    --max-width: 580px;
```

marcoambrosini: (since we have big paragraphs with explanations I would still keep this container on the narrow side and around/below 80 chars per line @juliusknorr )

juliusknorr: Adjusted 👍

marcoambrosini:
@juliusknorr in the last screenshot from @szaimen, the disable button that should be underneath the cards is not visible anymore, was it decided to remove that?

szaimen:
> @juliusknorr in the last screenshot from @szaimen, the disable button that should be underneath the cards is not visible anymore, was it decided to remove that?

I removed it in the case the containers are running because it does not make sense to show it while you cannot change your office selection. It gets shown again after you stop the containers

ghost:
Screenshot from 2026-01-27 12-03-25

just a simple 50% opacity :)

szaimen (approved):
LGTM now :)

szaimen:
Thanks everyone ☺️

szaimen:

PR #7436, opened 2026-01-16
move version to a dedicated file

szaimen:


szaimen:
> Looks good and works. 👍 (Just out of curiosity: what's the motivation?)

Easier to parse, e.g. from other scripts and also linking to it :)

szaimen:

PR #7437, opened 2026-01-16
insert the AIO version into Nextcloud's system config

szaimen:
- Needs https://github.com/nextcloud/all-in-one/pull/7436


php/src/Data/ConfigurationManager.php
```diff
@@ -122,6 +122,14 @@ public function GetBackupTimes() : array {
         return $backupTimes;
     }
 
+    public function GetAioVersion() : string {
```
pabzm: We should start function names with a lower case character. The PHP world is divided if function/method names should be written with underscores or in camelCase, but all agree to a starting lower case character. Starting upper case is usually reserved for class names and constants.

php/src/Data/DataConst.php
```diff
@@ -66,4 +66,8 @@ public static function GetCollaboraSeccompProfilePath() : string {
     public static function GetContainersDefinitionPath() : string {
         return (string)realpath(__DIR__ . '/../../containers.json');
     }
+
+    public static function GetAioVersionFile() : string {
```
pabzm, line 70: This function name should also start with a lower case letter.

php/src/Data/ConfigurationManager.php
```diff
@@ -122,6 +122,14 @@ public function GetBackupTimes() : array {
...
+        $path = DataConst::GetAioVersionFile();
+        if ($path !== '' && file_exists($path)) {
+            return trim((string)file_get_contents($path));
```
pabzm, line 426: `file_get_contents()` returns `false` in case of an error. While we do check the file path before there always might be a disk read error or other problems, which could result in a `false`. Then this code would return an empty string. That is also what we're returning as a default, but I would suggest to add some error handling that logs a message so people can get a clue on what's wrong if they see `NC_AIO_VERSION=""` in their config.

php/src/Data/ConfigurationManager.php
```diff
@@ -122,6 +122,14 @@ public function GetBackupTimes() : array {
         return $backupTimes;
     }
 
+    public function GetAioVersion() : string {
+        $path = DataConst::GetAioVersionFile();
+        if ($path !== '' && file_exists($path)) {
```
pabzm, line 425: `realpath()` already checks if the file doesn't exist (and returns `false`, which we cast into `""`), so there's no reason to repeat that check.

php/src/Data/ConfigurationManager.php
asavageiv: I agree with what you're saying, but consistency within a codebase is more important than with the broader world. Therefore, if we're going to change the style of the whole repository it should be done in one separate commit. Furthermore, the commit should be in the `.git-ignore-blame-revs` file so that it doesn't get in the way of git blame for figuring out when a function definition was last changed.

szaimen: done

php/src/Data/ConfigurationManager.php
szaimen: I'd say it is ok if this returns an empty string without logging in case of an i/o error as this is only used inside the system report of your nextcloud instance, so nothing too critical if it is empty

php/src/Data/ConfigurationManager.php
szaimen: I think I'd like to keep this :)

php/src/Data/DataConst.php
szaimen: Lets rename all the functions in this file in a follow-up :)

pabzm (approved):
LGTM!

PR #7440, opened 2026-01-16
update NPMplus images in reverse proxy guide

Zoey2936:
NPM also needs an update of its screenshots, will do this when I have time. I would also like to update the nginx (and forks) example

### DIFF
--- a/reverse-proxy.md
+++ b/reverse-proxy.md
@@ -564,19 +564,14 @@ Note: this will cause that a non root user can bind privileged ports.
 
 Second, see these screenshots for a working config:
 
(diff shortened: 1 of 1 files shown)


szaimen:
> NPM also needs an update of its screenshots, will do this when I have time. I would also like to update the nginx (and forks) example

Do you want to do this in this PR or in a separate one?

Zoey2936:
> > NPM also needs an update of its screenshots, will do this when I have time. I would also like to update the nginx (and forks) example
> 
> Do you want to do this in this PR or in a separate one?

seperate

PR #7458, opened 2026-01-19
aio-interface: show sub-steps for starting containers

szaimen:
- Close https://github.com/nextcloud/all-in-one/discussions/6877

### DIFF
--- a/php/src/Controller/DockerController.php
+++ b/php/src/Controller/DockerController.php
@@ -335,7 +335,7 @@ readonly class DockerController {
         <!DOCTYPE html>
         <html lang="en" class="overlay-iframe">
             <head>
-                <link rel="stylesheet" href="../../style.css?v7" media="all" />
+                <link rel="stylesheet" href="../../style.css?v8" media="all" />
                 <script>
                     const observer = new MutationObserver((records) => {
                         const node = records[0]?.addedNodes[0];
--- a/php/templates/layout.twig
+++ b/php/templates/layout.twig
@@ -1,10 +1,10 @@
 <html>
     <head>
         <title>AIO</title>
-        <link rel="stylesheet" href="style.css?v7" media="all" />
+        <link rel="stylesheet" href="style.css?v8" media="all" />
         <link rel="icon" href="img/favicon.png">
-        <script type="text/javascript" src="forms.js"></script>
-        <script type="text/javascript" src="toggle-dark-mode.js"></script>
+        <script type="text/javascript" src="forms.js?v1"></script>
+        <script type="text/javascript" src="toggle-dark-mode.js?v1"></script>
     </head>
 
     <body>
--- a/php/templates/containers.twig
+++ b/php/templates/containers.twig
@@ -27,7 +27,7 @@
             <script type="text/javascript" src="timezone.js"></script>
 
             {# js for optional containers and additional containers forms #}
-            <script type="text/javascript" src="containers-form-submit.js?v5"></script>
+            <script type="text/javascript" src="containers-form-submit.js?v6"></script>
 
             {% set hasBackupLocation = borg_backup_host_location or borg_remote_repo %}
             {% set isAnyRunning = false %}


php/public/overlay-log.js
```diff
@@ -0,0 +1,27 @@
...
+                try {
+                    let parsedPayload = JSON.parse(serverSentEvent.data);
+                    displayOverlayLogMessage(parsedPayload.name || serverSentEvent.data);
+                } catch (parseError) {
+                    displayOverlayLogMessage(serverSentEvent.data);
+                }
+            });
+            serverSentEventSource.onerror = function() { serverSentEventSource.close(); };
+        } catch (connectionError) {
+            /* ignore if Server-Sent Events are not available */
```
asavageiv: Should we log something in case there is another unexpected source of an exception?

php/src/Controller/DockerController.php
```diff
@@ -315,4 +367,32 @@ private function StopDomaincheckContainer() : void
         $id = 'nextcloud-aio-domaincheck';
         $this->PerformRecursiveContainerStop($id);
     }
+
+    // Write container event to events file and prune old events
+    private function writeEventsToFile(array $payload): void {
+        $eventJson = json_encode($payload);
+
+        // Append new event (atomic via LOCK_EX)
+        file_put_contents($eventsFile, $eventJson . PHP_EOL, FILE_APPEND | LOCK_EX);
```
asavageiv: Does this work? Is `$eventsFile` in scope?

php/src/Controller/DockerController.php
```diff
@@ -34,6 +35,15 @@ private function PerformRecursiveContainerStart(string $id, bool $pullImage = tr
             return;
         }
 
+        // Emit a container-start event for frontend clients (one JSON line per event)
```
asavageiv: I think DockerController is starting to have too many direct responsibilities. It already mixes the logic for container lifecycle with HTTP request/response handling and this adds ContainerEvents too. How about we create a ContainerEventLog class and encapsulate the file management there? I imagine an API like `emitEvent` and `open`, `readNextEvent`, `close` would be easy from where this is now.

php/src/Controller/DockerController.php
```diff
@@ -34,6 +35,15 @@ private function PerformRecursiveContainerStart(string $id, bool $pullImage = tr
...
+        try {
+            $this->pruneEventsFileIfTooLarge();
+            $this->writeEventsToFile(['event' => 'Starting container', 'name' => $id, 'time' => time()]);
+        } catch (\Throwable $e) {
+            // non-fatal, just log
+            error_log('Could not write container-start event: ' . $e->getMessage());
+        }
+
         $this->dockerActionManager->DeleteContainer($container);
```
asavageiv, line 39: This is out of scope, but the fact that there are so many calls in a row to `dockerActionManager` here make it smell like this function should be moved into the DockerActionManager. WDYT? Would you accept a PR?

Generally, I would slowly move towards DockerController focusing on handling HTTP requests and responses and calling into deeper layers for the logic of the requests. The handling would start to all look like:
1. read the request
2. call a method or two to make it happen
3. write back the response with the results

php/src/Controller/DockerController.php
```diff
@@ -269,6 +279,48 @@ public function StopContainer(Request $request, Response $response, array $args)
         return $response->withStatus(201)->withHeader('Location', '.');
     }
 
+    public function StreamContainerEvents(Response $response): Response {
+        $eventsFile = \AIO\Data\DataConst::GetContainerEventsFile();
```
asavageiv: Can this be moved next to the `fopen` or does it have to happen before setting the headers?

php/public/overlay-log.js
```diff
@@ -0,0 +1,27 @@
...
+        if (!overlayLogElement) {
+            return;
+        }
+        overlayLogElement.textContent = message;
+    }
+
+    // Attempt to connect to Server-Sent Events at /events/containers and listen for 'container-start' events
+    if (typeof EventSource !== 'undefined') {
+        try {
+            const serverSentEventSource = new EventSource('events/containers');
```
asavageiv: https://developer.mozilla.org/en-US/docs/Web/API/EventSource says that this has significant limitations when used over HTTP1. Are we concerned about that at all?

php/src/Controller/DockerController.php
```diff
@@ -269,6 +279,48 @@ public function StopContainer(Request $request, Response $response, array $args)
...
+        $fileHandle = fopen($eventsFile, 'r');
+        if ($fileHandle === false) {
+            $body->write('');
+            return $response;
+        }
+
+        // Start at end of file so only new events are streamed
+        fseek($fileHandle, 0, SEEK_END);
+
+        while (!connection_aborted()) {
```
asavageiv: In the good case where all the containers start, will the connection abort?

php/src/Controller/DockerController.php
```diff
@@ -269,6 +279,48 @@ public function StopContainer(Request $request, Response $response, array $args)
...
+            clearstatcache(false, $eventsFile);
+            $line = fgets($fileHandle);
+            if ($line !== false) {
+                $data = trim($line);
+                // Write SSE event
+                $body->write("event: container-start\n");
```
asavageiv: Creating a `ContainerEvent` class would also abstract the fact that right now these are only 'start' events.

php/src/Controller/DockerController.php
```diff
@@ -269,6 +279,48 @@ public function StopContainer(Request $request, Response $response, array $args)
...
+        }
+
+        $body = $response->getBody();
+        $response = $response
+            ->withHeader('Content-Type', 'text/event-stream')
+            ->withHeader('Cache-Control', 'no-cache')
+            ->withHeader('Connection', 'keep-alive');
+
+        $fileHandle = fopen($eventsFile, 'r');
+        if ($fileHandle === false) {
```
asavageiv: We should probably log an error if this happens since we try and create the file above. Something's configured wrong if this failed.

php/public/overlay-log.js
szaimen: We disallow multiple tabs so that should not be a problem but we should probably add a check here as well to abort any connection attempt if a second tab was opened like done here: https://github.com/nextcloud/all-in-one/blob/5752556e4ff2ad9025b73844a8c28c400cd00b82/php/public/second-tab-warning.js#L6

pabzm:
I implemented an alternative approach that polls for changes instead of relying on server-sent events, because in my experience server-sent events aren't really stable to use if the network isn't. With polling it doesn't matter if the connection drops once in a while, or some requests fail.

The implementation makes the `Container` class log events, which get stored in a file line by line. Javascript code repeatedly `GET`s an API endpoint that just deliverers that file unfiltered (it's only truncated if it gets too large, as in the SSE implementation. If the file didn't change since the last request a 304 response without content is served (via an Etag based on the `mtime` of the file)

I didn't remove any code, because I'm not finally sure which way we should chose. The polling implementation has the problem that sometimes its requests stay `pending` e.g. when at first setup the container images are pulled and the big spinner overlays the page: Sometimes the logged events are shown below the spinner, but sometimes they aren't (because they didn't receive a response yet). I currently don't know what this behaviour is due to. If anyone would have a hint, I'd be all ears.
I don't know if the SSE-implementation would suffer from the same problem.

The concept to log the events from the `Container` class (via a standalone model class) could be used in the SSE-implementation, too.

Both solutions conflict with the icons and the container state as rendered from `containers.twig` (once the containeres have been started for the first time). Those details rely on the automatic reload, which more or less breaks polling as well as SSE. Also they kind of show similar information.

pabzm:
Out application allows only one request per client at once due to the file-backed sessions. Those block as long as they are opened for writing.

I tried to introduce calls to `session_write_close()` in several places but didn't succeed to fix the issue. Only disabling sessions (and thus authentication) allowed me to get responses to multiple requests in parallel.

Since I'm a little wary about changing this situation (there could be wanted side effects of this behaviour) I thought about a third option how to get detailed status info to the client while the backend still pulls images, and I came up with just streaming HTML as response to the request and have it loaded into an iframe.
It took me a while to figure out how to disable output buffering on all layers (Apache, PHP, Slim framework), but now it works. One additional dependency would be required, but even less custom code than for the other implementations. 

Here's a screencast:


https://github.com/user-attachments/assets/e1281a53-f511-45b2-a1a9-0f548ecfec66



I added a little bit of styling, but that could of course be polished if we decide to go forward with it.

What do you think about it?

szaimen:
> Out application allows only one request per client at once due to the file-backed sessions. Those block as long as they are opened for writing.
> 
> I tried to introduce calls to `session_write_close()` in several places but didn't succeed to fix the issue. Only disabling sessions (and thus authentication) allowed me to get responses to multiple requests in parallel.
> 
> Since I'm a little wary about changing this situation (there could be wanted side effects of this behaviour) I thought about a third option how to get detailed status info to the client while the backend still pulls images, and I came up with just streaming HTML as response to the request and have it loaded into an iframe. It took me a while to figure out how to disable output buffering on all layers (Apache, PHP, Slim framework), but now it works. One additional dependency would be required, but even less custom code than for the other implementations.
> 
> Here's a screencast:
> 
>  Screen.Recording.2026-02-11.at.16.12.38.mov 
> I added a little bit of styling, but that could of course be polished if we decide to go forward with it.
> 
> What do you think about it?

I like it! Lets go forward with that solution. Thanks a lot! 😊

pabzm:
I remove the previous attempts from this branch (they're still available in branch `enh/6877/show-sub-steps_multi-solution`, in case we might want to re-visit them) and rebased it onto the latest of `main`.

I added `font-family: sans-serif` to the minimal styles that are sent to the iframe, but I'm hesitating to add more styling or UI fanciness (a tiny spinner on the last line?) in order to keep the thing simple.

@szaimen Are there any styling or UX improvements you'd like to see?

pabzm:
@asavageiv Since the code changed a lot from the initial approach I resolved all your comments' conversations. I'd be interested to hear about your thoughts on the new approach and code, if you can find the time!

Containers/mastercontainer/mastercontainer.conf
```diff
@@ -21,6 +21,11 @@ Listen 8080 https
     <FilesMatch "\.php$">
         SetHandler "proxy:fcgi://127.0.0.1:9000"
     </FilesMatch>
+
+    # Disable output buffering to enable streaming responses.
+    <Proxy "fcgi://127.0.0.1:9000/" enablereuse=on flushpackets=on max=10>
```
szaimen: We have this also already enabled for the apache container. Maybe we should standardize both configs? See https://github.com/nextcloud/all-in-one/blob/974f44345564fb2fc3a11d6987a8298f76a7f6e7/Containers/apache/nextcloud.conf#L17

php/public/containers-form-submit.js
```diff
@@ -1,4 +1,9 @@
 document.addEventListener("DOMContentLoaded", function () {
+    // Don't run if the expected form isn't present.
+    if (document.getElementById('options-form') === null) {
+        return;
```
szaimen, line 4: Do we need this extra check here? What does it fix?

php/src/Controller/DockerController.php
```diff
@@ -36,7 +37,13 @@ private function PerformRecursiveContainerStart(string $id, bool $pullImage = tr
 
         $this->dockerActionManager->DeleteContainer($container);
         $this->dockerActionManager->CreateVolumes($container);
+        if ($body) {
+            $body->write("<div>{$container->displayName}: Pulling image</div>");
+        }
```
szaimen: I guess we should insert the body into the `PullImage` method as 3rd argument as we do not always pull an image. Thus we can re-use the logic in there and only write the body if we actually pull the image

php/src/Controller/DockerController.php
```diff
@@ -36,7 +37,13 @@ private function PerformRecursiveContainerStart(string $id, bool $pullImage = tr
...
         $this->dockerActionManager->PullImage($container, $pullImage);
+        if ($body) {
+            $body->write("<div>{$container->displayName}: Starting container</div>");
+        }
```
szaimen: If we do the above, it makes sense to also insert it into the createcontainer function to be consistent

php/src/Controller/DockerController.php
```diff
@@ -307,4 +326,26 @@ private function StopDomaincheckContainer() : void
...
         $this->PerformRecursiveContainerStop($id);
     }
+
+    private function getStreamingResponseHtmlStart() : string {
+        return <<<END
+        <!DOCTYPE html>
+        <html lang='en'>
+            <head>
+                <style>
+                    body { font-family: sans-serif; color: white; }
```
szaimen: Lets maybe re-use the exising font-family to be consistent? https://github.com/nextcloud/all-in-one/blob/974f44345564fb2fc3a11d6987a8298f76a7f6e7/php/public/style.css#L72
```suggestion
                    body { font-family: system-ui, -apple-system, 'Segoe UI', Roboto, Oxygen-Sans, Cantarell, Ubuntu, 'Helvetica Neue', 'Noto Sans', 'Liberation Sans', Arial, sans-serif, 'Apple Color Emoji', 'Segoe UI Emoji', 'Segoe UI Symbol', 'Noto Color Emoji'; color: white; }
```

szaimen: Also, have you tested the log in dark mode and also in white mode?

php/templates/layout.twig
```diff
@@ -13,6 +13,7 @@
         </div>
         <div id="overlay">
             <div class="loader"></div>
+            <iframe name='overlay-log' id="overlay-log"></iframe>
```
szaimen:
```suggestion
            <iframe name="overlay-log" id="overlay-log"></iframe>
```

szaimen:
Please also fix psalm :)

php/src/Controller/DockerController.php
pabzm: Do you mean to log from inside the `StartContainer` method? Or to add a statement "Create Container", too?

php/public/containers-form-submit.js
pabzm: The script is also loaded from the domain setup view, and currently throws an exception there, which is fixed by this.
I didn't feel comfortable to decide how to only load the script when it's required, so I chose this way.

php/src/Controller/DockerController.php
szaimen: > Do you mean to log from inside the `StartContainer` method?

Yes :)

php/public/containers-form-submit.js
szaimen: Generally I'd say we should not load the file on templates where we don't need it

szaimen:
Thanks @pabzm! One thing that is still missing is that we need to increase the version number of all js and css files that were modified in the templates where it is loaded so that we can get the latest version automatically...

php/public/containers-form-submit.js
pabzm: I agree, but I was sure where to put it in this huge template. I'm very open to fix this in a better way, but I'd leave this for another PR.

php/src/Controller/DockerController.php
pabzm: I refactored the code flow a little bit, so now `DockerActionManager` adds the statements.

php/public/containers-form-submit.js
szaimen: Ok

pabzm:
I changed the styling in so far as now the iframe also just loads the styles.css (which is in cache already anyway).

It now looks like this

Screenshot 2026-02-17 at 10 40 22

-----------

Screenshot 2026-02-17 at 10 40 34

pabzm:
I condensed the commit history and rebased onto the latest of "main", now it's ready for merging, I think.

szaimen:

szaimen:
🎉🎉🎉🎉🎉

Issue #7466, opened 2026-01-21
Use StopSignal for stopping containers


szaimen:
Change to eg `"StopTimeout": 10,` during container creation

https://docs.docker.com/reference/api/engine/version/v1.44/#tag/Container/operation/ContainerCreate

PR #7468, opened 2026-01-21
`CreateContainer`: also insert the max shutdown time into the container itself

szaimen:
- Close https://github.com/nextcloud/all-in-one/issues/7466

### DIFF
--- a/php/src/Docker/DockerActionManager.php
+++ b/php/src/Docker/DockerActionManager.php
@@ -356,6 +356,11 @@ readonly class DockerActionManager {
 
         $requestBody['HostConfig']['Init'] = $container->init;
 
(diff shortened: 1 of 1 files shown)


pabzm:
The code looks good to me, and it's probably sensible to integrate that option. But it's not used yet by any of our containers, is it? Is there a concrete use for it?

szaimen:
> The code looks good to me, and it's probably sensible to integrate that option. But it's not used yet by any of our containers, is it? Is there a concrete use for it?

It is used for example by the database and nextcloud containers

szaimen:
CI failure is unrelated

szaimen:

PR #7476, opened 2026-01-22
Add Pull request Template

docjyJ:
See: https://github.com/nextcloud/server/blob/master/.github/pull_request_template.md

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -5,21 +5,4 @@
 -->
 
 * Resolves: # <!-- related github issue -->
-
-## Summary
-
-
-## TODO
-
-- [ ] ...
-
-## Checklist
-
-- Code is [properly formatted](https://docs.nextcloud.com/server/latest/developer_manual/digging_deeper/continuous_integration.html#linting)
-- [Sign-off message](https://github.com/src-d/guide/blob/master/developer-community/fix-DCO.md) is added to all commits
-- [ ] Tests ([unit](https://docs.nextcloud.com/server/latest/developer_manual/app_development/tutorial.html#unit-tests), [integration](https://docs.nextcloud.com/server/latest/developer_manual/app_development/tutorial.html#integration-tests), api and/or acceptance) are included
-- [ ] Screenshots before/after for front-end changes
-- [ ] Documentation ([manuals](https://github.com/nextcloud/documentation/) or wiki) has been updated or is not required
-- [ ] [Backports requested](https://github.com/nextcloud/backportbot/#usage) where applicable (ex: critical bugfixes)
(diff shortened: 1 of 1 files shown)


.github/pull_request_template.md
```diff
@@ -0,0 +1,25 @@
...
+- [ ] ...
+
+## Checklist
+
+- Code is [properly formatted](https://docs.nextcloud.com/server/latest/developer_manual/digging_deeper/continuous_integration.html#linting)
+- [Sign-off message](https://github.com/src-d/guide/blob/master/developer-community/fix-DCO.md) is added to all commits
+- [ ] Tests ([unit](https://docs.nextcloud.com/server/latest/developer_manual/app_development/tutorial.html#unit-tests), [integration](https://docs.nextcloud.com/server/latest/developer_manual/app_development/tutorial.html#integration-tests), api and/or acceptance) are included
+- [ ] Screenshots before/after for front-end changes
+- [ ] Documentation ([manuals](https://github.com/nextcloud/documentation/) or wiki) has been updated or is not required
+- [ ] [Backports requested](https://github.com/nextcloud/backportbot/#usage) where applicable (ex: critical bugfixes)
```
szaimen: I think we need to adjust the template a bit to actually fit the project

szaimen:
@pabzm what do you think about such a template?

.github/pull_request_template.md
docjyJ: Yes, probably. I just duplicated the file from the repo server.

pabzm:
I think such templates can help if too many issues are posted that lack relevant details or are unproductive. 

I'm not really sure if this is the case in this repository. From my current point of view I would say it's not necessary, but maybe I'm wrong.

If its not necessary I would refrain from adding a (complex) template, since reading and filling it might also put some people off, if it takes significantly longer than without a template. But then maybe a lean template could be an option, if you wish for one.

.github/pull_request_template.md
```diff
@@ -0,0 +1,25 @@
...
+- [ ] [Labels added](https://github.com/nextcloud/server/labels) where applicable (ex: bug/enhancement, `3. to review`, feature component)
+- [ ] [Milestone added](https://github.com/nextcloud/server/milestones) for target branch/version (ex: 32.x for `stable32`)
```
szaimen:
```suggestion
* [Sign-off message](https://github.com/src-d/guide/blob/master/developer-community/fix-DCO.md) is added to all commits
```

szaimen (changes requested):
I agree with Pablo and thought a bit more about this.

Lets streamline the template and do it like that:

.github/pull_request_template.md
szaimen: see below

PR #7477, opened 2026-01-22
deprecate nocodb community container

docjyJ:
- Fix https://github.com/nextcloud/all-in-one/discussions/7446


szaimen:

PR #7481, opened 2026-01-23
Fix: Additional Collabora options not working correctly

### LINKED ISSUE #7480: LanguageTool for not working in Collabora // Collabora options not correctly passed down
### Steps to reproduce
1. Try setting up LanguageTool as documented

### Expected behavior 
The collabora options defined in the web interface should be passed to collabora correctly.

### Actual behavior 
Multiply arguments are combined to one argument and thus stop working

MrAn0nym:
- Fixes #7480 
- Also linking to https://github.com/nextcloud/all-in-one/pull/7362
- Fixing https://github.com/nextcloud/all-in-one/discussions/4380#discussioncomment-15495887


szaimen:
Hi, thanks a lot for the PR! :)

Have already tested the changes?

php/src/Docker/DockerActionManager.php
```diff
@@ -428,7 +428,7 @@ public function CreateContainer(Container $container): void {
 
             // Additional Collabora options
             if ($this->configurationManager->GetAdditionalCollaboraOptions() !== '') {
-                $requestBody['Cmd'] = [$this->configurationManager->GetAdditionalCollaboraOptions()];
+                $requestBody['Cmd'] = explode(' ', $this->configurationManager->GetAdditionalCollaboraOptions());
```
szaimen: I remember now why we didnt do it like that. Exploding with space will not work for all configs. Best to split with `--o:` and re-add the string to all items

MrAn0nym: From looking at this [documentation](https://docs.pocoproject.org/current/Poco.Util.OptionProcessor.html) and the Poco [OptionProcessor](https://github.com/austinsc/Poco/blob/master/Util/src/OptionProcessor.cpp) splitting at `" -"` should properly handle all edge cases

php/src/Docker/DockerActionManager.php
```diff
@@ -428,7 +428,7 @@ public function CreateContainer(Container $container): void {
 
             // Additional Collabora options
             if ($this->configurationManager->GetAdditionalCollaboraOptions() !== '') {
-                $requestBody['Cmd'] = [$this->configurationManager->GetAdditionalCollaboraOptions()];
+                $requestBody['Cmd'] = preg_split('/\s+(?=-)/', $this->configurationManager->GetAdditionalCollaboraOptions());
```
MrAn0nym: The regex not not extremely readable, should be the most thorough though.
`\s` is any whitespace
`+` means one or more token
`(?=<characters>)` defines a positive lookahead (requirement but wont remove that character 
 `-` the required character after an space

MrAn0nym: tbh i'm not quite sure if this should rather be done in a loop that's easier to understand but longer :shrug:

szaimen: I am fine to use preg_split but maybe do it like this?
```suggestion
                // TODO: add a comment here what the regex does
                $regEx = '/\s+(?=-)/';
                $requestBody['Cmd'] = preg_split($regEx, $this->configurationManager->GetAdditionalCollaboraOptions());
```

pabzm:
Technically this looks good to me. I'd just suggest to rework the comment above the code to explain what the code does (split CLI options), and to explain what the regexp wants to achieve (split at every ` -` without swallowing anything) rather than what each character means.

And maybe add a little bit of error handling, since `preg_split()` technically can return `false` in case of an error. I don't know how such an error would be triggered in this case, but better safe than sorry, I'd say.

Let me know if you would appreciate a contribution to the branch instead of verbose text :)

szaimen:
> Let me know if you would appreciate a contribution to the branch instead of verbose text :)

Yes, that would be great, thanks! :)

php/src/Docker/DockerActionManager.php
```diff
@@ -420,7 +420,13 @@ public function CreateContainer(Container $container): void {
 
             // Additional Collabora options
             if ($this->configurationManager->collaboraAdditionalOptions !== '') {
-                $requestBody['Cmd'] = [$this->configurationManager->collaboraAdditionalOptions];
+                // The regex not not extremely readable, should be the most thorough though.
+                // \s is any whitespace
+                // + means one or more token
+                // (?=<characters>) defines a positive lookahead (requirement but won't remove that character
+                // - the required character after an space
```
pabzm: I'd like to suggest this comment, which explains the code's intention better IMHO:


```suggestion
                // Split the list of Collabora options, which are stored as a string but must be assigned as an array.
                // To avoid problems with whitespace or dashes in option arguments we use a regular expression
                // that splits the string at every position where a whitespace is followed by '--o:'.
                // The leading whitespace is removed in the split but the following characters are not.
                // Example: "--o:example_config1='some thing' --o:example_config2=something-else" -> ["--o:example_config1='some thing'", "--o:example_config2=something-else"] 
```

php/src/Docker/DockerActionManager.php
```diff
@@ -420,7 +420,13 @@ public function CreateContainer(Container $container): void {
...
+                $regEx = '/\s+(?=-)/';
```
pabzm: I would suggest to match also a character after the dash, in case the argument list might contain substrings like `-o "some - thing"`, which would be split into `-o "some` and `- thing"` with the current Regexp.


```suggestion
                $regEx = '/\s+(?=-[a-zA-Z0-9]+)/';
```

szaimen: I think collabora uses the following string for all options: `--o:`. So we can probalby simply check for that.
```suggestion
                $regEx = '/\s+(?=--o:)/';
```

php/src/Docker/DockerActionManager.php
```diff
@@ -420,7 +420,13 @@ public function CreateContainer(Container $container): void {
...
+                $requestBody['Cmd'] = preg_split($regEx, $this->configurationManager->collaboraAdditionalOptions);
```
```suggestion
                $requestBody['Cmd'] = preg_split($regEx, rtrim($this->configurationManager->collaboraAdditionalOptions));
```

php/src/Docker/DockerActionManager.php
pabzm: Makes it even better! ✨

szaimen:

PR #7501, opened 2026-01-28
Lower case method names in ConfigurationManager

pabzm:
This lower-cases all method names in `ConfigurationManager`, as is good practice in PHP.

Other classes would be touched subsequently, one after another (to avoid too big PR diffs and potential code conflicts).


szaimen (approved):
LGTM, thanks ☺️

PR #7502, opened 2026-01-28
nextcloud-entrypoint.sh: improve GPG key import method

szaimen:
- Try to fix https://github.com/nextcloud/all-in-one/discussions/6940


Containers/nextcloud/entrypoint.sh
```diff
@@ -182,8 +182,7 @@ if ! [ -f "$NEXTCLOUD_DATA_DIR/skip.update" ]; then
             curl -fsSL -o nextcloud.tar.bz2.asc "https://download.nextcloud.com/server/releases/latest-${NEXT_MAJOR}.tar.bz2.asc"
             GNUPGHOME="$(mktemp -d)"
             export GNUPGHOME
-            # gpg key from https://nextcloud.com/nextcloud.asc
-            gpg --batch --keyserver keyserver.ubuntu.com --recv-keys 28806A878AE423A28372792ED75899B9A724937A
+            curl -sSL https://nextcloud.com/nextcloud.asc | gpg --import
```
szaimen: Should also do the same in Dockerfile

Additionally fall back to old method if nextclohd cannot be contacted: `hkp://keyserver.ubuntu.com:80`

PR #7522, opened 2026-01-30
add a workflow that blocks merging if a pre-release was published

szaimen:


szaimen:
The workflow is ready (it is designed to fail currently, so working as intended ☺️)

.github/workflows/fail-on-prerelease.yml
```diff
@@ -0,0 +1,55 @@
...
+              } catch (err) {
+                if (err.status === 404) {
+                  core.info(`No release found for tag ${latestTag}; skipping prerelease check.`);
+                } else {
+                  throw err;
+                }
+              }
+
+            } catch (error) {
+              throw error;
```
pabzm: Why wrap this into a `try/catch` if we just re-throw the exception?

.github/workflows/fail-on-prerelease.yml
```diff
@@ -0,0 +1,55 @@
+name: Fail on prerelease
+
+on:
+  pull_request:
```
pabzm, line 4: Maybe add `workflow_dispatch` to the list, so this can be triggered manually after a final release was published (instead of having to push into the PR's branch again)?

.github/workflows/fail-on-prerelease.yml
```diff
@@ -0,0 +1,55 @@
+name: Fail on prerelease
```
pabzm: I'd suggest to rephrase this so it's comprehensible to contributors, already without having to read the job's output.

Maybe like this:


```suggestion
name: Block if prerelease is present
```

.github/workflows/fail-on-prerelease.yml
szaimen: We can simply re-try the workflow in that case. workflow_dispatch is not needed for that :)

.github/workflows/fail-on-prerelease.yml
szaimen: good idea and done :)

.github/workflows/fail-on-prerelease.yml
szaimen: Good point! Removed that :)

.github/workflows/fail-on-prerelease.yml
pabzm: The line is unchanged as far as I can see

szaimen: you are right, sorry. Fixed now

szaimen:
Thanks for the review Pablo! ☺️

PR #7532, opened 2026-02-02
Wrap ConfigurationController#SetConfig into a "transaction"

### LINKED ISSUE #7500: Wrap multiple config setters into a "transaction"
We want to wrap multiple writes to the config file into a "transaction" where they occur. This would speed up the execution and reduce IO-stress to the filesystem.

First candidate is `ConfigurationController#SetConfig()`

Follow-up from https://github.com/nextcloud/all-in-one/pull/7392

pabzm:
This avoids a lot of subsequent writes and reads from the file system, because now only commitTransaction() actually writes the config file.

* Close #7500 
* [Sign-off message](https://github.com/src-d/guide/blob/master/developer-community/fix-DCO.md) is added to all commits

### DIFF
--- a/php/src/Controller/ConfigurationController.php
+++ b/php/src/Controller/ConfigurationController.php
@@ -17,6 +17,7 @@ readonly class ConfigurationController {
 
     public function SetConfig(Request $request, Response $response, array $args): Response {
         try {
+            $this->configurationManager->startTransaction();
             if (isset($request->getParsedBody()['domain'])) {
                 $domain = $request->getParsedBody()['domain'] ?? '';
                 $skipDomainValidation = isset($request->getParsedBody()['skip_domain_validation']);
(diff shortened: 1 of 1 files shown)


php/src/Controller/ConfigurationController.php
```diff
@@ -137,6 +138,8 @@ public function SetConfig(Request $request, Response $response, array $args): Re
         } catch (InvalidSettingConfigurationException $ex) {
             $response->getBody()->write($ex->getMessage());
             return $response->withStatus(422);
+        } finally {
+            $this->configurationManager->startTransaction();
```
szaimen:
```suggestion
            $this->configurationManager->commitTransaction();
```

szaimen (approved):
LGTM, thanks! ☺️

PR #7541, opened 2026-02-03
add `declare(strict_types=1);` to all php files

szaimen:


php/domain-validator.php
```diff
@@ -1,4 +1,5 @@
 <?php
+declare(strict_types=1);
 
 $domain = $_GET['domain'] ?? '';
```
szaimen: maybe?

```suggestion
$domain = (string) $_GET['domain'] ?? '';
```

pabzm: Why? `$_GET[]` returns only strings or arrays of strings, AFAIK. If `$_GET['domain']` would be an array (in case someone tries to have fun), then `$domain` would become `"Array"` with this typecast.

To protect against that I would prefer some old-school type checking, because it's more obvious what the code is about:

```php
if (isset($_GET['domain']) && is_string($_GET['domain'])) {
    $domain = $_GET['domain'];
} else {
    $domain = '';
}
```

pabzm:
I do like to introduce strict typing, thanks! 🤩

php/domain-validator.php
pabzm: Or maybe as one-liner:

```php
$domain = filter_input(INPUT_GET, 'domain', FILTER_SANITIZE_URL) ?? '';
```

pabzm: Better:

```php
$domain = (string) filter_input(INPUT_GET, 'domain', FILTER_SANITIZE_URL);
```

pabzm: (Because `filter_input` might return `false`, too.)

szaimen: I will go with thr first suggestion. Thanks! :)

szaimen:
```suggestion
$domain = '';
if (isset($_GET['domain']) && is_string($_GET['domain'])) {
    $domain = $_GET['domain'];
}
```

pabzm (approved):
Nice structural improvement, thank you! 🧱💙

szaimen:
This is now released with v12.7.0 Beta. Testing and feedback is welcome! See https://github.com/nextcloud/all-in-one#how-to-switch-the-channel

PR #7587, opened 2026-02-13
aio-interface: show which containers have an update available

szaimen:
- Close https://github.com/nextcloud/all-in-one/issues/7572

| Before | After |
|---|---|
| image | image |


szaimen:
CI failure unrelated

Issue #7567, opened 2026-02-10
Add install docs for unraid


szaimen:
I found this: https://share.google/aimode/uySLUocD4RjrMWVY8

Because of https://www.reddit.com/r/unRAID/comments/1r0gdhi/unraid_keeps_orphaning_my_docker_images_on_a/

Also maybe document how the container might need to be started...

`ExecStartPost=/usr/bin/docker exec --env START_CONTAINERS=1 nextcloud-aio-mastercontainer /daily-backup.sh`

Also see https://github.com/nextcloud/all-in-one/discussions/7469

PR #7606, opened 2026-02-16
readme: add install docs for unraid

szaimen:
- Close https://github.com/nextcloud/all-in-one/issues/7567
- See https://github.com/nextcloud/all-in-one/discussions/7469

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -88,9 +88,10 @@ Included are:
 ## How to use this?
 
 The steps below are written for Linux. For platform-specific guidance see:
-- macOS: [How to run AIO on macOS](#how-to-run-aio-on-macos)
-- Windows: [How to run AIO on Windows](#how-to-run-aio-on-windows)
-- Synology DSM: [How to run AIO on Synology DSM](#how-to-run-aio-on-synology-dsm)
+- macOS: [How to run AIO on macOS?](#how-to-run-aio-on-macos)
+- Windows: [How to run AIO on Windows?](#how-to-run-aio-on-windows)
+- Unraid: [How to run AIO on Unraid?](#how-to-run-aio-on-unraid)
+- Synology DSM: [How to run AIO on Synology DSM?](#how-to-run-aio-on-synology-dsm)
 - TrueNAS SCALE: [Can I run AIO on TrueNAS SCALE?](#can-i-run-aio-on-truenas-scale)
 
 > [!IMPORTANT]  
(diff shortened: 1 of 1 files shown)


readme.md
```diff
@@ -598,6 +600,23 @@ Also, you may be interested in adjusting Nextcloud's Datadir to store the files
 > [!NOTE]  
 > Almost all commands in this project's documentation use `sudo docker ...`. Since `sudo` is not available on Windows, you simply remove `sudo` from the commands and they should work.  
 
+### How to run AIO on Unraid?
+
+Generally on Unraid, before installing Nextcloud AIO, you should make sure to specify a directory for all docker related files instead of using the default docker image before continuing the installation process. See this guide for more details: https://forums.unraid.net/topic/103924-how-to-turn-my-docker-file-to-a-docker-folder/#comment-962317
```
pabzm, line 605: Maybe we could add a reason for this suggestion. I don't know much about unraid, so maybe that's my own fault, but I immediately wondered why this would be necessary.

readme.md
```diff
@@ -598,6 +600,23 @@ Also, you may be interested in adjusting Nextcloud's Datadir to store the files
...
+
+Additionally, you might need to create a custom user script to start all sibling containers of AIO automatically whenever the server reboots (since the docker plugin in Unraid does not respect the normal docker restart-policy that is configured for all AIO containers). See the User scripts docs: https://docs.unraid.net/unraid-os/using-unraid-to/run-docker-containers/managing-and-customizing-containers/#user-scripts-plugin. You would need to use a script like the one below and use the schedule option `@reboot` to start the containers automatically.
+```bash
+#!/bin/bash
+
+# Wait 120 seconds for other server components to start correctly
+sleep 120
+
+# Execute the actual command
+docker exec --env START_CONTAINERS=1 nextcloud-aio-mastercontainer /daily-backup.sh`
```
pabzm, line 615: The description says that this script would start all containers after a reboot, but actually it performs a backup, I think?

Either the description or the ticket would need an adjustment, from my understanding. I'm not sure which.

readme.md
szaimen: This is necessary because in some cases the docker image file gets recreated and loses all persistent data like docker volumes. Also this image is apparently not resistant for unexpected restarts like power loss of the server and can break unexpectedly. See also for example https://www.reddit.com/r/unRAID/comments/1r0gdhi/unraid_keeps_orphaning_my_docker_images_on_a/

readme.md
szaimen: No, it does not perform a backup. See https://github.com/nextcloud/all-in-one?tab=readme-ov-file#how-to-stopstartupdate-containers-or-trigger-the-daily-backup-from-a-script-externally

The daily-backup.sh file is simply a wrapper for all this logic.

Issue #7601, opened 2026-02-15
Document how to add extra hosts via /etc/hosts file to containers


szaimen:
Linking to https://github.com/nextcloud/all-in-one/discussions/2395

PR #7607, opened 2026-02-16
readme: add guide how to add extra hosts to the containers

szaimen:
- Close https://github.com/nextcloud/all-in-one/issues/7601
- See https://github.com/nextcloud/all-in-one/discussions/2395

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -380,7 +380,7 @@ Apart from that there is now a community container that can be added to the AIO 
 
 For some use cases, you might need to overwrite the local DNS resolution of some domains inside the containers. On Linux, you can do so either by using a local DNS server as described in the section above and add a local DNS entry into the dns server and make your containers use that DNS server.
 
-Another solution on Linux is to simply edit the `/etc/hosts` file and add your custom entry like `172.18.0.1 mail.example.com` as additional line to the file and restart docker which should automatically push the entry to all docker containers.
+On Linux, depending on your network and docker configuration, it might also work to simply edit the `/etc/hosts` file. Add your custom entry like `172.18.0.1 mail.example.com` as additional line to the file and restart docker which should automatically push the entry to all docker containers.
 
 ### How to skip the domain validation?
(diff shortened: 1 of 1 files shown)


readme.md
```diff
@@ -375,6 +376,12 @@ Now that this is out of the way, the recommended way how to access Nextcloud loc
 - https://dockerlabs.collabnix.com/intermediate/networking/Configuring_DNS.html
 Apart from that there is now a community container that can be added to the AIO stack: https://github.com/nextcloud/all-in-one/tree/main/community-containers/pi-hole
 
+### How to overwrite the local DNS resolution for some domains or add extra hosts to the containers?
+
+For some use cases, you might need to overwrite the local DNS resolution of some domains inside the containers. On Linux, you can do so either by using a local DNS server as described in the section above and add a local DNS entry into the dns server and make your containers use that DNS server.
+
+Another solution on Linux is to simply edit the `/etc/hosts` file and add your custom entry like `172.18.0.1 mail.example.com` as additional line to the file and restart docker which should automatically push the entry to all docker containers.
```
pabzm: This surprised me, and https://docs.docker.com/engine/network/#custom-hosts also says:

> Custom hosts, defined in /etc/hosts on the host machine, aren't inherited by containers.

Did you test that locally? Is it maybe a platform-specific feature?

Generally I would suggest to avoid documenting 3rd-party-software, but rather to link to its own documentation, to avoid having additional work with keeping it up to date (a link checker to detect broken URLs is extremely easy to run from a github workflow, checking if written documentation is still correct is a human's task).

szaimen: > Did you test that locally? Is it maybe a platform-specific feature?

Yes, actually this works on my Ubuntu server test server. Might be possible that a platform specific feature.

```suggestion
Another solution on Linux is to simply edit the `/etc/hosts` file and add your custom entry like `172.18.0.1 mail.example.com` as additional line to the file and restart docker which should automatically push the entry to all docker containers. (This is confirmed to work on Ubuntu Server).
```

pabzm:
```suggestion
On Linux, depending on your network and docker configuration, it might also work to simply edit the `/etc/hosts` file. Add your custom entry like `172.18.0.1 mail.example.com` as additional line to the file and restart docker which should automatically push the entry to all docker containers.
```

pabzm: I would suggest to rephrase the paragraph like this, since this behaviour appears to depend on networking and docker config details.

szaimen: Looks good, thanks! :)

szaimen:
Thanks for the review Pablo ☺️

Issue #7600, opened 2026-02-15
Add docs how to apply different uids


szaimen:
See https://www.reddit.com/r/NextCloud/comments/1bwf4m2/comment/o5hyaud/?utm_source=share&utm_medium=mweb3x&utm_name=mweb3xcss&utm_term=1&utm_content=share_button

Also see https://github.com/nextcloud/all-in-one/discussions/1974

PR #7608, opened 2026-02-16
readme: describe how to configure custom UID/GID

szaimen:
- Close https://github.com/nextcloud/all-in-one/issues/7600
- See https://github.com/nextcloud/all-in-one/discussions/1974


readme.md
```diff
@@ -453,6 +454,18 @@ You can configure the Nextcloud container to use a specific directory on your ho
     ```
     In this example, it would mount `E:\your\data\path` into the volume so for a different location you need to adjust `/host_mnt/e/your/data/path` accordingly.
 
+### How to configure custom UID/GID?
+There are two ways to configure custom UIDs or GIDs that will be used inside the installation.
```
pabzm, line 458:
```suggestion
### How to access content files using a custom UID/GID?
Files containing Nextcloud user content belong to the user and group ID of the system account that is running the PHP process inside the container (currently `33` and `33`). If you need write access from the host system to those files while using a different, unchangeable user or group ID, you have two options:
```

PR #7616, opened 2026-02-17
Docs: clarify filtering of archives during borg backup pruning

th0rgall:
Clarify the retention policy adjustment instructions by mentioning to not include the '-a' or '--glob-archives' option.

I included `--glob-archives="*-nextcloud-aio"` the retention policy env variable, and it lead to a borg pruning error due a the duplicate `-a` ... `--glob-archives` options. Only one of these options can be provided, since they are the same, and borg doesn't accept duplicates.

I added `--glob-archives` to the retention policy because I didn't know that AIO already provided it by default, and I was worried that it would prune out other non-Nextcloud-AIO backups in the repo. The solution was to remove my unnecessary addition.

At the moment of writing, these are the default prune options in AIO:

https://github.com/nextcloud/all-in-one/blob/58e5d8534b115614c9333966694ac767ecb9a6d9/Containers/borgbackup/backupscript.sh#L236

This PR is the smallest addition I could think of to clarify that filtering is built-in.

As an alternative to this edit, Nextcloud AIO's default filtering behavior could be clarified elsewhere is a positive way.

Two ideas:

1. Modify the start of the retention policy FAQ answer, like:
   
   > After a backup, AIO prunes your borg archives with the `--glob-archives '*_*-nextcloud-aio'` option, so that it does not affect archives unrelated to AIO. The built-in retention policy is `--keep-within=7d --keep-weekly=4 --keep-monthly=6`. [...]
   
2. Or perhaps, there can be another FAQ question about backups, like _"Will Nextcloud AIO touch other backups in my borg repository?"_ or _"Can I use my borg repository for backups unrelated to Nextcloud AIO?",_ with the answer:

   > No/Yes, AIO internally uses the borg `--glob-archives '*_*-nextcloud-aio'` option to only prune AIO archives, so Nextcloud AIO will not affect archives/backups unrelated to AIO. Be careful that other systems don't prune your AIO archives, however.


readme.md
```diff
@@ -856,7 +856,7 @@ Be aware that this solution does not back up files and folders that are mounted
...
+The built-in borg-based backup solution has by default a retention policy of `--keep-within=7d --keep-weekly=4 --keep-monthly=6`. See https://borgbackup.readthedocs.io/en/stable/usage/prune.html for what these values mean. You can adjust the retention policy by providing `--env BORG_RETENTION_POLICY="--keep-within=7d --keep-weekly=4 --keep-monthly=6"` to the docker run command of the mastercontainer (but before the last line `ghcr.io/nextcloud-releases/all-in-one:latest`! If it was started already, you will need to stop the mastercontainer, remove it (no data will be lost) and recreate it using the docker run command that you initially used) and customize the value to your fitting. ⚠️ Please make sure that this value is valid, otherwise backup pruning will bug out! Also, don't include the `-a` or `--glob-archives` option, since AIO [already provides it](https://github.com/nextcloud/all-in-one/blob/58e5d8534b115614c9333966694ac767ecb9a6d9/Containers/borgbackup/backupscript.sh#L236) and you can't override it.
```
```suggestion
The built-in borg-based backup solution has by default a retention policy of `--keep-within=7d --keep-weekly=4 --keep-monthly=6`. See https://borgbackup.readthedocs.io/en/stable/usage/prune.html for what these values mean. You can adjust the retention policy by providing `--env BORG_RETENTION_POLICY="--keep-within=7d --keep-weekly=4 --keep-monthly=6"` to the docker run command of the mastercontainer (but before the last line `ghcr.io/nextcloud-releases/all-in-one:latest`! If it was started already, you will need to stop the mastercontainer, remove it (no data will be lost) and recreate it using the docker run command that you initially used) and customize the value to your fitting. ⚠️ Please make sure that this value is valid, otherwise backup pruning will bug out! Also, don't include the `-a` or `--glob-archives` option, since AIO already provides it and you can't override it. See https://github.com/nextcloud/all-in-one/pull/7616
```

th0rgall:
Thanks for reviewing this so quickly! I like the idea of simply linking to this PR for more context, rather than a code line that may get outdated. I'll still add that line in the opening post of this PR for clarification, since it's now used as a reference.

PR #7623, opened 2026-02-18
update mastercontainer to php 8.5

szaimen:
- Close https://github.com/nextcloud/all-in-one/issues/7197


szaimen:
Playwright test is expected to fail at this point as the container image is still running php 8.4 and thus not compatible with php 8.5 requirements

szaimen:

PR #7631, opened 2026-02-20
aio-interface: 404 error handler for less app output pollution

pabzm:
This introduces a custom error handler for 404 responses, so they don't show up as exceptions in the process output anymore.


szaimen:

Issue #7581, opened 2026-02-12
Auto-refresh the container logs page and scroll to the latest logs


szaimen:
Possibly use streaming response. Will be possible after https://github.com/nextcloud/all-in-one/pull/7458

szaimen:
@pabzm can we after https://github.com/nextcloud/all-in-one/pull/7458 also use a streaming response for this? :)

pabzm:
Not really, because the process will run into a timeout after a while. Might be a long while, but a finite while.

So we some code for reloading anyway, and then we can also use that from the beginning. Also using a long-running streaming response would probably lock the main web UI (since we allow only one request per client), which I think wouldn't be good UX.

And for scrolling to the bottom we also need a sprinkle of Javascript, so we don't get around that in the first place.

pabzm:
But I think injecting a small javascript file is not much work, I'll implement that.

szaimen:
> But I think injecting a small javascript file is not much work, I'll implement that.

Thanks ☺️

PR #7634, opened 2026-02-20
aio-interface: Auto-reload the log view, and scroll to bottom.

pabzm:
Includes a button to disable the automatic reloading (useful when inspecting some lines in the middle).

The button is fixed in the top right corner of the view and mostly unstyled:

Screenshot 2026-02-20 at 14 24 42

When clicked it changes to this appearance:

image

--------

- Closes https://github.com/nextcloud/all-in-one/issues/7581

### DIFF
--- a/php/templates/log.twig
+++ b/php/templates/log.twig
@@ -13,13 +13,26 @@
                 // Give the browser a short moment to render all text and calculate the scrollHeight, to avoid
                 // problems with scrolling to the bottom.
                 setTimeout(() => window.scrollTo(0, document.body.scrollHeight), 100);
-                const reloadTimer = setTimeout(() =>  location.reload(), 10000);
+                // Reload after a short while if the window is visible to the user. 
+                const reloadTimer = setTimeout(() =>  {
+                    if (document.visibilityState === 'visible') {
+                        location.reload()
+                    }
+                }, 5000);
+                // Provide a button that allows to disable the reloads.
                 const button = document.querySelector('button');
                 document.querySelector('button').addEventListener('click', (event) => {
                     event.preventDefault();
                     clearTimeout(reloadTimer);
                     button.disabled = true;
                     button.textContent = 'Reloading was disabled';
+                });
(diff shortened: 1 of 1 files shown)


php/templates/log.twig
```diff
@@ -0,0 +1,32 @@
...
+                right: 1rem;
+                font-size: 1.3rem;
+            }
+        </style>
+        <script>
+            document.addEventListener("DOMContentLoaded", () => {
+                // Give the browser a short moment to render all text and calculate the scrollHeight, to avoid
+                // problems with scrolling to the bottom.
+                setTimeout(() => window.scrollTo(0, document.body.scrollHeight), 100);
+                const reloadTimer = setTimeout(() =>  location.reload(), 10000);
```
szaimen: Maybe 10s is a bit too slow? Wdyt about 5s?

---

Also we cpuld maybe add a function that disables auto reloading if the focus is removed from the page?

Something like https://github.com/nextcloud/all-in-one/blob/1ddaa30b0ff22a984765b175e81197f455cdfa75/php/public/automatic_reload.js#L2

szaimen (commented):
Thanks a lot! 😊

One comment below:

php/templates/log.twig
szaimen: Something like this:
```suggestion
                const reloadTimer = setTimeout(() =>  if (document.hasFocus()) { location.reload() }, 5000);
```

pabzm: I'm not sure about stopping the reload, because I think it could be confusing if sometimes the window content shows the latest log entries, and sometimes (if I shortly put another window into the focus) it doesn't. At least if we don't resume the automatic reloading.

Could you elaborate on what you envision exactly?

szaimen: I just imagine a case where someone opens a lot of log pages and all of them auto-reload in the background. This would bombard the server with requests... So I would like to limit this. Maybe we could though trigger another reload automatically if the page gets focus again without having to wait another 5 seconds?

szaimen:

PR #7650, opened 2026-02-24
Auto reload with less flickering

### LINKED ISSUE #7648: auto-reloading leads to the page jumping
If you open a container log with a lot of lines, e.g. when restoring a backup, it will load the lines and then scroll down which is fine. However auto-reloading will do this "jumping" in a loop. Maybe it might be possible to cache the already written lines in browser and only load changed lines further down so that we do not have this "jumping" behaviour?

- Follow-up to https://github.com/nextcloud/all-in-one/pull/7634

pabzm:
Closes #7648

### DIFF
--- a/php/templates/log.twig
+++ b/php/templates/log.twig
@@ -26,7 +26,7 @@
                 }, 5000);
                 // Provide a button that allows to disable the reloads.
                 const button = document.querySelector('button');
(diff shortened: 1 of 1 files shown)


szaimen:
Superceded by https://github.com/nextcloud/all-in-one/pull/7656

PR #7656, opened 2026-02-24
aio-interface: auto-load only new log data

### LINKED ISSUE #7648: auto-reloading leads to the page jumping

pabzm:
This is way more complicated that just reloading the log file, but also way nicer.

- Closes https://github.com/nextcloud/all-in-one/issues/7648
- Follow-up to https://github.com/nextcloud/all-in-one/pull/7634

### DIFF
--- a/php/public/index.php
+++ b/php/public/index.php
@@ -11,6 +11,7 @@ ini_set('max_execution_time', '7200');
 ini_set('log_errors_max_len', '0');
 
 use DI\Container;
+use DI\NotFoundException;
 use Slim\Csrf\Guard;
 use Slim\Factory\AppFactory;
 use Slim\Views\Twig;
@@ -170,6 +171,15 @@ $app->get('/setup', function (Request $request, Response $response, array $args)
             'password' => $setup->Setup(),
         ]
     );
+});
+$app->get('/log', function (Request $request, Response $response, array $args) use ($container) {
+    $params = $request->getQueryParams();
+    $id = $params['id'] ?? '';
+    if (!str_starts_with($id, 'nextcloud-aio-')) {
+        throw new DI\NotFoundException();
+    }
+    $view = Twig::fromRequest($request);
+    return $view->render($response, 'log.twig', ['id' => $id]);
 });
 
 // Auth Redirector
--- a/php/public/log-view.js
+++ b/php/public/log-view.js
@@ -1,0 +1,142 @@
+class LogViewer {
+    // Configure the interval in seconds for autoloading log data.
+    autoloadIntervalSec = 5;
+    // Set to true to see some debug log statements in the browser console.
+    debugLog = false;
+
+    // Don't touch these, please.
+    containerId;
+    apiBaseUrl = 'api/docker/logs';
+    autoloadIntervalId = null;
+    logElem;
+    lastLogTimestamp = '';
+    autoloadingDisabledFromButton = false;
+    loaderElem;
+    dataLoadingLock;
+
+    constructor() {
+        const id = document.body.dataset.containerId;
+        if (typeof(id) !== 'string' || !id.startsWith('nextcloud-aio-')) {
+            throw new Exception('Invalid container ID');
+        }
+        this.containerId = id;
+        this.logElem = document.querySelector('pre');
+        this.loaderElem = document.querySelector('.loader');
+        this.initAutoloadingControls();
+        // Enable automatic log data loading.
+        this.startAutoloading();
+    }
+
+    startAutoloading() {
+        // Load log data immediately.
+        this.loadAndAppendLogData();
+        // Load new log data repeatedly.
+        this.debug("Starting autoloading");
+        this.autoloadIntervalId = setInterval(() => {
+            if (this.isAutoloadingEnabled()) {
+                this.loadAndAppendLogData();
+            }
+        }, 5000);
+    }
+
+    stopAutoloading() {
+        this.debug("Stopping autoloading");
+        clearInterval(this.autoloadIntervalId);
+        this.autoloadIntervalId = null;
+    }
+
+    isAutoloadingEnabled() {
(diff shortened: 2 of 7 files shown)


pabzm:
I forgot to add a description:

This changes the access to logs so that a proper view is rendered, from which Javascript is pulling the actual logs.

To avoid page jumping all log lines are appended one by one, and immediately scrolled to it.

The subsequent log fetches only ask for lines newer than the docker-generated datetime of the last known line. This datetime is included in the request to the docker API, so no lines are uselessly sent around over and over.

The auto-loading gets disabled whenever the page's visibility turns to `hidden` (e.g. if a another browser tab gets the focus, or the whole window is not visible on the display. When re-focussing the tab, the auto-loading is continued (unless it was disabled through the button).

The button in the top right corner now has two states, to disable and to re-enable auto-loading. This would be handy e.g. if one would want to compare log lines higher up on the page (which the automatic scrolling to the bottom might make cumbersome).

> Screenshot 2026-02-24 at 23 42 18

> image

Reloading the page resets all state and pulls all log lines freshly.

php/public/log-view.js
```diff
@@ -0,0 +1,118 @@
+class LogViewer {
+    // Configure the interval in seconds for autoloading log data.
+    autoloadIntervalSec = 5;
+
+    // Don't touch these, please.
+    containerId;
+    apiBaseUrl = '/api/docker/logs';
```
szaimen: I am not 100% sure but we might should use `api/docker/logs` here

php/templates/log.twig
```diff
@@ -1,45 +1,32 @@
...
-            });
-        </script>
+        <script src="/log-view.js"></script>
     </head>
-    <body>
-        <button>Disable auto-reloading</button>
+    <body data-container-id="{{ id }}">
+        <div id="autoloading-box">
+            <div>
+                Automatic loading of new log data is
```
szaimen: can we maybe add some styling for the text? So adjusting the text family? I wonder if it would be possible to load the usual css file here?

php/public/index.php
```diff
@@ -171,6 +172,15 @@
         ]
     );
 });
+$app->get('/log', function (Request $request, Response $response, array $args) use ($container) {
+    $params = $request->getQueryParams();
```
szaimen, line 176: do we need to add a check for authentication here or is this by default covered by the authentication middleware?

szaimen (commented):
Thanks a lot! Looks like a great improvement to me ☺️

However didnt test yet and have some comments below:

php/src/Controller/DockerController.php
```diff
@@ -353,4 +358,38 @@ private function getStreamingResponseHtmlStart() : string {
     private function getStreamingResponseHtmlEnd() : string {
         return "\n  </body>\n</html>";
     }
+
+    protected function getTimestampForDockerLogsApiSince(string $input) : string
```
szaimen: to be consistent would be better to use private

php/public/index.php
pabzm: That is covered by the middleware, I just verified it.

pabzm:
I discovered that loading big logs (some thousand lines) would take a lot of time with the previous solution (e.g. 20 seconds for 5000 lines).

Now this is changed to inserting all received text into the DOM at once, which is significantly faster (300ms for 5000 lines) and renders visually equally nice, I think.

Since 300ms is still noticable I added a loading spinner. The spinner also shows up on autoloading new data, since in slow networks that maybe could take a moment, too.

This is how it looks now:


https://github.com/user-attachments/assets/69ce216c-3818-4c1b-904d-679e40430bc7

szaimen:
> I discovered that loading big logs (some thousand lines) would take a lot of time with the previous solution (e.g. 20 seconds for 5000 lines).
> 
> Now this is changed to inserting all received text into the DOM at once, which is significantly faster (300ms for 5000 lines) and renders visually equally nice, I think.
> 
> Since 300ms is still noticable I added a loading spinner. The spinner also shows up on autoloading new data, since in slow networks that maybe could take a moment, too.
> 
> This is how it looks now:
> 
>  Screen.Recording.2026-02-25.at.14.54.16.mov

Great, thanks! :)

Would it be possible to make the spinner a bit smaller and move it directly to the left of the floating element that shows that auto-loading is enabled? :)
I think having in the middle of the screen is a bit too much...

pabzm:
Like this? (This is recorded with a log of 5100 lines and a network throttled to "fast 4G").

https://github.com/user-attachments/assets/f4f7cb31-7278-41f7-a350-1be0e89471f0

szaimen:
> Like this? (This is recorded with a log of 5100 lines and a network throttled to "fast 4G").
> 
>  Screen.Recording.2026-02-25.at.15.56.06.mov

Yes, looks good, thanks! 😊

szaimen (approved):
LGTM but didnt test yet. Thanks a lot! 😊

szaimen:
Feel free to merge yourself :)

pabzm:
🎉

szaimen:
🎉🎉🎉🎉🎉

szaimen:

Silun:
Is there a way to just simply disable this altogether?

Or alternatively, could it fallback to the way it was before?

To me, this is a fix for something that wasn't broken. On the contrary, ad and script blockers break the functionality of this viewer and that means I just get a blank screen with a non-working button.

image

For clarity, it was broken in this way since the day it was released, I just figured a fix was coming.

Issue #7664, opened 2026-02-26
Extend showing substeps to further endpoints


szaimen:
We could maybe also show sub-steps for stopping containers, starting backup container and restore and others.

endpoints:
- [x] /api/docker/start https://github.com/nextcloud/all-in-one/pull/7458
  - [ ] /api/docker/watchtower https://github.com/nextcloud/all-in-one/pull/7674
  - [ ] /api/docker/backup-check https://github.com/nextcloud/all-in-one/pull/7675
  - [ ] /api/docker/backup-list
  - [ ] /api/docker/backup-check-repair
  - [x] ~~/api/docker/getwatchtower~~ (doesnt make sense)
- [ ] /api/docker/stop https://github.com/nextcloud/all-in-one/pull/7676
  - [ ] /api/docker/backup
  - [ ] /api/docker/restore
  - [ ] /api/docker/backup-test
- [x] ~~/api/docker/logs~~ (doesnt make sense)

szaimen:
Done

szaimen:
This is now released with v12.9.0 Beta. Testing and feedback is welcome! See https://github.com/nextcloud/all-in-one#how-to-switch-the-channel

PR #7674, opened 2026-02-27
Dockercontroller: refactor streaming responses into dedicated functions and use them  in StartWatchtowerContainer method

szaimen:
- Close part of https://github.com/nextcloud/all-in-one/issues/7664

### DIFF
--- a/php/src/Controller/DockerController.php
+++ b/php/src/Controller/DockerController.php
@@ -202,23 +202,9 @@ readonly class DockerController {
             error_log('WARNING: Not pulling container images. Instead, using local ones.');
         }
         
-        $nonbufResp = $response
-            ->withBody(new NonBufferedBody())
-            ->withHeader('Content-Type', 'text/html; charset=utf-8')
-            ->withHeader('X-Accel-Buffering', 'no')
-            ->withHeader('Cache-Control', 'no-cache');
-            
-        // Text written into this body is immediately sent to the client, without waiting for later content.
-        $streamingResponseBody = $nonbufResp->getBody();
-        
-        $streamingResponseBody->write($this->getStreamingResponseHtmlStart());
-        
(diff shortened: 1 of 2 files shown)


php/src/Controller/DockerController.php
```diff
@@ -354,7 +348,37 @@ private function getStreamingResponseHtmlStart() : string {
             
         END;
     }
-    
+
+    private function getStreamingResponseStart(Response $response) : Response {
```
pabzm: This should be named `startStreamingResponse()` or similar, IMHO. The `get` prefix really doesn't fit, I think.

php/src/Controller/DockerController.php
```diff
@@ -354,7 +348,37 @@ private function getStreamingResponseHtmlStart() : string {
...
+        // if it'll actually pull an image), but which should not need to know anything about the
+        // wanted markup or formatting.
+        $addToStreamingResponseBody = function (Container $container, string $message) use ($nonbufResp) : void {
+            $nonbufResp->getBody()->write("<div>{$container->displayName}: {$message}</div>");
+        };
+
+        return $addToStreamingResponseBody;
+    }
+
+    private function endStreamingResponse(Response $nonbufResp) : void {
```
pabzm: The method doesn't actually end the response, it just adds an ending to the HTML.

Maybe calling it `finalizeStreamingResponse()` would be more fitting?

pabzm (changes requested):
Besides my remarks about the method names this looks good to me!

php/src/Controller/DockerController.php
szaimen: done :)

php/src/Controller/DockerController.php
szaimen: Sounds good and done :)

PR #7677, opened 2026-02-27
aio-interface: offer system prune button

### LINKED ISSUE #7666: While containers are running, offer button to run docker system prune

szaimen:
- Close https://github.com/nextcloud/all-in-one/issues/7666
- https://docs.docker.com/reference/api/engine/version/v1.44/

### DIFF
--- a/php/src/Docker/DockerActionManager.php
+++ b/php/src/Docker/DockerActionManager.php
@@ -997,4 +997,71 @@ readonly class DockerActionManager {
             return $this->dockerHubManager->GetLatestDigestOfTag($imageName, $tag);
         }
     }
+
+    public function SystemPrune(?\Closure $addToStreamingResponseBody = null): void {
+        $endpoints = [
+            // Remove stopped containers
+            'containers/prune',
+            // Remove unused images
+            'images/prune',
+            // Remove unused volumes
+            'volumes/prune',
+            // Remove unused networks
+            'networks/prune',
+            // Prune build cache
+            'build/prune',
+        ];
+
+        foreach ($endpoints as $endpoint) {
+            // Special-case images prune to include the dangling filter as requested
+            if ($endpoint === 'images/prune') {
+                $filters = json_encode(['dangling' => ['false']]);
+                $url = $this->BuildApiUrl($endpoint . '?filters=' . urlencode((string) $filters));
+            } else {
+                $url = $this->BuildApiUrl($endpoint);
+            }
+
+            if ($addToStreamingResponseBody !== null) {
+                $addToStreamingResponseBody("Running $endpoint...");
+            }
+
+            try {
+                $response = $this->guzzleClient->post($url);
+                if ($addToStreamingResponseBody !== null) {
+                    $data = json_decode((string)$response->getBody(), true);
+                    $deleted = 0;
+                    foreach (['ContainersDeleted', 'ImagesDeleted', 'VolumesDeleted', 'NetworksDeleted', 'CachesDeleted'] as $key) {
+                        if (isset($data[$key]) && is_array($data[$key])) {
+                            $deleted += count($data[$key]);
+                        }
+                    }
+                    $reclaimed = $data['SpaceReclaimed'] ?? 0;
+                    $parts = [];
+                    if ($deleted > 0) {
+                        $parts[] = "$deleted item(s) deleted";
+                    }
+                    if ($reclaimed > 0) {
+                        $i = (int)floor(log($reclaimed, 1024));
+                        $parts[] = 'Space reclaimed: ' . (string)round($reclaimed / (1024 ** $i), 2) . ' ' . ['B','KB','MB','GB'][$i];
+                    }
+                    $addToStreamingResponseBody(!empty($parts) ? implode('. ', $parts) . '.' : 'Nothing to prune.');
+                }
+            } catch (RequestException $e) {
+                error_log(sprintf('Docker prune (%s) failed: %s', $endpoint, $e->getMessage()));
+                if ($addToStreamingResponseBody !== null) {
+                    $addToStreamingResponseBody('Error: ' . $e->getMessage());
+                }
+                // continue with next prune step
+            }
+        }
+
+        if ($addToStreamingResponseBody !== null) {
+            $addToStreamingResponseBody("Docker system prune completed.");
+            sleep(1);
+
+            // We automatically reload after 10s so that the output can be read or copied if necessary
+            $addToStreamingResponseBody("Automatically reloading the page after 10s.");
+            sleep(10);
+        }
+    }
 }
--- a/php/templates/containers.twig
+++ b/php/templates/containers.twig
@@ -582,6 +582,18 @@
 
                         {% if is_backup_container_running == false %}
                             {% if isApacheStarting == false %}
+                                {% if isAnyRunning == true %}
+                                    <h2>Docker System Prune</h2>
+                                    <details>
+                                        <summary>Click here to reveal a button to prune the docker system.</summary>
+                                        <p>By clicking the button below you can run "docker system prune -a". This will remove unused images, containers, networks, volumes and build cache. It will not delete data of running containers.</p>
+                                        <form method="POST" action="api/docker/prune" target="overlay-log">
+                                            <input type="hidden" name="{{csrf.keys.name}}" value="{{csrf.name}}">
+                                            <input type="hidden" name="{{csrf.keys.value}}" value="{{csrf.value}}">
+                                            <input type="submit" value="Prune docker system" data-confirm="Run docker system prune -a? This will remove unused images, containers, networks, volumes and build cache. It will not delete data of running containers. Continue?" />
+                                        </form>
+                                    </details>
+                                {% endif %}
                                 <h2>AIO passphrase change</h2>
                                 <details>
                                     <summary>Click here to change your AIO passphrase</summary>
--- a/php/public/index.php
+++ b/php/public/index.php
@@ -70,6 +70,7 @@ $app->post('/api/docker/backup-check-repair', AIO\Controller\DockerController::c
 $app->post('/api/docker/backup-test', AIO\Controller\DockerController::class . ':StartBackupContainerTest');
 $app->post('/api/docker/restore', AIO\Controller\DockerController::class . ':StartBackupContainerRestore');
 $app->post('/api/docker/stop', AIO\Controller\DockerController::class . ':StopContainer');
+$app->post('/api/docker/prune', AIO\Controller\DockerController::class . ':SystemPrune');
 $app->get('/api/docker/logs', AIO\Controller\DockerController::class . ':GetLogs');
 $app->post('/api/auth/login', AIO\Controller\LoginController::class . ':TryLogin');
 $app->get('/api/auth/getlogin', AIO\Controller\LoginController::class . ':GetTryLogin');
--- a/php/src/Controller/DockerController.php
+++ b/php/src/Controller/DockerController.php
@@ -328,6 +328,22 @@ readonly class DockerController {
         return $nonbufResp;
     }
 
+    public function SystemPrune(Request $request, Response $response, array $args) : Response {
+        // Get streaming response start and closure
+        $nonbufResp = $this->startStreamingResponse($response);
+
+        $body = $nonbufResp->getBody();
+        $addToStreamingResponseBody = function (string $message) use ($body) : void {
+            $body->write("<div>$message</div>");
+        };
+
+        $this->dockerActionManager->SystemPrune($addToStreamingResponseBody);
+
+        // End streaming response
+        $this->finalizeStreamingResponse($nonbufResp);
+        return $nonbufResp;
+    }
+
     public function stopTopContainer() : void {
         $id = self::TOP_CONTAINER;
         $this->PerformRecursiveContainerStop($id);


php/src/Docker/DockerActionManager.php
```diff
@@ -983,4 +983,36 @@ public function GetLatestDigestOfTag(string $imageName, string $tag): ?string {
...
+            // Special-case images prune to include the dangling filter as requested
+            if ($endpoint === 'images/prune') {
+                $filters = json_encode(['dangling' => ['false']]);
+                $url = $this->BuildApiUrl($endpoint . '?filters=' . urlencode($filters));
+            } else {
+                $url = $this->BuildApiUrl($endpoint);
+            }
+
+            try {
+                $this->guzzleClient->post($url);
```
szaimen: Should show the current step in the aio interface.

php/templates/containers.twig
```diff
@@ -322,6 +322,11 @@
                                 <input type="hidden" name="{{csrf.keys.value}}" value="{{csrf.value}}">
                                 <input type="submit" value="Stop containers" />
                             </form>
+                            <form method="POST" action="api/docker/prune" class="xhr">
+                                <input type="hidden" name="{{csrf.keys.name}}" value="{{csrf.name}}">
+                                <input type="hidden" name="{{csrf.keys.value}}" value="{{csrf.value}}">
+                                <input type="submit" value="Run docker system prune" onclick="return confirm('Run docker system prune? This will remove unused images, containers and volumes. Continue?')" />
+                            </form>
```
szaimen: probably move this somewhere else and use the loader to show the current process

php/src/Docker/DockerActionManager.php
```diff
@@ -983,4 +983,50 @@ public function GetLatestDigestOfTag(string $imageName, string $tag): ?string {
...
+            }
+
+            if ($addToStreamingResponseBody !== null) {
+                $addToStreamingResponseBody("Running $endpoint...");
+            }
+
+            try {
+                $resp = $this->guzzleClient->post($url);
+                if ($addToStreamingResponseBody !== null) {
+                    $addToStreamingResponseBody((string)$resp->getBody());
```
```suggestion
                $response = $this->guzzleClient->post($url);
                if ($addToStreamingResponseBody !== null) {
                    $addToStreamingResponseBody((string)$response ->getBody());
```

php/templates/containers.twig
```diff
@@ -370,6 +370,13 @@
                                     </form>
                                 {% endif %}
                             {% endif %}
+                            {% if was_start_button_clicked == true %}
+                                <form method="POST" action="api/docker/prune" target="overlay-log">
+                                    <input type="hidden" name="{{csrf.keys.name}}" value="{{csrf.name}}">
+                                    <input type="hidden" name="{{csrf.keys.value}}" value="{{csrf.value}}">
+                                    <input type="submit" value="Run docker system prune" onclick="return confirm('Run docker system prune? This will remove unused images, containers and volumes. Continue?')" />
+                                </form>
+                            {% endif %}
```
szaimen: still need to check the best position of this

php/src/Docker/DockerActionManager.php
```diff
@@ -983,4 +983,50 @@ public function GetLatestDigestOfTag(string $imageName, string $tag): ?string {
...
+            'networks/prune',
+            // Prune build cache
+            'build/prune',
+        ];
+
+        foreach ($endpoints as $endpoint) {
+            // Special-case images prune to include the dangling filter as requested
+            if ($endpoint === 'images/prune') {
+                $filters = json_encode(['dangling' => ['false']]);
+                $url = $this->BuildApiUrl($endpoint . '?filters=' . urlencode($filters));
```
szaimen: todo: fix this: `Error: src/Docker/DockerActionManager.php:1005:79: PossiblyFalseArgument: Argument 1 of urlencode cannot be false, possibly string value expected (see https://psalm.dev/104)`

php/templates/containers.twig
```diff
@@ -582,6 +582,17 @@
...
+                                {% if isAnyRunning == true %}
+                                    <h2>Docker System Prune</h2>
+                                    <details>
+                                        <summary>Click here to reveal a button to trigger docker system prune on your docker daemon host. This will remove unused images, containers and volumes.</summary>
+                                        <form method="POST" action="api/docker/prune" target="overlay-log">
+                                            <input type="hidden" name="{{csrf.keys.name}}" value="{{csrf.name}}">
+                                            <input type="hidden" name="{{csrf.keys.value}}" value="{{csrf.value}}">
+                                            <input type="submit" value="Run docker system prune" onclick="return confirm('Run docker system prune? This will remove unused images, containers and volumes. Continue?')" />
+                                        </details>
+                                    </form>
```
ernolf: The `</details>` and `</form>` closing tags are swapped — the form is opened inside `<details>` but closed after it. This is invalid HTML.
```suggestion
                                        </form>
                                    </details>
```

php/src/Docker/DockerActionManager.php
```diff
@@ -983,4 +983,50 @@ public function GetLatestDigestOfTag(string $imageName, string $tag): ?string {
...
+            }
+
+            if ($addToStreamingResponseBody !== null) {
+                $addToStreamingResponseBody("Running $endpoint...");
+            }
+
+            try {
+                $response = $this->guzzleClient->post($url);
+                if ($addToStreamingResponseBody !== null) {
+                    $addToStreamingResponseBody((string)$response->getBody());
```
ernolf: It outputs pure json, not verry usefull for Users.
Maybe something like this?
```suggestion
                    $data = json_decode((string)$response->getBody(), true);
                    $reclaimed = $data['SpaceReclaimed'] ?? 0;
                    if ($reclaimed > 0) {
                        $i = (int)floor(log($reclaimed, 1024));
                        $addToStreamingResponseBody(round($reclaimed / (1024 ** $i), 2) . ' ' . ['B','KB','MB','GB'][$i] . ' reclaimed.');
                    } else {
                        $addToStreamingResponseBody('Nothing to prune.');
                    }
```

ernolf (commented):
Tested the feature, overall it works. A few things I noticed:
1. HTML bug in containers.twig (blocking)
The `</form>` and `</details>` closing tags are swapped — the form is opened inside `<details>` but closed after it. This is invalid HTML. See my inline suggestion.
2. Page reloads immediately after prune completes
In forms.js, all target="overlay-log" forms trigger a location.reload(true) as soon as the iframe finishes loading. This makes sense for actions like "Stop containers" where the UI state changes, but for the prune action there's no reason to reload — it just causes the overlay to disappear before the user can read the output. The prune form should be excluded from this auto-reload behavior.
3. Raw JSON output
The Docker API responses are written directly into the overlay as raw JSON, e.g.:
{"ContainersDeleted":["b2773fb64d35..."],"SpaceReclaimed":0}
This isn't very user-friendly. It would be nice to parse the response in DockerActionManager.php and show something human-readable instead, e.g. images/prune done. Space reclaimed: 5.47 GB or Nothing to prune. — the SpaceReclaimed field from each endpoint response would be enough.
Due to issue # 2 (the auto-reload) I wasn't even able to properly verify the JSON output during testing since the overlay closes too quickly. So there's some work left on the UX side of this feature — fixing the reload behavior first would make it possible to then also iterate on the output formatting.

php/templates/containers.twig
szaimen: good catch, thanks! :)

php/src/Docker/DockerActionManager.php
szaimen: Thanks but I've chosen an alternative approach :)

php/templates/containers.twig
```diff
@@ -582,6 +582,17 @@
...
                         {% if is_backup_container_running == false %}
                             {% if isApacheStarting == false %}
+                                {% if isAnyRunning == true %}
+                                    <h2>Docker System Prune</h2>
+                                    <details>
+                                        <summary>Click here to reveal a button to trigger docker system prune on your docker daemon host. This will remove unused images, containers and volumes.</summary>
+                                        <form method="POST" action="api/docker/prune" target="overlay-log">
+                                            <input type="hidden" name="{{csrf.keys.name}}" value="{{csrf.name}}">
+                                            <input type="hidden" name="{{csrf.keys.value}}" value="{{csrf.value}}">
+                                            <input type="submit" value="Run docker system prune" onclick="return confirm('Run docker system prune? This will remove unused images, containers and volumes. Continue?')" />
```
szaimen:
```suggestion
                                        <summary>Click here to reveal a button to trigger docker system prune -a on your docker daemon host. This will remove unused images, containers, networks volumes and build cache.</summary>
                                        <form method="POST" action="api/docker/prune" target="overlay-log">
                                            <input type="hidden" name="{{csrf.keys.name}}" value="{{csrf.name}}">
                                            <input type="hidden" name="{{csrf.keys.value}}" value="{{csrf.value}}">
                                            <input type="submit" value="Run docker system prune" onclick="return confirm('Run docker system prune -a? This will remove unused images, containers, networks volumes and build cache. Continue?')" />
```

szaimen:
Thanks for the review and testing @ernolf ☺️

Please have another look :)

php/templates/containers.twig
```diff
@@ -582,6 +582,17 @@
...
                         {% if is_backup_container_running == false %}
                             {% if isApacheStarting == false %}
+                                {% if isAnyRunning == true %}
+                                    <h2>Docker System Prune</h2>
+                                    <details>
+                                        <summary>Click here to reveal a button to trigger "docker system prune -a" on your docker daemon host. This will remove unused images, containers, networks, volumes and build cache.</summary>
+                                        <form method="POST" action="api/docker/prune" target="overlay-log">
+                                            <input type="hidden" name="{{csrf.keys.name}}" value="{{csrf.name}}">
+                                            <input type="hidden" name="{{csrf.keys.value}}" value="{{csrf.value}}">
+                                            <input type="submit" value="Run docker system prune" onclick="return confirm('Run docker system prune -a? This will remove unused images, containers, networks, volumes and build cache. Continue?')" />
```
pabzm: I would like to see a little less technical text here. Something like:

```suggestion
                                        <summary>Click here to reveal a button to prune the docker system.</summary>
                                        <p>By clicking the button below you can run "docker system prune -a". This will remove unused images, containers, networks, volumes and build cache. It will not delete data of running containers.</p>
                                        <form method="POST" action="api/docker/prune" target="overlay-log">
                                            <input type="hidden" name="{{csrf.keys.name}}" value="{{csrf.name}}">
                                            <input type="hidden" name="{{csrf.keys.value}}" value="{{csrf.value}}">
                                            <input type="submit" value="Prune docker system" onclick="return confirm('Run "docker system prune -a"? This will remove unused images, containers, networks, volumes and build cache. It will not delete data of running containers. Continue?')" />
```

php/src/Docker/DockerActionManager.php
```diff
@@ -983,4 +983,61 @@ public function GetLatestDigestOfTag(string $imageName, string $tag): ?string {
...
+
+        if ($addToStreamingResponseBody !== null) {
+            $addToStreamingResponseBody("Docker system prune completed.");
+            sleep(1);
+
+            // We automatically reload after 10s so that the output can be read or copied if necessary
+            $addToStreamingResponseBody("Automatically reloading the page after 10s.");
+            for ($seconds = 10; $seconds >= 1; $seconds--) {
+                sleep(1);
+                $addToStreamingResponseBody((string)$seconds);
```
pabzm: Adding output makes the iframe scroll to the bottom, which makes it impossible to read the earlier output.

The simple fix would be to wait silently, without count down.

pabzm: Or alternatively we could avoid the automatic reloading and use a button that closes the overlay.

szaimen: Makes sense. Lets remove that then :)

php/src/Docker/DockerActionManager.php
```diff
@@ -983,4 +983,61 @@ public function GetLatestDigestOfTag(string $imageName, string $tag): ?string {
...
+            }
```
```suggestion
            sleep(10);
```

php/src/Docker/DockerActionManager.php
```diff
@@ -983,4 +983,61 @@ public function GetLatestDigestOfTag(string $imageName, string $tag): ?string {
...
+                $addToStreamingResponseBody("Running $endpoint...");
+            }
+
+            try {
+                $response = $this->guzzleClient->post($url);
+                if ($addToStreamingResponseBody !== null) {
+                    // Re-format the json and make it pretty for better UX
+                    $prettyJson = (string)json_encode(json_decode((string)$response->getBody()), JSON_PRETTY_PRINT);
+                    // Wrap the json in a pre html tag so that the formatting is kept
+                    $addToStreamingResponseBody("<pre>" . $prettyJson . "</pre>");
```
pabzm: With this the data is a little more humanly readable, but it's not easily understandable. E.g. the amount of regained disk space is shown as integer, without unit.

And the data still scrolls vertically, which I find ugly.

I don't have an easy suggestion to offer, sorry.

szaimen: I would suggest to improve this detail in a follow-up PR :)

ernolf (changes requested):
I tested the latest version thoroughly (video below). The functionality works correctly: named volumes are protected (explicitly testet!), anonymous volumes / stopped containers / unused images / networks / build cache are all cleaned up as expected.

That said, I'd like to raise the output UX one more time, because I think the current state — while an improvement over the raw JSON — still ships with two problems that compound each other:

**1. `JSON_PRETTY_PRINT` makes the output longer, not more readable**

On a real run with ~5.5 GB reclaimed, `images/prune` alone returns 30–50 deleted/untagged entries. With `JSON_PRETTY_PRINT`, each SHA256 hash gets its own indented line — that's 150–300 lines of output that no user will ever need to act on.

**2. The countdown actively prevents reading the output it was meant to protect**

Each countdown number is a new `<div>`, which triggers the iframe's auto-scroll. Any attempt to scroll up to read the prune results causes the next number to scroll back down. The countdown was intended to give users time to read — in practice it achieves the opposite.

---

I believe both issues can be addressed cleanly with a small, self-contained change. I've tested this locally and the result is:

https://github.com/user-attachments/assets/2d2d0ee6-06f1-4fe7-a21f-04d444378507

- Human-readable per-endpoint summary while the prune runs (live-scrolling)
- "Docker system prune completed." at the end
- An OK button that lets the user reload when they're ready

The complete diff across the four touched files:

```diff
diff --git a/php/src/Docker/DockerActionManager.php b/php/src/Docker/DockerActionManager.php
--- a/php/src/Docker/DockerActionManager.php
+++ b/php/src/Docker/DockerActionManager.php
@@ -1014,10 +1014,24 @@
             try {
                 $response = $this->guzzleClient->post($url);
                 if ($addToStreamingResponseBody !== null) {
-                    // Re-format the json and make it pretty for better UX
-                    $prettyJson = (string)json_encode(json_decode((string)$response->getBody()), JSON_PRETTY_PRINT);
-                    // Wrap the json in a pre html tag so that the formatting is kept
-                    $addToStreamingResponseBody("<pre>" . $prettyJson . "</pre>");
+                    $data = json_decode((string)$response->getBody(), true);
+                    $deleted = 0;
+                    foreach (['ContainersDeleted', 'ImagesDeleted', 'VolumesDeleted', 'NetworksDeleted', 'CachesDeleted'] as $key) {
+                        if (isset($data[$key]) && is_array($data[$key])) {
+                            $deleted += count($data[$key]);
+                        }
+                    }
+                    $reclaimed = $data['SpaceReclaimed'] ?? 0;
+                    $parts = [];
+                    if ($deleted > 0) {
+                        $parts[] = "$deleted item(s) deleted";
+                    }
+                    if ($reclaimed > 0) {
+                        $i = (int)floor(log($reclaimed, 1024));
+                        $parts[] = 'Space reclaimed: ' . round($reclaimed / (1024 ** $i), 2) . ' ' . ['B','KB','MB','GB'][$i];
+                    }
+                    $addToStreamingResponseBody(!empty($parts) ? implode('. ', $parts) . '.' : 'Nothing to prune.');
                 }
             } catch (RequestException $e) {
@@ -1029,14 +1049,7 @@
         }
 
         if ($addToStreamingResponseBody !== null) {
-            $addToStreamingResponseBody("Docker system prune completed.");
-            sleep(1);
-
-            // We automatically reload after 10s so that the output can be read or copied if necessary
-            $addToStreamingResponseBody("Automatically reloading the page after 10s.");
-            for ($seconds = 10; $seconds >= 1; $seconds--) {
-                sleep(1);
-                $addToStreamingResponseBody((string)$seconds);
-            }
+            $addToStreamingResponseBody("Docker system prune completed.");
         }
     }
 }

diff --git a/php/src/Controller/DockerController.php b/php/src/Controller/DockerController.php
--- a/php/src/Controller/DockerController.php
+++ b/php/src/Controller/DockerController.php
@@ -339,6 +339,8 @@
 
         $this->dockerActionManager->SystemPrune($addToStreamingResponseBody);
 
+        $body->write('<div><br><button onclick="window.parent.location.reload()">OK</button></div>');
+
         // End streaming response
         $this->finalizeStreamingResponse($nonbufResp);
         return $nonbufResp;

diff --git a/php/public/forms.js b/php/public/forms.js
--- a/php/public/forms.js
+++ b/php/public/forms.js
@@ -82,10 +82,24 @@
         form.onsubmit = function() {
             enableSpinner();
             document.getElementById('overlay-log')?.classList.add('visible');
-            // Reload the page after the response was fully loaded into the iframe.
-            document.querySelector('iframe[name="overlay-log"]').addEventListener('load', () => {
-                location.reload(true);
-            });
+            const iframe = document.querySelector('iframe[name="overlay-log"]');
+            if (!form.classList.contains('no-reload-on-complete')) {
+                // Reload the page after the response was fully loaded into the iframe.
+                iframe.addEventListener('load', () => {
+                    location.reload(true);
+                });
+            } else {
+                // Auto-scroll iframe content while streaming; stop once response is complete.
+                const scrollInterval = setInterval(() => {
+                    try {
+                        const doc = iframe.contentDocument || iframe.contentWindow?.document;
+                        if (doc) {
+                            doc.documentElement.scrollTop = doc.documentElement.scrollHeight;
+                        }
+                    } catch (e) {}
+                }, 200);
+                iframe.addEventListener('load', () => clearInterval(scrollInterval), { once: true });
+            }
         };
     }
   }

diff --git a/php/templates/containers.twig b/php/templates/containers.twig
--- a/php/templates/containers.twig
+++ b/php/templates/containers.twig
@@ -589,7 +589,7 @@
-                                        <form method="POST" action="api/docker/prune" target="overlay-log">
+                                        <form method="POST" action="api/docker/prune" target="overlay-log" class="no-reload-on-complete">
```

**What this does, file by file:**

- **`DockerActionManager.php`** — replaces the JSON dump with a per-endpoint human-readable summary ("X item(s) deleted. Space reclaimed: Y GB." / "Nothing to prune."); removes the countdown and its `sleep()` calls
- **`DockerController.php`** — appends an OK button at the end of the streaming response; clicking it triggers `window.parent.location.reload()` from within the iframe
- **`forms.js`** — introduces a `no-reload-on-complete` opt-out: forms with this class get a live auto-scroll instead of an automatic page reload on completion; all other forms behave exactly as before
- **`containers.twig`** — adds `class="no-reload-on-complete"` to the prune form, wiring it up to the new forms.js behaviour; without this the other three changes have no effect

The `no-reload-on-complete` mechanism is generic and could be reused for other long-running operations in the future where reading the output matters. But that's entirely your call — if you'd prefer to keep it prune-specific, that's fine too.

**On the "follow-up PR" question:** I understand the intention, and I'm not asking for a polished, feature-rich output view — that can absolutely come later. What I'm asking for is the minimum that makes this feature *usable on day one*. The diff above is 4 files, ~35 lines net. It's not a redesign. Without it, we're shipping a button that produces output nobody can read and that scrolls itself out of view before the user gets a chance to try. A follow-up PR that improves something working is a great plan; a follow-up PR that fixes something broken is avoidable friction. I'd feel much more comfortable approving once this baseline is in — and I genuinely think it'll save us both time in the community forum.

Happy to adjust anything here. And again, thank you for building this — it's a genuinely useful feature.

szaimen:
Thanks for the suggestions @ernolf! I'll look into them :)

szaimen:
Hi @ernolf can you please review again? I just rebased and squashed all fixes into one commit and applied some of your suggestions :)

ernolf (approved):
Re-tested with commit f25f588. The parsed output works great — clean, readable, and the 10s window is enough to review the results. Approved.

szaimen:
Thanks a lot for your help on this @ernolf 😊

szaimen:

PR #7684, opened 2026-03-02
feat(talk): trust custom CA certificates via NEXTCLOUD_TRUSTED_CACERTS_DIR

michnovka:
## Summary

When `NEXTCLOUD_TRUSTED_CACERTS_DIR` is configured on the mastercontainer, the Nextcloud container already receives the custom CA certificates. However, the Talk signaling container does not — causing the signaling server to fail TLS verification when connecting back to Nextcloud instances that use private/internal CA certificates (e.g. for internal domains with a corporate CA).

This means users with custom CAs currently have to either:
- Manually patch `skipverify = true` in the signaling config after every container recreation
- Accept broken Talk functionality

This PR mounts the same CA certificate directory into the Talk container and extends the startup script to build a custom CA bundle that includes the additional certificates.

## Changes

- **`php/containers.json`**: Added a `volumes` entry to mount `%NEXTCLOUD_TRUSTED_CACERTS_DIR%` into `/usr/local/share/ca-certificates` (read-only) on the Talk container, matching the existing Nextcloud container behavior. The mount is skipped when the variable is empty.

- **`Containers/talk/start.sh`**: Added CA certificate handling before the signaling config is generated. Since the Talk container uses a read-only root filesystem, `update-ca-certificates` cannot be used. Instead, the script copies the system CA bundle to `/tmp` (tmpfs), appends any custom `.crt` files, and sets `SSL_CERT_FILE` so Go's TLS stack uses the extended bundle.

## How it works

1. AIO mastercontainer mounts the host CA directory into the Talk container at `/usr/local/share/ca-certificates/` (read-only)
2. On startup, `start.sh` detects `.crt` files in that directory
3. Copies `/etc/ssl/certs/ca-certificates.crt` to `/tmp/ca-certificates.crt`
4. Appends each custom certificate
5. Sets `SSL_CERT_FILE=/tmp/ca-certificates.crt`
6. The Go-based signaling server picks up the env var and verifies TLS using the extended bundle

No changes to the Dockerfile are needed — `ca-certificates` is already installed and `SKIP_CERT_VERIFY=false` remains the correct default.

## Testing

Tested with a Nextcloud AIO instance behind a reverse proxy using a private CA on an internal domain. Before this change, the signaling server logged `certificate signed by unknown authority` errors and Talk room creation returned HTTP 500. After this change, TLS verification succeeds without `skipverify = true`.

### DIFF
--- a/Containers/talk/start.sh
+++ b/Containers/talk/start.sh
@@ -28,7 +28,6 @@ if mountpoint -q /usr/local/share/ca-certificates; then
     for cert in /usr/local/share/ca-certificates/*; do
         if [ -f "$cert" ]; then
             cat "$cert" >> /tmp/ca-certificates.crt
-            echo "  Added: $(basename "$cert")"
         fi
     done
     export SSL_CERT_FILE=/tmp/ca-certificates.crt


michnovka:
@szaimen Im not sure I understand the failing check correctly... do I need to do anything, or its all good for you? its a very simple PR. Thanks!

szaimen:
Hi, first of all, thanks for your PR!

> @szaimen Im not sure I understand the failing check correctly... do I need to do anything, or its all good for you? its a very simple PR. Thanks!

You don't need to do anything. This is just blocking the merge for now as we currently have a beta release pending that will soon be a stable release. Afterwards merging is allowed for us again.

I'll review your PR going forward 👍

Containers/talk/start.sh
```diff
@@ -18,6 +18,21 @@ elif [ -z "$INTERNAL_SECRET" ]; then
...
 
+# Trust additional CA certificates, if the user provided NEXTCLOUD_TRUSTED_CACERTS_DIR
+# The container is read-only, so we build a custom bundle in /tmp (tmpfs) and
+# point Go's TLS stack to it via SSL_CERT_FILE.
+if [ -n "$(find /usr/local/share/ca-certificates -name '*.crt' -type f 2>/dev/null)" ]; then
+    echo "Trusting additional CA certificates..."
+    cp /etc/ssl/certs/ca-certificates.crt /tmp/ca-certificates.crt
+    for cert in /usr/local/share/ca-certificates/*.crt; do
+        if [ -f "$cert" ]; then
+            cat "$cert" >> /tmp/ca-certificates.crt
```
szaimen, line 30: Hi, can you please check how big this file is becoming in a default setup?

michnovka: its 217kB

szaimen: ok, sounds fine

Containers/talk/start.sh
```diff
@@ -18,6 +18,21 @@ elif [ -z "$INTERNAL_SECRET" ]; then
...
+            echo "  Added: $(basename "$cert")"
+        fi
+    done
+    export SSL_CERT_FILE=/tmp/ca-certificates.crt
```
szaimen: Some suggestion below. FYI: to use the mountpoint cmd, you need to add `util-linux-misc` into the container via apk.
```suggestion
if mountpoint -q /usr/local/share/ca-certificates; then
    echo "Trusting additional CA certificates..."
    set -x
    cp /etc/ssl/certs/ca-certificates.crt /tmp/ca-certificates.crt
    for cert in /usr/local/share/ca-certificates/*; do
        if [ -f "$cert" ]; then
            cat "$cert" >> /tmp/ca-certificates.crt
            echo "  Added: $(basename "$cert")"
        fi
    done
    export SSL_CERT_FILE=/tmp/ca-certificates.crt
    set +x
```

michnovka: Pushed a followup

Containers/talk/start.sh
```diff
@@ -18,6 +18,23 @@ elif [ -z "$INTERNAL_SECRET" ]; then
...
+# The container is read-only, so we build a custom bundle in /tmp (tmpfs) and
+# point Go's TLS stack to it via SSL_CERT_FILE.
+if mountpoint -q /usr/local/share/ca-certificates; then
+    echo "Trusting additional CA certificates..."
+    set -x
+    cp /etc/ssl/certs/ca-certificates.crt /tmp/ca-certificates.crt
+    for cert in /usr/local/share/ca-certificates/*; do
+        if [ -f "$cert" ]; then
+            cat "$cert" >> /tmp/ca-certificates.crt
+            echo "  Added: $(basename "$cert")"
```
szaimen: Maybe this can be removed now since we have a very verbose output anyways already?
```suggestion
```

szaimen:

PR #7689, opened 2026-03-04
aio-interface: improve session config and cookie store

Zoey2936:
- Close part of #7685

### DIFF
--- a/php/public/index.php
+++ b/php/public/index.php
@@ -38,14 +38,15 @@ $container->set(Guard::class, function () use ($responseFactory) {
 
 // Register Middleware To Be Executed On All Routes
 session_start([
-    "save_path" => $dataConst->GetSessionDirectory(), // where to save the session files
-    "gc_maxlifetime" => 86400, // delete sessions after 24 hours ... // https://www.php.net/manual/en/session.configuration.php#ini.session.gc-maxlifetime
-    "gc_probability" => 1, // ... to ... // https://www.php.net/manual/en/session.configuration.php#ini.session.gc-probability
+    "save_path" => $dataConst->GetSessionDirectory(), // Where to save the session files
+    "cookie_lifetime" => 0, // Delete the session cookie whenever the browser is closed. See https://www.php.net/manual/en/session.configuration.php#ini.session.cookie-lifetime
+    "gc_maxlifetime" => 86400, // Delete sessions after 24 hours. See https://www.php.net/manual/en/session.configuration.php#ini.session.gc-maxlifetime
+    "gc_probability" => 1, // Probability that the session cleanup starts. See https://www.php.net/manual/en/session.configuration.php#ini.session.gc-probability
     "gc_divisor" => 1, // gc_probability/gc_divisor = 1/1 = 100%, meaning that *all* outdated sessions get deleted when the cleanup job runs. See https://www.php.net/manual/en/session.configuration.php#ini.session.gc-divisor
-    "use_strict_mode" => true, // only allow initialized session IDs // https://www.php.net/manual/en/session.configuration.php#ini.session.use-strict-mode
-    "cookie_secure" => true, // only send cookies over https (not http) // https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Set-Cookie#secure
-    "cookie_httponly" => true, // block the cookie from being read with js in the browser, will still be send for fetch request triggered by js // https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Set-Cookie#httponly
-    "cookie_samesite" => "Strict", // only send the cookie with requests triggered by AIO itself // https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Set-Cookie#samesitesamesite-value
+    "use_strict_mode" => true, // Only allow initialized session IDs. See https://www.php.net/manual/en/session.configuration.php#ini.session.use-strict-mode
+    "cookie_secure" => true, // Only send cookies over https (not http). See https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Set-Cookie#secure
+    "cookie_httponly" => true, // Block the cookie from being read with js in the browser, will still be send for fetch request triggered by js. See https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Set-Cookie#httponly
+    "cookie_samesite" => "Strict", // Only send the cookie with requests triggered by AIO itself. See https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Set-Cookie#samesitesamesite-value
 ]);
 $app->add(Guard::class);
 
--- a/php/public/index.php
+++ b/php/public/index.php
@@ -41,7 +41,7 @@ session_start([
     "save_path" => $dataConst->GetSessionDirectory(), // where to save the session files
     "gc_maxlifetime" => 86400, // delete sessions after 24 hours ... // https://www.php.net/manual/en/session.configuration.php#ini.session.gc-maxlifetime
     "gc_probability" => 1, // ... to ... // https://www.php.net/manual/en/session.configuration.php#ini.session.gc-probability
-    "gc_divisor" => 1, // 100% (gc_probability/gc_divisor = 1/1 = 100%)  // https://www.php.net/manual/en/session.configuration.php#ini.session.gc-divisor
+    "gc_divisor" => 1, // gc_probability/gc_divisor = 1/1 = 100%, meaning that *all* outdated sessions get deleted when the cleanup job runs. See https://www.php.net/manual/en/session.configuration.php#ini.session.gc-divisor
     "use_strict_mode" => true, // only allow initialized session IDs // https://www.php.net/manual/en/session.configuration.php#ini.session.use-strict-mode
     "cookie_secure" => true, // only send cookies over https (not http) // https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Set-Cookie#secure
     "cookie_httponly" => true, // block the cookie from being read with js in the browser, will still be send for fetch request triggered by js // https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Set-Cookie#httponly
--- a/php/public/index.php
+++ b/php/public/index.php
@@ -23,13 +23,6 @@ require __DIR__ . '/../vendor/autoload.php';
 
 $container = \AIO\DependencyInjection::GetContainer();
 $dataConst = $container->get(\AIO\Data\DataConst::class);
-ini_set('session.save_path', $dataConst->GetSessionDirectory());
-
-// Auto logout on browser close
-ini_set('session.cookie_lifetime', '0');
-
-# Keep session for 24h max
-ini_set('session.gc_maxlifetime', '86400');
 
 // Create app
 AppFactory::setContainer($container);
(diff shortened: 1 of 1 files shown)


php/public/index.php
```diff
@@ -23,13 +23,6 @@
 
 $container = \AIO\DependencyInjection::GetContainer();
 $dataConst = $container->get(\AIO\Data\DataConst::class);
-ini_set('session.save_path', $dataConst->GetSessionDirectory());
-
-// Auto logout on browser close
-ini_set('session.cookie_lifetime', '0');
```
Zoey2936, line 29: I removed this since it is the default

php/public/index.php
```diff
@@ -23,13 +23,6 @@
 
 $container = \AIO\DependencyInjection::GetContainer();
 $dataConst = $container->get(\AIO\Data\DataConst::class);
-ini_set('session.save_path', $dataConst->GetSessionDirectory());
```
Zoey2936, line 26: moved to session_start

php/public/index.php
```diff
@@ -23,13 +23,6 @@
...
-
-// Auto logout on browser close
-ini_set('session.cookie_lifetime', '0');
-
-# Keep session for 24h max
-ini_set('session.gc_maxlifetime', '86400');
```
Zoey2936, line 32: moved to session_start

php/public/index.php
```diff
@@ -44,7 +37,16 @@
...
-session_start();
+session_start([
+    "save_path" => $dataConst->GetSessionDirectory(),
+    "gc_maxlifetime" => 86400,
+    "gc_probability" => 1,
+    "gc_divisor" => 1,
+    "use_strict_mode" => true,
+    "cookie_secure" => true,
+    "cookie_httponly" => true,
+    "cookie_samesite" => "Strict",
```
Zoey2936: I'm not sure if Strict or Lax should be used. Lax works in any case. But strict should be perefferd, see here: https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Set-Cookie#samesitesamesite-value

php/public/index.php
```diff
@@ -44,7 +37,16 @@
...
 // Register Middleware To Be Executed On All Routes
-session_start();
+session_start([
+    "save_path" => $dataConst->GetSessionDirectory(),
+    "gc_maxlifetime" => 86400,
+    "gc_probability" => 1,
+    "gc_divisor" => 1,
+    "use_strict_mode" => true,
+    "cookie_secure" => true,
+    "cookie_httponly" => true,
```
Zoey2936: since the browser/frontend never access the token with javascript this will make the token to be only send with https requests, while blocking javascript from reading it

php/public/index.php
```diff
@@ -44,7 +37,16 @@
...
 
 // Register Middleware To Be Executed On All Routes
-session_start();
+session_start([
+    "save_path" => $dataConst->GetSessionDirectory(),
+    "gc_maxlifetime" => 86400,
+    "gc_probability" => 1,
+    "gc_divisor" => 1,
+    "use_strict_mode" => true,
+    "cookie_secure" => true,
```
Zoey2936: this makes the cookie be only send over https, but not http

php/public/index.php
```diff
@@ -44,7 +37,16 @@
 });
 
 // Register Middleware To Be Executed On All Routes
-session_start();
+session_start([
+    "save_path" => $dataConst->GetSessionDirectory(),
+    "gc_maxlifetime" => 86400,
+    "gc_probability" => 1,
+    "gc_divisor" => 1,
+    "use_strict_mode" => true,
```
Zoey2936: this is recommended by php. https://www.php.net/manual/en/session.configuration.php#ini.session.use-strict-mode

php/public/index.php
```diff
@@ -44,7 +37,16 @@
 });
 
 // Register Middleware To Be Executed On All Routes
-session_start();
+session_start([
+    "save_path" => $dataConst->GetSessionDirectory(),
+    "gc_maxlifetime" => 86400,
+    "gc_probability" => 1,
+    "gc_divisor" => 1,
```
Zoey2936: In the past only gc_maxlifetime was set. It controls after which time a session may be deleted. The comment said "Keep session for 24h max", but this is wrong. If only gc_maxlifetime is changed but not gc_probability/gc_divisor it only means that a session can be deleted after 24 hours with a chance of 1%. Setting gc_probability/gc_divisor both to 1 means that the session will be deleted to 100% after 24 hours

https://www.php.net/manual/en/session.configuration.php#ini.session.gc-probability

php/public/index.php
pabzm: `Strict` should work, so let's use that, please. I would consider anything that breaks using `Strict` a bug.

php/public/index.php
pabzm: Since a short while we do `fetch()` some data from Javascript, so we need the cookie in JS, too, please.

pabzm (changes requested):
Thank you for this contribution, highly appreciated! ~~Please allow cookie access from Javascript, though, we need it to fetch log data.~~

php/public/index.php
Zoey2936: I think you have a small missunderstanding here, even with httponly set, the cookie will still be sent for fetch request. httponly, only means that the js cannot read the cookie, it will still be sent for all matching http(s) requests truggered by js

Zoey2936: https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Set-Cookie#httponly

pabzm: You are right, I wasn't concentrating. Sorry & thank you!

pabzm (approved):
Thank you very much! 👍

szaimen (commented):
Thanks a lot @Zoey2936! 

One suggestion: can you please add the comments that you added in the review into the codebase so that we now directly what it does? Thanks in advance! :)

Zoey2936:
> Thanks a lot @Zoey2936!
> 
> One suggestion: can you please add the comments that you added in the review into the codebase so that we now directly what it does? Thanks in advance! :)

done

php/public/index.php
```diff
@@ -44,7 +37,16 @@
 });
 
 // Register Middleware To Be Executed On All Routes
-session_start();
+session_start([
+    "save_path" => $dataConst->GetSessionDirectory(), // where to save the session files
+    "gc_maxlifetime" => 86400, // delete sessions after 24 hours ... // https://www.php.net/manual/en/session.configuration.php#ini.session.gc-maxlifetime
+    "gc_probability" => 1, // ... to ... // https://www.php.net/manual/en/session.configuration.php#ini.session.gc-probability
+    "gc_divisor" => 1, // 100% // https://www.php.net/manual/en/session.configuration.php#ini.session.gc-divisor
```
pabzm: Could you clarify these two comments a little bit, please? It would be helpful to mention their effect, not just their value. Just as you did in the comment to this PR. (Still linking to the docs is fine, I think.)

pabzm (changes requested):
This looks good and behaves as expected, thank you!

I'd just like to ask you to rephrase two comments to make them more helpful, please.

php/public/index.php
Zoey2936: done

php/public/index.php
```diff
@@ -44,7 +37,16 @@
 });
 
 // Register Middleware To Be Executed On All Routes
-session_start();
+session_start([
+    "save_path" => $dataConst->GetSessionDirectory(), // where to save the session files
+    "gc_maxlifetime" => 86400, // delete sessions after 24 hours ... // https://www.php.net/manual/en/session.configuration.php#ini.session.gc-maxlifetime
+    "gc_probability" => 1, // ... to ... // https://www.php.net/manual/en/session.configuration.php#ini.session.gc-probability
+    "gc_divisor" => 1, // 100% (gc_probability/gc_divisor = 1/1 = 100%)  // https://www.php.net/manual/en/session.configuration.php#ini.session.gc-divisor
```
pabzm: Let's please include a hint on what these 100% mean. E.g.:

```suggestion
    "gc_divisor" => 1, // gc_probability/gc_divisor = 1/1 = 100%, meaning that *all* outdated sessions get deleted when the cleanup job runs. See https://www.php.net/manual/en/session.configuration.php#ini.session.gc-divisor
```

pabzm (dismissed):
I added one suggestion to change the comment a little more, could you have a look at that, please?

php/public/index.php
szaimen: @Zoey2936 can you please address this comment? (by mentioning in every comment what the reason for the setting is). I'd say then is this PR ready to go :)

php/public/index.php
```diff
@@ -44,7 +37,16 @@
...
+    "gc_probability" => 1, // ... to ... // https://www.php.net/manual/en/session.configuration.php#ini.session.gc-probability
+    "gc_divisor" => 1, // gc_probability/gc_divisor = 1/1 = 100%, meaning that *all* outdated sessions get deleted when the cleanup job runs. See https://www.php.net/manual/en/session.configuration.php#ini.session.gc-divisor
+    "use_strict_mode" => true, // only allow initialized session IDs // https://www.php.net/manual/en/session.configuration.php#ini.session.use-strict-mode
+    "cookie_secure" => true, // only send cookies over https (not http) // https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Set-Cookie#secure
+    "cookie_httponly" => true, // block the cookie from being read with js in the browser, will still be send for fetch request triggered by js // https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Set-Cookie#httponly
+    "cookie_samesite" => "Strict", // only send the cookie with requests triggered by AIO itself // https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Set-Cookie#samesitesamesite-value
+]);
```
szaimen: Some late suggestions:

```suggestion
    "save_path" => $dataConst->GetSessionDirectory(), // Where to save the session files
    "cookie_lifetime" => 0, // Delete the session cookie whenever the browser is closed. See https://www.php.net/manual/en/session.configuration.php#ini.session.cookie-lifetime
    "gc_maxlifetime" => 86400, // Delete sessions after 24 hours. See https://www.php.net/manual/en/session.configuration.php#ini.session.gc-maxlifetime
    "gc_probability" => 1, // Probability that the session cleanup starts. See https://www.php.net/manual/en/session.configuration.php#ini.session.gc-probability
    "gc_divisor" => 1, // gc_probability/gc_divisor = 1/1 = 100%, meaning that *all* outdated sessions get deleted when the cleanup job runs. See https://www.php.net/manual/en/session.configuration.php#ini.session.gc-divisor
    "use_strict_mode" => true, // Only allow initialized session IDs. See https://www.php.net/manual/en/session.configuration.php#ini.session.use-strict-mode
    "cookie_secure" => true, // Only send cookies over https (not http). See https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Set-Cookie#secure
    "cookie_httponly" => true, // Block the cookie from being read with js in the browser, will still be send for fetch request triggered by js. See https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Set-Cookie#httponly
    "cookie_samesite" => "Strict", // Only send the cookie with requests triggered by AIO itself. See https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Set-Cookie#samesitesamesite-value
]);
```

Zoey2936: but cookie_lifetime defaults to 0?

szaimen: yes, I would still like to keep it for easier understanding and better maintainability in that case :)

szaimen (approved):
LGTM. Thanks a lot! ☺️

Zoey2936:
We should also adjust the cookie name: https://www.php.net/manual/en/session.configuration.php#ini.session.name
so we can set the cookie prefix to `__Host-Http-`: https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Set-Cookie#cookie_prefixes

szaimen:

Issue #5251, opened 2024-09-12
Feature Requests and others: Contributors Wanted!


szaimen:
This is a collection of open feature requests that need help by contributors.

## Classic Feature requests
- https://github.com/nextcloud/all-in-one/discussions/4223
- https://github.com/nextcloud/all-in-one/discussions/3889
- https://github.com/nextcloud/all-in-one/discussions/4480
- https://github.com/nextcloud/all-in-one/discussions/4751
- https://github.com/nextcloud/all-in-one/discussions/3576
- https://github.com/nextcloud/all-in-one/discussions/4501
- https://github.com/nextcloud/all-in-one/discussions/4422
- https://github.com/nextcloud/all-in-one/discussions/5095
- https://github.com/nextcloud/all-in-one/discussions/5012
- https://github.com/nextcloud/all-in-one/discussions/5147
- https://github.com/nextcloud/all-in-one/discussions/4277

## Add Community Containers (maintainers wanted)
- https://github.com/nextcloud/all-in-one/discussions/3549
- https://github.com/nextcloud/all-in-one/discussions/3061
- https://github.com/nextcloud/all-in-one/discussions/5239
- https://github.com/nextcloud/all-in-one/discussions/5150
- https://github.com/nextcloud/all-in-one/discussions/392
- https://github.com/nextcloud/all-in-one/discussions/3696
- https://github.com/nextcloud/all-in-one/discussions/4380
- https://github.com/nextcloud/all-in-one/discussions/4438
- https://github.com/nextcloud/all-in-one/discussions/3218
- https://github.com/nextcloud/all-in-one/discussions/4967

PR #7690, opened 2026-03-04
aio-interface: improve headers

Zoey2936:
- Close part of #7685


Containers/mastercontainer/acme.Caddyfile
```diff
@@ -23,10 +23,38 @@
...
 
 http://:80 {
+	header {
+		Content-Security-Policy "default-src 'self'; base-uri 'self'; worker-src 'none'; object-src 'none'; upgrade-insecure-requests;"
+        X-Content-Type-Options "nosniff"
+        X-Frame-Options "DENY"
+        X-Permitted-Cross-Domain-Policies "none"
+        X-DNS-Prefetch-Control "off"
+        Referrer-Policy "no-referrer"
+        X-Robots-Tag "noindex, nofollow"
```
Zoey2936: This tells crawlers like google to not index the page

Containers/mastercontainer/acme.Caddyfile
```diff
@@ -23,10 +23,38 @@
 }
 
 http://:80 {
+	header {
+		Content-Security-Policy "default-src 'self'; base-uri 'self'; worker-src 'none'; object-src 'none'; upgrade-insecure-requests;"
+        X-Content-Type-Options "nosniff"
+        X-Frame-Options "DENY"
+        X-Permitted-Cross-Domain-Policies "none"
+        X-DNS-Prefetch-Control "off"
+        Referrer-Policy "no-referrer"
```
Zoey2936: this means that browsers should never send referer headers when leaving the aio panel

Containers/mastercontainer/acme.Caddyfile
```diff
@@ -23,10 +23,38 @@
 }
 
 http://:80 {
+	header {
+		Content-Security-Policy "default-src 'self'; base-uri 'self'; worker-src 'none'; object-src 'none'; upgrade-insecure-requests;"
+        X-Content-Type-Options "nosniff"
+        X-Frame-Options "DENY"
+        X-Permitted-Cross-Domain-Policies "none"
+        X-DNS-Prefetch-Control "off"
```
Zoey2936: https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/X-DNS-Prefetch-Control
This header can also stay unset, but it is very little privacy improvement

Containers/mastercontainer/acme.Caddyfile
```diff
@@ -23,10 +23,38 @@
 }
 
 http://:80 {
+	header {
+		Content-Security-Policy "default-src 'self'; base-uri 'self'; worker-src 'none'; object-src 'none'; upgrade-insecure-requests;"
+        X-Content-Type-Options "nosniff"
+        X-Frame-Options "DENY"
+        X-Permitted-Cross-Domain-Policies "none"
```
Zoey2936: https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/X-Permitted-Cross-Domain-Policies AIO seems to not use flashplayer

Containers/mastercontainer/acme.Caddyfile
```diff
@@ -23,10 +23,38 @@
 }
 
 http://:80 {
+	header {
+		Content-Security-Policy "default-src 'self'; base-uri 'self'; worker-src 'none'; object-src 'none'; upgrade-insecure-requests;"
+        X-Content-Type-Options "nosniff"
+        X-Frame-Options "DENY"
```
Zoey2936: This forbids embedding the aio panel in an iframe

Containers/mastercontainer/acme.Caddyfile
```diff
@@ -23,10 +23,38 @@
 }
 
 http://:80 {
+	header {
+		Content-Security-Policy "default-src 'self'; base-uri 'self'; worker-src 'none'; object-src 'none'; upgrade-insecure-requests;"
+        X-Content-Type-Options "nosniff"
```
Zoey2936: https://developer.mozilla.org/de/docs/Web/HTTP/Reference/Headers/X-Content-Type-Options
AIO seems to always send a mime type header so we can forbid the browser to sniff the mime type

Containers/mastercontainer/acme.Caddyfile
```diff
@@ -23,10 +23,38 @@
 }
 
 http://:80 {
+	header {
+		Content-Security-Policy "default-src 'self'; base-uri 'self'; worker-src 'none'; object-src 'none'; upgrade-insecure-requests;"
```
Zoey2936: https://developer.mozilla.org/en-US/docs/Web/HTTP/Guides/CSP

Since AIO doesn't connect to any cross-origin domains or uses inline styles and scripts, etc. It seems like the strictest possible CSP can be used

Zoey2936: needs to revert, AIO uses inline script, so either unsafe-inline needs to be allowed, or hashes/nonces set

Zoey2936: https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Content-Security-Policy#directives

Also all other options needs to be checked if they can be further limited / need to be expanded

szaimen: > needs to revert, AIO uses inline script

Honestly we should remove inline scripts imho and migrate them to proper scripts...

Zoey2936: is there a good way to remove `onClick`for buttons?

szaimen: Yes migrate it to a proper js file with selector...

Zoey2936: maybe the csp header should be moved to a different which also changes inline scripts?

asavageiv:
It might be nice to include comments with the rationale for each header in the code.

Containers/mastercontainer/acme.Caddyfile
pabzm: I really like this PR and would contribute moving the onclick-handlers into standalone javascript files.

Should I push into the branch or file a new PR against this PR?

szaimen:
This PR needs a rebase

Containers/mastercontainer/acme.Caddyfile
Zoey2936: for me it is fine if you push directly

pabzm: Now a strict CSP should work

Containers/mastercontainer/Dockerfile
```diff
@@ -4,6 +4,8 @@ FROM docker:29.3.0-cli AS docker
 
 ARG CADDY_REMOTE_HOST_HASH=b21775afa730ffb52a24ddff310c8a6d1fd37276
 
+ARG CADDY_REMOTE_HOST_HASH=b21775afa730ffb52a24ddff310c8a6d1fd37276
```
pabzm, line 5: Why repeat this?

pabzm (dismissed):
Looks already good to me!

Just the one repetition confuses me. If there's any reason to it, please explain it.

Thank you very much!

Containers/mastercontainer/Dockerfile
Zoey2936: I think this occurred while rebasing and can be removed

Containers/mastercontainer/acme.Caddyfile
Zoey2936: I want to later check again if it maybe can be made more strict

szaimen:
Hey @Zoey2936 first of all, thanks a lot for the PR! ☺️

I was wondering can we maybe set the headers via php? Then it should hopefully set the headers for all endpoints without us having to duplicate them in every Caddyfile and different endpoints and dev server, no?

Zoey2936:
It should also work by setting the headers in php

Zoey2936:
I want to work further on this, but I'm confused because of #7820, should I push my changes there?

szaimen:
> I want to work further on this, 

Thank you very much! :)

> but I'm confused because of #7820, should I push my changes there?

I'd say pushing your changes to https://github.com/nextcloud/all-in-one/pull/7820 directly makes most sense. Do you agree? :)

Zoey2936:
> It should also work by setting the headers in php

I need to revert my words, since then the header will not be sent for other files, which are never touched by php

Zoey2936:
I've tried to work on a more stirct CSP, but this could be improved in the source files:
```
Content-Security-Policy: Die Einstellungen der Seite haben die Ausführung eines Ereignis-Handlers (script-src-attr) blockiert, weil er gegen folgende Direktive verstößt: "default-src 'none'". Ziehen Sie die Verwendung eines Hashs ('sha256-/9h5/n5yL7jhlwt387Bkaq9iLJ9eWZ53JtGvRLcqIOI=') zusammen mit 'unsafe-hashes' in Erwägung.
Source: showPassword('current-master-password')
Content-Security-Policy: Die Einstellungen der Seite haben die Ausführung eines Ereignis-Handlers (script-src-attr) blockiert, weil er gegen folgende Direktive verstößt: "default-src 'none'". Ziehen Sie die Verwendung eines Hashs ('sha256-5A3XM6HcvFhh9xP7kCNZfN0CGPwX97AvizFmXhdfAzc=') zusammen mit 'unsafe-hashes' in Erwägung.
Source: showPassword('new-master-password')
Content-Security-Policy: Die Einstellungen der Seite haben die Ausführung eines Inline-Skripts (script-src-elem) blockiert, da es gegen folgende Direktive verstößt: "script-src-elem 'self'". Ziehen Sie die Verwendung eines Hashs ('sha256-6bNyZN1sRD0pH97KGhXK/G/B9DKHZKxOiD7jfyz+EQ4=') oder einer Nonce in Erwägung. (in stop/start)
Content-Security-Policy: Die Einstellungen der Seite haben die Anwendung eines Inline-Styles (style-src-attr) blockiert, da er gegen folgende Direktive verstößt: "default-src 'none'". Ziehen Sie die Verwendung eines Hashs ('sha256-BmMpY1aLoCxBMFnwjzq/SVkHNNPpa2IUx4R/PPxgwV8=', benötigt 'unsafe-hashes' für Stil-Attribute) oder einer Nonce in Erwägung.
Source: vertical-align: middle; margin-left: 4px… 2 (in containers)
Content-Security-Policy: Die Einstellungen der Seite haben die Anwendung eines Inline-Styles (style-src-attr) blockiert, da er gegen folgende Direktive verstößt: "default-src 'none'". Ziehen Sie die Verwendung eines Hashs ('sha256-Vc6YdsPwJmyOyOWY3K0+DEFqLyvNxfsrADVC0UvQkxI=', benötigt 'unsafe-hashes' für Stil-Attribute) oder einer Nonce in Erwägung.
Source: vertical-align: middle; margin-right: 6p… (in containers)
Content-Security-Policy: Die Einstellungen der Seite haben die Anwendung eines Inline-Styles (style-src-elem) blockiert, da er gegen folgende Direktive verstößt: "style-src-elem 'self'". Ziehen Sie die Verwendung eines Hashs ('sha256-5v2scpgjNvJlJ7X3jvyxT8eEABeZfYB+9eiR+YLtQqo=', benötigt 'unsafe-hashes' für Stil-Attribute) oder einer Nonce in Erwägung. (in log)
```

and `<!DOCTYPE html>` seems to be missing at the top of all html files?

Zoey2936:
I've also added  `require-trusted-types-for 'script'; trusted-types 'none';`: https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Content-Security-Policy/require-trusted-types-for
it tends to break pages, but AIO seems to work.
For the CSP errors I wrote above: I've handeled this for now by allowing the hash, but moving them to a js/css file should be the better way
I've also allowed sameorigin embeding since it is used by AIO. And added three more headers:
https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Cross-Origin-Opener-Policy
https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Cross-Origin-Embedder-Policy
https://developer.mozilla.org/en-US/docs/Web/HTTP/Guides/Cross-Origin_Resource_Policy

Containers/mastercontainer/internal.Caddyfile
```diff
@@ -24,6 +24,23 @@
...
+        Referrer-Policy "no-referrer"
+        X-Robots-Tag "noindex, nofollow"
+        Cross-Origin-Opener-Policy "same-origin";
+        Cross-Origin-Embedder-Policy "require-corp";
+        Cross-Origin-Resource-Policy "same-origin";
+
+        -Server
+        -X-Powered-By
+        -Via
+    }
```
szaimen: Can we please move all of this to a dedicated file and import it in all necessary places? Then we don't need to duplicate this in all places :)

Zoey2936: done

Zoey2936:
(will add comments on the values in the file after everything else is done)

Containers/mastercontainer/acme.Caddyfile
```diff
@@ -29,10 +29,13 @@
 }
 
 http://:80 {
+    import headers.Caddyfile
```
szaimen: do we actually need the headers here if we always redirect to https?

Containers/mastercontainer/headers.Caddyfile
```diff
@@ -0,0 +1,16 @@
...
+    Content-Security-Policy "default-src 'none'; script-src-elem 'self' 'sha256-6bNyZN1sRD0pH97KGhXK/G/B9DKHZKxOiD7jfyz+EQ4='; script-src-attr 'unsafe-hashes' 'sha256-/9h5/n5yL7jhlwt387Bkaq9iLJ9eWZ53JtGvRLcqIOI=' 'sha256-5A3XM6HcvFhh9xP7kCNZfN0CGPwX97AvizFmXhdfAzc='; style-src-elem 'self' 'sha256-5v2scpgjNvJlJ7X3jvyxT8eEABeZfYB+9eiR+YLtQqo='; style-src-attr 'unsafe-hashes' 'sha256-BmMpY1aLoCxBMFnwjzq/SVkHNNPpa2IUx4R/PPxgwV8=' 'sha256-Vc6YdsPwJmyOyOWY3K0+DEFqLyvNxfsrADVC0UvQkxI='; img-src 'self'; connect-src 'self'; frame-src 'self'; base-uri 'none'; form-action 'self'; frame-ancestors 'self'; upgrade-insecure-requests; require-trusted-types-for 'script'; trusted-types 'none';"
+    X-Content-Type-Options "nosniff"
+    X-Frame-Options "SAMEORIGIN"
+    X-Permitted-Cross-Domain-Policies "none"
+    X-DNS-Prefetch-Control "off"
+    Referrer-Policy "no-referrer"
+    X-Robots-Tag "noindex, nofollow"
+    Cross-Origin-Opener-Policy "same-origin";
+    Cross-Origin-Embedder-Policy "require-corp";
+    Cross-Origin-Resource-Policy "same-origin";
```
szaimen: Please indeed add comments here to what each header does

szaimen (changes requested):
Thanks! A few suggestions and then this is ready I think :)

Containers/mastercontainer/acme.Caddyfile
Zoey2936: can be removed
```suggestion
```

Containers/mastercontainer/headers.Caddyfile
Zoey2936: so do you want to keep hash way? It works currently, but it will break if one space, char, line, etc is changed in the css/js. It would be better if all css/js would be moved to a file instead of being inline, but I currently don't have time for this

szaimen: You are right. Moving them to a separate file would be preferred. However I am not sure where inline styles and inline scripts are still used. Do you have a list of files that I could have a look at?

szaimen: I found the following places:
- https://github.com/nextcloud/all-in-one/blob/c104b179b75e866a8a1490d35358e56a041266b4/php/templates/log.twig#L5
- https://github.com/nextcloud/all-in-one/blob/c104b179b75e866a8a1490d35358e56a041266b4/php/src/Controller/DockerController.php#L384

So I guess these are the culprits currently or do we have more?

Zoey2936: - [x] https://github.com/nextcloud/all-in-one/blob/c104b179b75e866a8a1490d35358e56a041266b4/php/templates/login.twig#L14
- [x] https://github.com/nextcloud/all-in-one/blob/c104b179b75e866a8a1490d35358e56a041266b4/php/templates/containers.twig#L590
- [x] https://github.com/nextcloud/all-in-one/blob/c104b179b75e866a8a1490d35358e56a041266b4/php/templates/containers.twig#L591
- [x] https://github.com/nextcloud/all-in-one/blob/c104b179b75e866a8a1490d35358e56a041266b4/php/templates/includes/optional-containers.twig#L46
- [x] https://github.com/nextcloud/all-in-one/blob/c104b179b75e866a8a1490d35358e56a041266b4/php/templates/includes/optional-containers.twig#L81
- [x] https://github.com/nextcloud/all-in-one/blob/c104b179b75e866a8a1490d35358e56a041266b4/php/templates/includes/optional-containers.twig#L102
- [x] https://github.com/nextcloud/all-in-one/blob/c104b179b75e866a8a1490d35358e56a041266b4/php/templates/log.twig#L5-L55
- [x] https://github.com/nextcloud/all-in-one/blob/c104b179b75e866a8a1490d35358e56a041266b4/php/src/Controller/DockerController.php#L384-L394

Zoey2936: and `<!DOCTYPE html>` seems to be missing on nearly all pages

Containers/mastercontainer/headers.Caddyfile
```diff
@@ -0,0 +1,16 @@
+header {
+    Content-Security-Policy "default-src 'none'; script-src-elem 'self' 'sha256-6bNyZN1sRD0pH97KGhXK/G/B9DKHZKxOiD7jfyz+EQ4='; script-src-attr 'unsafe-hashes' 'sha256-/9h5/n5yL7jhlwt387Bkaq9iLJ9eWZ53JtGvRLcqIOI=' 'sha256-5A3XM6HcvFhh9xP7kCNZfN0CGPwX97AvizFmXhdfAzc='; style-src-elem 'self' 'sha256-5v2scpgjNvJlJ7X3jvyxT8eEABeZfYB+9eiR+YLtQqo='; style-src-attr 'unsafe-hashes' 'sha256-BmMpY1aLoCxBMFnwjzq/SVkHNNPpa2IUx4R/PPxgwV8=' 'sha256-Vc6YdsPwJmyOyOWY3K0+DEFqLyvNxfsrADVC0UvQkxI='; img-src 'self'; connect-src 'self'; frame-src 'self'; base-uri 'none'; form-action 'self'; frame-ancestors 'self'; upgrade-insecure-requests; require-trusted-types-for 'script'; trusted-types 'none';"
```
Zoey2936: I'm not sure if `require-trusted-types-for 'script'; trusted-types 'none'` could break things like these:
```
public/automatic_reload.js:7:            list[i].style.display = 'none';
public/containers-form-submit.js:10:        element.style.display = 'none';
public/containers-form-submit.js:14:    communityFormSubmit.style.display = 'none';
public/containers-form-submit.js:83:            element.style.display = hasChanges ? 'block' : 'none';
public/containers-form-submit.js:98:        communityFormSubmit.style.display = hasChanges ? 'block' : 'none';
```
needs to be tested

Zoey2936: see https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Content-Security-Policy/require-trusted-types-for for what this option does in CSP

Containers/mastercontainer/headers.Caddyfile
szaimen: should all be done now :)

Containers/mastercontainer/headers.Caddyfile
szaimen: I am also not sure. Can you please try? :)

szaimen: @Zoey2936 I think we can try to adjust the csp now again. Would you please go ahead? :)

Zoey2936: yes

Zoey2936:
```suggestion
    Content-Security-Policy "default-src 'none'; script-src-elem 'self'; style-src-elem 'self'; img-src 'self'; connect-src 'self'; frame-src 'self'; base-uri 'none'; form-action 'self'; frame-ancestors 'self'; upgrade-insecure-requests; require-trusted-types-for 'script'; trusted-types 'none';"
```

Zoey2936: do you know how this js is triggered?

szaimen: I think as long as we dont use innerhtml, this csp should work, no?

szaimen: Ok, would you push this to the branch and lets see if playwright still works? :)

Zoey2936: I think yes

szaimen: then lets try it out :)

szaimen:
@Zoey2936 it seems to work. At least the playwright test suceeds 🥳

Containers/mastercontainer/headers.Caddyfile
```diff
@@ -0,0 +1,16 @@
...
+    Content-Security-Policy "default-src 'none'; script-src-elem 'self'; style-src-elem 'self'; img-src 'self'; connect-src 'self'; frame-src 'self'; base-uri 'none'; form-action 'self'; frame-ancestors 'self'; upgrade-insecure-requests; require-trusted-types-for 'script'; trusted-types 'none';"
+    X-Content-Type-Options "nosniff"
+    X-Frame-Options "SAMEORIGIN"
+    X-Permitted-Cross-Domain-Policies "none"
+    X-DNS-Prefetch-Control "off"
+    Referrer-Policy "no-referrer"
+    X-Robots-Tag "noindex, nofollow"
+    Cross-Origin-Opener-Policy "same-origin";
+    Cross-Origin-Embedder-Policy "require-corp";
+    Cross-Origin-Resource-Policy "same-origin";
```
szaimen: can you please add comments now to what each item does? :)

Containers/mastercontainer/headers.Caddyfile
```diff
@@ -0,0 +1,16 @@
+header {
+    Content-Security-Policy "default-src 'none'; script-src-elem 'self'; style-src-elem 'self'; img-src 'self'; connect-src 'self'; frame-src 'self'; base-uri 'none'; form-action 'self'; frame-ancestors 'self'; upgrade-insecure-requests; require-trusted-types-for 'script'; trusted-types 'none';"
```
Zoey2936, line 13:
```suggestion
    Content-Security-Policy "default-src 'none'; script-src-elem 'self'; style-src-elem 'self'; img-src 'self'; connect-src 'self'; frame-src 'self'; base-uri 'none'; form-action 'self'; frame-ancestors 'self'; upgrade-insecure-requests; require-trusted-types-for 'script'; trusted-types 'none';" # CSP limits "what" can be used, by default we allow nothing and only allow required options (https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Content-Security-Policy)
```

Containers/mastercontainer/headers.Caddyfile
```diff
@@ -0,0 +1,16 @@
...
+    X-Content-Type-Options "nosniff"
```
Zoey2936:
```suggestion
    X-Content-Type-Options "nosniff" # This forces the browser to use the MIME type of the Content-Type header (https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/X-Content-Type-Options)
```

Containers/mastercontainer/headers.Caddyfile
szaimen: please make more specific which setting does what in the comment :)

Containers/mastercontainer/headers.Caddyfile
szaimen:
```suggestion
    X-Content-Type-Options "nosniff" # This forces the browser to use the MIME type of the Content-Type header. See https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/X-Content-Type-Options
```

Containers/mastercontainer/headers.Caddyfile
```diff
@@ -0,0 +1,16 @@
...
+    X-Frame-Options "SAMEORIGIN"
```
Zoey2936:
```suggestion
    X-Frame-Options "SAMEORIGIN" # only allow AIO itself to embed itself, this is also enfored as part of the CSP (frame-ancestors) (https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/X-Frame-Options)
```

szaimen:
```suggestion
    X-Frame-Options "SAMEORIGIN" # Only allow AIO itself to embed itself, this is also enforced as part of the CSP frame-ancestors. See https://developer.mozilla.org/de/docs/Web/HTTP/Reference/Headers/X-Frame-Options
```

Containers/mastercontainer/headers.Caddyfile
```diff
@@ -0,0 +1,16 @@
...
+    X-Permitted-Cross-Domain-Policies "none"
```
Zoey2936:
```suggestion
    X-Permitted-Cross-Domain-Policies "none" # we block all cross origin request, including ones from Adobe Acrobat or Microsoft Silverlight and Adobe Flash Player (https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/X-Permitted-Cross-Domain-Policies)
```

szaimen:
```suggestion
    X-Permitted-Cross-Domain-Policies "none" # We block all cross origin request, including ones from Adobe Acrobat or Microsoft Silverlight and Adobe Flash Player. See https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/X-Permitted-Cross-Domain-Policies
```

Containers/mastercontainer/headers.Caddyfile
```diff
@@ -0,0 +1,16 @@
...
+    X-DNS-Prefetch-Control "off"
```
Zoey2936:
```suggestion
    X-DNS-Prefetch-Control "off" # tells the browser to not prefetch the dns of linked pages (https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/X-DNS-Prefetch-Control)
```

Containers/mastercontainer/headers.Caddyfile
```diff
@@ -0,0 +1,16 @@
...
+    Referrer-Policy "no-referrer"
```
Zoey2936:
```suggestion
    Referrer-Policy "no-referrer" # tells the browser to never sent a Referer header (https://developer.mozilla.org/de/docs/Web/HTTP/Reference/Headers/Referrer-Policy)
```

Containers/mastercontainer/headers.Caddyfile
```diff
@@ -0,0 +1,16 @@
...
+    X-Robots-Tag "noindex, nofollow"
```
Zoey2936:
```suggestion
    X-Robots-Tag "noindex, nofollow" # tells web crawlers to not index this page (https://developer.mozilla.org/de/docs/Web/HTTP/Reference/Headers/X-Robots-Tag)
```

Containers/mastercontainer/headers.Caddyfile
szaimen:
```suggestion
    X-DNS-Prefetch-Control "off" # Tells the browser to not pre-fetch the DNS of linked pages. See https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/X-DNS-Prefetch-Control
```

Containers/mastercontainer/headers.Caddyfile
szaimen:
```suggestion
    Referrer-Policy "no-referrer" # Tells the browser to never sent a Referrer header. See https://developer.mozilla.org/de/docs/Web/HTTP/Reference/Headers/Referrer-Policy
```

Containers/mastercontainer/headers.Caddyfile
szaimen:
```suggestion
    X-Robots-Tag "noindex, nofollow" # Tells web crawlers to not index this page. See https://developer.mozilla.org/de/docs/Web/HTTP/Reference/Headers/X-Robots-Tag
```

Containers/mastercontainer/headers.Caddyfile
```diff
@@ -0,0 +1,16 @@
...
+    Cross-Origin-Opener-Policy "same-origin";
```
Zoey2936:
```suggestion
    Cross-Origin-Opener-Policy "same-origin"; # AIO does not use popup (Window.open()), still we can isolate its BCG if it is opened as a pop up by another page (https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Cross-Origin-Opener-Policy)
```

Containers/mastercontainer/headers.Caddyfile
```diff
@@ -0,0 +1,16 @@
...
+    Cross-Origin-Embedder-Policy "require-corp";
```
Zoey2936:
```suggestion
    Cross-Origin-Embedder-Policy "require-corp"; # harder rules for cross origin embeds (https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Cross-Origin-Embedder-Policy)
```

Containers/mastercontainer/headers.Caddyfile
Zoey2936:
```suggestion
    Cross-Origin-Resource-Policy "same-origin"; # only allow the same origin to load resources (https://developer.mozilla.org/en-US/docs/Web/HTTP/Guides/Cross-Origin_Resource_Policy)
```

Containers/mastercontainer/headers.Caddyfile
szaimen:
```suggestion
    Cross-Origin-Opener-Policy "same-origin"; # AIO does not use popup (Window.open()), still we can isolate its BCG if it is opened as a pop up by another page. See https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Cross-Origin-Opener-Policy
```

Containers/mastercontainer/headers.Caddyfile
szaimen:
```suggestion
    Cross-Origin-Embedder-Policy "require-corp"; # Harder rules for cross origin embeds. See https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Cross-Origin-Embedder-Policy

```

Containers/mastercontainer/headers.Caddyfile
Zoey2936: the comment will be very long

Containers/mastercontainer/headers.Caddyfile
szaimen:
```suggestion
    Cross-Origin-Resource-Policy "same-origin"; # Only allow the same origin to load resources. See https://developer.mozilla.org/en-US/docs/Web/HTTP/Guides/Cross-Origin_Resource_Policy
```

Containers/mastercontainer/headers.Caddyfile
Zoey2936: the header is called `Referer`, yes there is a typo in the header name

Containers/mastercontainer/headers.Caddyfile
szaimen: Yeah I see. Maybe we can split the comment into multiple lines? Something like:
```suggestion
    # CSP limits "what" can be used, by default we allow nothing and only allow required options. See https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Content-Security-Policy
    # default-src 'none'; <description>
    # script-src-elem 'self'; <description>
    # ...
    Content-Security-Policy "default-src 'none'; script-src-elem 'self'; style-src-elem 'self'; img-src 'self'; connect-src 'self'; frame-src 'self'; base-uri 'none'; form-action 'self'; frame-ancestors 'self'; upgrade-insecure-requests; require-trusted-types-for 'script'; trusted-types 'none';"
```

Containers/mastercontainer/headers.Caddyfile
Zoey2936: https://en.wikipedia.org/wiki/HTTP_referer#Etymology

Containers/mastercontainer/headers.Caddyfile
Zoey2936: not sure how to best write popup, with/without space or a hypthen etc

Containers/mastercontainer/headers.Caddyfile
szaimen: oh wow 🙈

szaimen:
```suggestion
    Referrer-Policy "no-referrer" # Tells the browser to never sent a Referer header. See https://developer.mozilla.org/de/docs/Web/HTTP/Reference/Headers/Referrer-Policy
```

Containers/mastercontainer/headers.Caddyfile
szaimen: Maybe?
```suggestion
    Cross-Origin-Opener-Policy "same-origin"; # AIO does not use any popup, still we can isolate its BCG if it is opened as a pop up by another page. See https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Cross-Origin-Opener-Policy
```

Containers/mastercontainer/headers.Caddyfile
Zoey2936:
```suggestion
    # CSP limits "what" can be used, by default we allow nothing and only allow required options. See https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Content-Security-Policy
    # default-src 'none'; allow nothing by default
    # script-src-elem/style-src-elem 'self'; only allow loading css/js files from same origin (AIO itself) while blocking all inline css/js
    # img-src 'self'; only allow loading images from same origin (from AIO itself)
    # connect-src 'self'; allow fetch to only connect same origin (to AIO itself)
    # frame-src 'self'; allow AIO to only embed itself "what can be embeded"
    # base-uri 'none'; this does not fallback to default-src, AIO does not use the html base tag
    # form-action 'self'; html forms are only allowed to submit to AIO and not cross origin
    # frame-ancestors 'self'; only allow AIO itself to embed it self "who can embed"
    # upgrade-insecure-requests; upgrade all http embedings to https
    # require-trusted-types-for 'script'; trusted-types 'none'; blocks DOM changes via js
    Content-Security-Policy "default-src 'none'; script-src-elem 'self'; style-src-elem 'self'; img-src 'self'; connect-src 'self'; frame-src 'self'; base-uri 'none'; form-action 'self'; frame-ancestors 'self'; upgrade-insecure-requests; require-trusted-types-for 'script'; trusted-types 'none';"
```

szaimen:
```suggestion
    # CSP limits which features can be used. By default we allow nothing and only allow required options. See https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Content-Security-Policy
    # default-src 'none'; Allow nothing by default
    # script-src-elem/style-src-elem 'self'; Only allow loading css/js files from same origin (AIO itself) while blocking all inline css/js
    # img-src 'self'; Only allow loading images from same origin (from AIO itself)
    # connect-src 'self'; Allow fetch to only connect same origin (to AIO itself)
    # frame-src 'self'; Allow AIO to only embed itself "what can be embeded"
    # base-uri 'none'; This does not fallback to default-src, AIO does not use the html base tag
    # form-action 'self'; Html forms are only allowed to submit to AIO and not cross origin
    # frame-ancestors 'self'; Only allow AIO itself to embed it self "who can embed"
    # upgrade-insecure-requests; Upgrade all http embedings to https
    # require-trusted-types-for 'script'; trusted-types 'none'; Blocks DOM changes via js
    Content-Security-Policy "default-src 'none'; script-src-elem 'self'; style-src-elem 'self'; img-src 'self'; connect-src 'self'; frame-src 'self'; base-uri 'none'; form-action 'self'; frame-ancestors 'self'; upgrade-insecure-requests; require-trusted-types-for 'script'; trusted-types 'none';"
```

szaimen (approved):
LGTM. Thank you very much! ☺️

Containers/mastercontainer/headers.Caddyfile
```diff
@@ -0,0 +1,28 @@
+header {
+    # CSP limits which features can be used. By default we allow nothing and only allow required options. See https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Content-Security-Policy
+    # default-src 'none'; Allow nothing by default
+    # script-src-elem/style-src-elem 'self'; Only allow loading css/js files from same origin (AIO itself) while blocking all inline css/js
+    # img-src 'self'; Only allow loading images from same origin (from AIO itself)
+    # connect-src 'self'; Allow fetch to only connect same origin (to AIO itself)
+    # frame-src 'self'; Allow AIO to only embed itself "what can be embeded"
```
Zoey2936:
```suggestion
    # frame-src 'self'; Allow AIO to only embed itself "what can be embedded"
```

Containers/mastercontainer/headers.Caddyfile
```diff
@@ -0,0 +1,28 @@
...
+    X-DNS-Prefetch-Control "off" # Tells the browser to not pre-fetch the DNS of linked pages. See https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/X-DNS-Prefetch-Control
+    Referrer-Policy "no-referrer" # Tells the browser to never sent a Referer header. See https://developer.mozilla.org/de/docs/Web/HTTP/Reference/Headers/Referrer-Policy
+    X-Robots-Tag "noindex, nofollow" # Tells web crawlers to not index this page. See https://developer.mozilla.org/de/docs/Web/HTTP/Reference/Headers/X-Robots-Tag
+    Cross-Origin-Opener-Policy "same-origin"; # AIO does not use any popup, still we can isolate its BCG if it is opened as a pop up by another page. See https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Cross-Origin-Opener-Policy
+    Cross-Origin-Embedder-Policy "require-corp"; # Harder rules for cross origin embeds. See https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Cross-Origin-Embedder-Policy
+
+    Cross-Origin-Resource-Policy "same-origin"; # Only allow the same origin to load resources. See https://developer.mozilla.org/en-US/docs/Web/HTTP/Guides/Cross-Origin_Resource_Policy
```
Zoey2936:
```suggestion
    Cross-Origin-Resource-Policy "same-origin"; # Only allow the same origin to load resources. See https://developer.mozilla.org/en-US/docs/Web/HTTP/Guides/Cross-Origin_Resource_Policy
```

szaimen:
🎉🎉🎉🎉🎉

Zoey2936:
I just hope nothing breaks! But else good to see this merged!

In theory there are three other header that can be set:
- https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Strict-Transport-Security, nextcloud itself requires this header, but not sure if this makes sense for AIO, for the selfsigned port this cannot be used, but maybe for the port with signed certs
- https://developer.mozilla.org/en-US/docs/Web/HTTP/Guides/Permissions_Policy / https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Permissions-Policy I've never worked with this header (since it only works in chromium based browser), but in theory AIO doesn't need any of these permissions, but sadly there seems to be no "deny all" option, instead any currently existing and maybe in the future added options need to be added to the header and disabled there...
- https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Origin-Agent-Cluster and this header, I've also never used it and don't know about possible downsides, but it also sounds good

Zoey2936:
Origin-Agent-Cluster seems to be good to set, should I create a PR?

szaimen:
> * https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Strict-Transport-Security, nextcloud itself requires this header, but not sure if this makes sense for AIO, for the selfsigned port this cannot be used, but maybe for the port with signed certs

Indeed that would only make sense for the caddy instance runing on port 8443...

> Origin-Agent-Cluster seems to be good to set, should I create a PR?

Yes, please go ahead :)

Zoey2936:
Created a PR for Origin-Agent-Cluster.
For the other two headers, please tell me what to do with them, I'm not sure

szaimen:
> Created a PR for Origin-Agent-Cluster.

Thanks! :)

> For the other two headers, please tell me what to do with them, I'm not sure

I would leave out the permission policy for now.

HSTS generally sounds good to add to port 8443. However obviously only there.

Zoey2936:
for hsts, with the same values as here: https://github.com/nextcloud/all-in-one/blob/12e129f1f637b553420d9bb4e5bf4bab0af4c2ae/Containers/apache/Caddyfile#L71 ?
(also the header is only applied for the nextcloud route there, but the header only makes sense when applied to all routes)

Zoey2936:
> I would leave out the permission policy for now.

can you maybe add it to a list or similiar so it doesn't get lost?

szaimen:
> for hsts, with the same values as here:
> 
> https://github.com/nextcloud/all-in-one/blob/12e129f1f637b553420d9bb4e5bf4bab0af4c2ae/Containers/apache/Caddyfile#L71
> 
> ?

Yes sounds good

> (also the header is only applied for the nextcloud route there, but the header only makes sense when applied to all routes)

Hm right... I am not sure if other applications have different requirements for this header but generally we can peobably add it to all routes

szaimen:
> > I would leave out the permission policy for now.
> 
> can you maybe add it to a list or similiar so it doesn't get lost?

Maybe it would make sense to add the header here in the file but comment it and add a todo comment above it?

Zoey2936:
> > > I would leave out the permission policy for now.
> > 
> > 
> > can you maybe add it to a list or similiar so it doesn't get lost?
> 
> Maybe it would make sense to add the header here in the file but comment it and add a todo comment above it?



> > > I would leave out the permission policy for now.
> > 
> > 
> > can you maybe add it to a list or similiar so it doesn't get lost?
> 
> Maybe it would make sense to add the header here in the file but comment it and add a todo comment above it?

I though about adding it to #5251

szaimen:
> > > > I would leave out the permission policy for now.
> > > 
> > > 
> > > can you maybe add it to a list or similiar so it doesn't get lost?
> > 
> > 
> > Maybe it would make sense to add the header here in the file but comment it and add a todo comment above it?
> 
> > > > I would leave out the permission policy for now.
> > > 
> > > 
> > > can you maybe add it to a list or similiar so it doesn't get lost?
> > 
> > 
> > Maybe it would make sense to add the header here in the file but comment it and add a todo comment above it?
> 
> I though about adding it to #5251

Ah I see. Ok, we can do this. Would you mind creating the issue for tracking? I can then add it to the overview ticket :)

Zoey2936:
#7931 and #7930

szaimen:

cybertaurus:
With Mac Safari 17.* there is an issue as `style-src-elem 'self'` is not yet supported. So it might be a solution to add `style-src 'self'` to the content-security-policies in headers.Caddyfile.

Zoey2936:
`style-src` is less strict and allows inline css, also it was fixed in safari 26.2 https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Content-Security-Policy/style-src-elem, so is this something we need to fix if it can be fixed by updating the browser?

PR #7693, opened 2026-03-05
docs: add warnings about IP address and self-signed certificate limitations

tjaadvd:
Many users attempt to run Nextcloud AIO on a local network using an IP address or a self-signed certificate, only to discover this is unsupported after going through the setup process. I personally tried this too, even though i (quickly) read the README and Local-instance pages.

This PR makes the limitation more visible upfront so others won't make the same mistake.

## Changes
- `local-instance.md`: Added a `[!WARNING]` callout at the top of the page clearly stating that AIO requires a real domain name and a valid TLS certificate, and that IP addresses and self-signed certificates are not supported.
- `readme.md`: Added a `[!WARNING]` callout in the "How to use this?" section directing local-only users to `local-instance.md` before proceeding with setup.

## Motivation
The existing documentation mentions these limitations, but only in passing within a longer paragraph. Users who skim the docs (or jump straight to the setup steps) miss this entirely. The new callouts ensure the constraint is visible before any setup steps are taken.

### DIFF
--- a/local-instance.md
+++ b/local-instance.md
@@ -1,4 +1,10 @@
 # Local instance
+
+> [!WARNING]
+> AIO requires a **real domain name** and a **valid TLS certificate** to function correctly.
+> Accessing Nextcloud via a plain IP address or a self-signed certificate is **not supported**.
+> The options below describe how to satisfy these requirements while keeping Nextcloud local.
+
 It is possible due to several reasons that you do not want or cannot open Nextcloud to the public internet. Perhaps you were hoping to access AIO directly from an `ip.add.r.ess` (unsupported) or without a valid domain.  However, AIO requires a valid certificate to work correctly. Below is discussed how you can achieve both: Having a valid certificate for Nextcloud and only using it locally.
 
 ### Content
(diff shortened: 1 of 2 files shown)


local-instance.md
```diff
@@ -1,4 +1,10 @@
 # Local instance
+
+> [!WARNING]
+> AIO requires a **real domain name** and a **valid TLS certificate** to function correctly.
```
pabzm, line 4: 1. "real domain name" is not very precise, our requirement is "publicly resolveable domain" (or hostname). That's not as catchy, if you can think of another wording, please let me know. But please replace "real".
3. Technically a self-signed certificate is also valid. You (probably) correctly mean a publicly verifiable certificate (issued by a well-known, public certificate authority), so please change the phrase accordingly.

local-instance.md
```diff
@@ -1,4 +1,10 @@
 # Local instance
+
+> [!WARNING]
+> AIO requires a **real domain name** and a **valid TLS certificate** to function correctly.
+> Accessing Nextcloud via a plain IP address or a self-signed certificate is **not supported**.
```
pabzm, line 5: Using a local CA is also unsupported.

pabzm (changes requested):
Thank you for the contribution!

I like the idea but would like to ask you for two text changes. Thanks!

local-instance.md
tjaadvd: Hi,

I'm having trouble coming up with the correct text. What do you think about this?

```
# Local instance

> [!WARNING]
> AIO requires a **publicly resolvable domain** and a **publicly verifiable certificate** (issued by a well-known, public certification authority).
```

local-instance.md
tjaadvd: How about this?

`
Accessing Nextcloud via a **plain IP address**, a **self-signed certificate**, or a **certificate from a local certificate authority (CA)** is **not supported**.
`

PR #7696, opened 2026-03-05
use a public/private key logic for token generation

Zoey2936:
see #7686

This would be my idea how this can be implemented, if other ways of doing something similar are preferred just close this


php/src/Auth/AuthManager.php
```diff
@@ -20,7 +20,38 @@ public function CheckCredentials(string $password) : bool {
...
+        }
+
+        $timestamp = sodium_crypto_sign_open($tokenBin, $publicKeyBin);
+
+        if ($timestamp === false) {
+            return false;
+        }
+
+        $timeElapsed = time() - (int) $timestamp;
+        if ($timeElapsed > 60) {
```
Zoey2936:
```suggestion
        if ($timeElapsed > 60 || $timeElapsed < 0) {
```
maybe block invalid times?

php/src/Auth/AuthManager.php
```diff
@@ -20,7 +20,38 @@ public function CheckCredentials(string $password) : bool {
...
+        }
+
+        $timestamp = sodium_crypto_sign_open($tokenBin, $publicKeyBin);
+
+        if ($timestamp === false) {
+            return false;
+        }
+
+        $timeElapsed = time() - (int) $timestamp;
+        if ($timeElapsed > 60 || $timeElapsed < 0) {
```
Zoey2936, line 42: not sure if this is enough time, if this is changed then apcu_add also needs to be changed

szaimen:
Honestly, I am not really sure of the implications of this PR. I would need to discuss this with @pabzm when he is back...

Zoey2936:
> Honestly, I am not really sure of the implications of this PR. I would need to discuss this with @pabzm when he is back...

any update?

szaimen:
@pabzm can you pleaee have a look at the proposed implementation? 

Maybe lets have a call about this soon

Zoey2936:
> @pabzm can you pleaee have a look at the proposed implementation?
> 
> Maybe lets have a call about this soon

sorry to ask again, any update?

Zoey2936:
renaming the branch seems to not work while keeping the renamed branch linked to this PR...

app/lib/Settings/Admin.php
```diff
@@ -55,7 +55,12 @@ public function getForm(): TemplateResponse {
...
 
-		$token = urlencode(getenv('AIO_TOKEN'));
+		$privateKeyBase64 = getenv('AIO_TOKEN');
+		$privateKeyBin = sodium_base642bin($privateKeyBase64, SODIUM_BASE64_VARIANT_URLSAFE_NO_PADDING);
+		$timestamp = (string) time();
+		$tokenBin = sodium_crypto_sign($timestamp, $privateKeyBin);
+		$token = sodium_bin2base64($tokenBin, SODIUM_BASE64_VARIANT_URLSAFE_NO_PADDING);
+
 		$params = [
 			'AIOLoginUrl' => 'https://' . getenv('AIO_URL') . '/api/auth/getlogin' . '?token=' . $token,
```
Zoey2936, line 65: to improve the validity of a token the buttong could also first redirect to a dummy page of the aio-app which then first created the token and redirects instantly, so tokens will only be issued if needed and not just because the admin overview is opened. Also the redirect should never take longer than 60 seconds, so the problem that someone opens the admin overview, waits sometime and then tries to login to aio with an expired token would be fixed

szaimen: Yes, this would be a good ida 👍

php/src/Auth/AuthManager.php
szaimen: maybe we can use a const and use that in both palces?

szaimen:
Hey @Zoey2936 I'll try to push forward with this. Sorry that it took so long.

szaimen:
I think the general idea is a good direction that we should improve on.

szaimen:
I'll try to include this in the next major release of AIO. See https://github.com/nextcloud/all-in-one/issues/8700

szaimen (changes requested):
Some valid comments by Claude Opus 5:

- **`php/src/Data/ConfigurationManager.php:28` — reuses the `AIO_TOKEN` config key**
`aioPublicKey` is stored under the same key that held the legacy 48-hex-char secret. On upgrade of an existing instance, that value base64-decodes to 36 bytes and `sodium_crypto_sign_open` throws, 500ing `/api/auth/getlogin`. This needs a distinct key name plus a migration path.
- **`php/src/Auth/AuthManager.php:35` — `sodium_crypto_sign_open()` is outside the try/catch**
`?token=abc` base64-decodes fine but then throws an uncaught `SodiumException` on an unauthenticated endpoint, instead of returning false.

php/src/Data/ConfigurationManager.php
```diff
@@ -1199,7 +1201,7 @@ private function getPlaceholderValue(string $placeholder) : string {
         return match ($placeholder) {
             'NC_DOMAIN' => $this->domain,
             'NC_BASE_DN' => $this->getBaseDN(),
-            'AIO_TOKEN' => $this->aioToken,
+            'AIO_TOKEN' => $this->aioPrivateKey,
```
szaimen, line 1204: btw, maybe we should rename the variable and name it AIO_PRIVATE_KEY or something alike?

Zoey2936:
I'm currently mainly working on NPMplus, so feel free to push/work on this branch if you want, otherwise it will take some time currently from my side

Zoey2936:
> php/src/Data/ConfigurationManager.php:28 — reuses the AIO_TOKEN config key
aioPublicKey is stored under the same key that held the legacy 48-hex-char secret. On upgrade of an existing instance, that value base64-decodes to 36 bytes and sodium_crypto_sign_open throws, 500ing /api/auth/getlogin. This needs a distinct key name plus a migration path.

The AIO_TOKEN is rotatet on every start, even curretnly with the shared secret, so this should not be a problem

szaimen:
> I'm currently mainly working on NPMplus, so feel free to push/work on this branch if you want, 

I see. Thanks for starting to work on this!

> otherwise it will take some time currently from my side

No problem at all! Take your time! 

If timing doesnt work out, I will take over :)

Zoey2936:
Feel free to take over

PR #7702, opened 2026-03-06
Fix portmapping in yml file

Maximilian-Maag:
When using the compose.yml file, docker compose throws an error about wrong port mapping notations. I changed the notation from:

 ports:
   - 80:80

to

 ports:
  - "80:80"
To address this issue. Hope this fixes the same issue for anybody else that uses current docker compose versions.

PR #7704, opened 2026-03-06
compose.yaml: fix port mappings

Maximilian-Maag:
Fixing port mappings in compose.yml for newer docker compose versions.
Now with proper rebaseing  to the correct main branch :).


Maximilian-Maag:
This PR adresses comments in PR #7702

Issue #7725, opened 2026-03-09
Wrong AIO_URL with Caddy HTTPS reverse proxy for AIO interface


LetUsFlow:
### Is your feature request related to a problem? Please describe. 

I use Caddy as reverse proxy for Nextcloud as well as the AIO interface.
After updating Caddy to the newest version 2.11, the autologin feature ("Open Nextcloud AIO Interface ↗") broke, redirecting to `localhost:8080/api/auth/getlogin?token=...` instead of my AIO URL.

I used the Caddy config from reverse-proxy.md, "6. Optional: get a valid certificate for the AIO interface":
```
https://<your-nc-domain>:8443 {
    reverse_proxy https://localhost:8080 {
        transport http {
            tls_insecure_skip_verify
        }
    }
}
```

Due to the change in caddyserver/caddy#7454, the Host header is rewritten to the upstream host when using reverse_proxy with https. So every time when starting the nextcloud containers from the AIO interface, `AIO_URL` gets set to `localhost:8080` instead of the actual AIO URL (in the mastercontainer configuration.json as well as the env variable).
Adding `header_up Host {host}` to the Caddy reverse_proxy config seems to fixe the issue.

### Describe the solution you'd like

Adding `header_up Host {host}` may be worth mentioning in the reverse-proxy.md docs, "6. Optional: get a valid certificate for the AIO interface":
```
https://<your-nc-domain>:8443 {
    reverse_proxy https://localhost:8080 {
        header_up Host {host}
        transport http {
            tls_insecure_skip_verify
        }
    }
}
```

### Describe alternatives you've considered

Alternatively, the handling of reverse proxies could be updated to honor X-Forwarded headers. But i don't know how niche this setup is, so just mentioning it in the docs probably makes more sense.

szaimen:
Hi, thanks for the report! We'll document this with https://github.com/nextcloud/all-in-one/pull/7728.

PR #7728, opened 2026-03-10
rp-docs: document new header_up requirement for proxying the aio-interface

szaimen:
- Close https://github.com/nextcloud/all-in-one/issues/7725


Zoey2936:
should this maybe also be added to the reverse proxy guide for the caddy container and all caddyfiles from aio?

szaimen:
> should this maybe also be added to the reverse proxy guide for the caddy container and all caddyfiles from aio?

As far as I understand this only applies for internal https connections so for internal http connection the behaviour is unchanged...

Zoey2936:
Interesting, yes this seems to be the case, but I don't understand the sense of this change and also why it only applies to https. 

https://caddyserver.com/docs/caddyfile/directives/reverse_proxy#https

the docs seem to recommend {hostport} instead of {host}

PR #7731, opened 2026-03-10
readme: add section how to move the appdata folder from the datadir to an ssd to improve the performance

szaimen:
Linking to https://help.nextcloud.com/t/symlink-for-appdata-folder-not-working/241468

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -476,7 +476,7 @@ Another solution if you really need to use host mounts is to use a bind mount to
 /source/path  /target/path/where/the/source/directory/will/be/mounted/on/the/server  fuse.bindfs  force-user=33,force-group=33,allow_other  0  0
 ```
 
-Then use `sudo mount -a` to mount it directly.
+Then use `sudo mount /target/path/where/the/source/directory/will/be/mounted/on/the/server` to mount it directly.
 
 You can afterwards use `--env NEXTCLOUD_DATADIR="/target/path/where/the/source/directory/will/be/mounted/on/the/server"` as described in the section above.
 
@@ -493,7 +493,7 @@ After the initial installation is done and all datadir files of Nextcloud are st
 ```
 Do not forget to adjust `<appdata-path>` to the correct `appdata_*` name that your installation initially created automatically.
 
-Then use `sudo mount -a` to mount it directly.
+Then use `sudo mount /target/path/<appdata-path>` to mount it directly.
 
 Afterwards things should be speed up.
 


readme.md
```diff
@@ -748,7 +768,10 @@ password=<password>
 ```
 (Of course you need to modify `<smb/cifs username>` and `<password>` for your specific case.)
 
-Now you can use `/mnt/storagebox` as Nextcloud's datadir like described in the section above this one.
+Now you can use `/mnt/storagebox` as Nextcloud's datadir like described in the section [here](#how-to-change-the-default-location-of-nextclouds-datadir).
+
+> [!NOTE]
```
szaimen, line 773: The initial installation

readme.md
```diff
@@ -475,7 +476,26 @@ Another solution if you really need to use host mounts is to use a bind mount to
...
+> The following steps only work if you already configured and used NEXTCLOUD_DATADIR as mentioned [two sections above](#how-to-change-the-default-location-of-nextclouds-datadir). 
+> In this example here, we assume that you used `NEXTCLOUD_DATADIR="/target/path/`.
+
+After the initial installation is done and all datadir files of Nextcloud are stored inside the configured `/target/path` directory, you will also see an `appdata_*` folder in there that stores app-related data. You can now move that folder to a faster SSD if the target dir is not already positioned on an SSD by first using `rsync` to sync the files a location on an SSD. Afterwards rename the appdata folder in the datadir to something like `appdata_*-backup`. Afterwards add the following line to `/etc/fstab`:
+```
+/source/path/on/ssd  /target/path/<appdata-path>  fuse.bindfs  force-user=33,force-group=33,allow_other  0  0
+```
+Do not forget to adjust `<appdata-path>` to the correct `appdata_*` name that your installation initially created automatically.
+
+Then use `sudo mount -a` to mount it directly.
```
pabzm: Same as above


```suggestion
Then use `sudo mount /target/path/<appdata-path>` to mount it directly.
```

readme.md
```diff
@@ -475,7 +476,26 @@ Another solution if you really need to use host mounts is to use a bind mount to
 /source/path  /target/path/where/the/source/directory/will/be/mounted/on/the/server  fuse.bindfs  force-user=33,force-group=33,allow_other  0  0
 ```
 
-You can then use `--env NEXTCLOUD_DATADIR="/target/path/where/the/source/directory/will/be/mounted/on/the/server"` as described in the section above.
+Then use `sudo mount -a` to mount it directly.
```
pabzm: `mount -a` mounts everything that's marked as auto-mountable in `/etc/fstab`, so calling this could have side effects. How about:


```suggestion
Then use `sudo mount /target/path/where/the/source/directory/will/be/mounted/on/the/server` to mount it.
```

Most people will either have a short enough path for that, or have the path in their paste buffer anyways, I'd guess.

readme.md
szaimen: good idea! fixed!

readme.md
szaimen: good idea! fixed!

PR #7737, opened 2026-03-11
Exclude /lost+found from rsync

pabzm:
This prevents an error that sometimes leads to irritation (even though it's harmless)


szaimen:

PR #7741, opened 2026-03-12
Give all containers a display_name, hide from list by dedicated attribute

pabzm:
We want a display-name for all containers in the overlay-log, but still want to exclude some containers from the list on top of the web UI.

### DIFF
--- a/php/src/Container/Container.php
+++ b/php/src/Container/Container.php
@@ -38,8 +38,8 @@ readonly class Container {
         public string                        $imageTag,
         public AioVariables                  $aioVariables,
         public string                        $documentation,
-        private DockerActionManager           $dockerActionManager,
-        public bool                          $hideFromList
+        public bool                          $hideFromList,
+        private DockerActionManager           $dockerActionManager
     ) {
     }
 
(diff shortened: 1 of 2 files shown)


php/src/Container/Container.php
```diff
@@ -38,7 +38,8 @@ public function __construct(
         public string                        $imageTag,
         public AioVariables                  $aioVariables,
         public string                        $documentation,
-        private DockerActionManager           $dockerActionManager
+        private DockerActionManager           $dockerActionManager,
+        public bool                          $hideFromList
```
szaimen: Lets invert the order
```suggestion
        public bool                          $hideFromList
        private DockerActionManager           $dockerActionManager,
```

php/src/ContainerDefinitionFetcher.php
```diff
@@ -349,7 +351,8 @@ private function GetDefinition(): array
                 $imageTag,
                 $aioVariables,
                 $documentation,
-                $this->container->get(DockerActionManager::class)
+                $this->container->get(DockerActionManager::class),
+                $hideFromList
```
szaimen: same

```suggestion
                $hideFromList,
                $this->container->get(DockerActionManager::class)
```

php/src/Container/Container.php
pabzm: Thank you! (I didn't apply your suggestion because it contains a syntax error.)

szaimen:

PR #7762, opened 2026-03-17
fix collabora not working anymore

### LINKED ISSUE #7757: develop-tag: after recent collabora update, it does not work anymore and shows a unautorized wopi url error
### Steps to reproduce
1. after recent collabora container image upgrade, collabora does not open anymore and shows `unautorized wopi url error` (the problem is currently only present in the develop channel)
2. I get the following error in the collabora container logs: `No authorized hosts found matching the target host [nextcloud-aio-apache] in config`

### Expected behavior 
It should open the editor

### Actual behavior 
It doesnt.

asavageiv:
- Close https://github.com/nextcloud/all-in-one/issues/7757

CollaboraOnline/online/pull/13869 prevents http://nextcloud-aio-apache:23973 from being correctly recognized as a valid URI and thus breaks matching as an authorized WOPI host.

This uses http://nextcloud-aio-apache.nextcloud-aio:23973 which is correctly recognized as a valid URI.


manual-install/latest.yml
```diff
@@ -249,7 +249,7 @@ services:
     expose:
       - "9980"
     environment:
-      - aliasgroup1=https://${NC_DOMAIN}:443,http://nextcloud-aio-apache:23973
+      - aliasgroup1=https://${NC_DOMAIN}:443,http://nextcloud-aio-apache.nextcloud-aio:23973
```
szaimen: Please revert this. It willl be adjusted by a script later on automatically.

nextcloud-aio-helm-chart/templates/nextcloud-aio-collabora-deployment.yaml
```diff
@@ -28,7 +28,7 @@ spec:
             - name: TZ
               value: "{{ .Values.TIMEZONE }}"
             - name: aliasgroup1
-              value: https://{{ .Values.NC_DOMAIN }}:443,http://nextcloud-aio-apache:23973
+              value: https://{{ .Values.NC_DOMAIN }}:443,http://nextcloud-aio-apache.nextcloud-aio:23973
```
szaimen: Please revert this. It will be adjusted by a script later on automatically.

szaimen (changes requested):
Hey, thanks a lot! I have two comments below:

asavageiv:
Done! Thanks for the quick review!

szaimen (approved):
LGTM, thanks! :)

szaimen:

chanceymathews:
I'm using v12.9.2 and this breaks things for me on Podman, because containers can't resolve `nextcloud-aio-apache.nextcloud-aio`. It can only resolve `nextcloud-aio-apache`. I can only get it to resolve by manually adding nextcloud-aio-apache.nextcloud-aio to the hosts file on -collabora container. 

I imagine this is a Docker/Podman compatibility problem so not supported, but thought it would help anyone else using Podman who runs into this.

asavageiv:
Ah, yeah podman doesn't alter search domains like docker. I'm fine with
reverting this change once the upstream fix is released (I haven't
checked). You can use --network-alias with podman to work around it.

apparle:
This is affecting me as well on podman -- podman doesn't resolve `nextcloud-aio-apache.nextcloud-aio` similar to docker.
I see that collabara didn't merge the fix for the original issue, but I'm not sure of the release cadence. Does Nextcloud AIO use only the released version or does it pull from github at a certain cadence?

@asavageiv because AIO creates the nextcloud-aio-apache on the fly, I'm not sure how to add `--network-alias` to the container . Is there a custom variable or similar I can use to setup an alias?

chanceymathews:
> This is affecting me as well on podman -- podman doesn't resolve `nextcloud-aio-apache.nextcloud-aio` similar to docker. I see that collabara didn't merge the fix for the original issue, but I'm not sure of the release cadence. Does Nextcloud AIO use only the released version or does it pull from github at a certain cadence?
> 
> @asavageiv because AIO creates the nextcloud-aio-apache on the fly, I'm not sure how to add `--network-alias` to the container . Is there a custom variable or similar I can use to setup an alias?

My temporary workaround is adding the entry to the hosts files on the containers that need it. In my quadlet file for the master container I have:

ExecStartPost=sh -c 'echo $(podman exec -i nextcloud-aio-apache hostname -i) nextcloud-aio-apache.nextcloud-aio | podman exec -i -u0 nextcloud-aio-collabora tee --append /etc/hosts'
ExecStartPost=sh -c 'echo $(podman exec -i nextcloud-aio-apache hostname -i) nextcloud-aio-apache.nextcloud-aio | podman exec -i -u0 nextcloud-aio-nextcloud tee --append /etc/hosts'

Could probably do something with `podman network connect --alias` but this method works well enough for now.

apparle:
> My temporary workaround is adding the entry to the hosts files on the containers that need it. In my quadlet file for the master container I have:
> 
> ExecStartPost=sh -c 'echo $(podman exec -i nextcloud-aio-apache hostname -i) nextcloud-aio-apache.nextcloud-aio | podman exec -i -u0 nextcloud-aio-collabora tee --append /etc/hosts' ExecStartPost=sh -c 'echo $(podman exec -i nextcloud-aio-apache hostname -i) nextcloud-aio-apache.nextcloud-aio | podman exec -i -u0 nextcloud-aio-nextcloud tee --append /etc/hosts'
> 
> Could probably do something with `podman network connect --alias` but this method works well enough for now.

That won't work when the nextcloud-aio-apache container is automatically restarted by mastercontainer for daily backups + updates right?

chanceymathews:
> That won't work when the nextcloud-aio-apache container is automatically restarted by mastercontainer for daily backups + updates right?

It will still work if `nextcloud-aio-apache` is restarted, because the entry is added to `nextcloud-aio-nextcloud` and `nextcloud-aio-collabora` containers, so it will persist until they are restarted. I don't use the built-in backups so I can't say for sure. 

A more robust approach might be to add a service to monitor podman events and then reattach the container to the network with the alias. I tested this and it works fine and should handle container reboots without any trouble.

`nextcloud-apache-fix.service`
```
[Unit]
Documentation=https://github.com/nextcloud/all-in-one/pull/7762#issuecomment-4288719521

[Service]
Restart=always
ExecStart=sh -c 'podman events -f container=nextcloud-aio-apache -f type=container -fevent=start | while read -r line; do podman network disconnect nextcloud-aio nextcloud-aio-apache; podman network connect --alias nextcloud-aio-apache.nextcloud-aio nextcloud-aio nextcloud-aio-apache; done'
```

Add `Wants=nextcloud-apache-fix.service` to your AIO container unit.

asavageiv:
I'm going to try reverting this to fix podman since I think the upstream is fixed now

alutarius:
> > That won't work when the nextcloud-aio-apache container is automatically restarted by mastercontainer for daily backups + updates right?
> 
> It will still work if `nextcloud-aio-apache` is restarted, because the entry is added to `nextcloud-aio-nextcloud` and `nextcloud-aio-collabora` containers, so it will persist until they are restarted. I don't use the built-in backups so I can't say for sure.
> 
> A more robust approach might be to add a service to monitor podman events and then reattach the container to the network with the alias. I tested this and it works fine and should handle container reboots without any trouble.
> 
> `nextcloud-apache-fix.service`
> 
> ```
> [Unit]
> Documentation=https://github.com/nextcloud/all-in-one/pull/7762#issuecomment-4288719521
> 
> [Service]
> Restart=always
> ExecStart=sh -c 'podman events -f container=nextcloud-aio-apache -f type=container -fevent=start | while read -r line; do podman network disconnect nextcloud-aio nextcloud-aio-apache; podman network connect --alias nextcloud-aio-apache.nextcloud-aio nextcloud-aio nextcloud-aio-apache; done'
> ```
> 
> Add `Wants=nextcloud-apache-fix.service` to your AIO container unit.

this works around the issue well, thanks

apparle:
> I'm going to try reverting this to fix podman since I think the upstream is fixed now

Is there an issue or open PR tracking this? I couldn't locate one.
If this isn't being done now and needs to wait for some versions getting deployed, it'll be good to create an open issue to track this.

PR #7763, opened 2026-03-17
aio-interface: Fix bypass_container_update when not detected as available

asavageiv:
isAnyUpdateAvailable is not necessarily synced with the PullImage logic from DockerActionManager. This ensures no images are pulled regardless of detection.

There's also a minor code improvement to avoid unnecessary interpolation.


szaimen:

Issue #7778, opened 2026-03-19
develop-tag: aio-interface feels very slow now


szaimen:
### Steps to reproduce
1. set up new instance with develop tag
2. see that loading is very slow
3. I suspect this is due to https://github.com/nextcloud/all-in-one/blob/90d5d6bcf1b6bb7de7955ee3ff98d29a3d9bf928/Containers/mastercontainer/internal.Caddyfile#L23-L24

### Expected behavior 
it should feel fast as before

### Actual behavior 
it doesnt

szaimen:
Cc @pabzm

szaimen:
- Possible fix here: https://github.com/nextcloud/all-in-one/pull/7779

szaimen:
This is now released with v12.9.0 Beta. Testing and feedback is welcome! See https://github.com/nextcloud/all-in-one#how-to-switch-the-channel

Issue #7765, opened 2026-03-18
playwright tests broken


szaimen:
### Steps to reproduce
1. Currently the playwright tests are broken and need fixing. See e.g. https://github.com/nextcloud/all-in-one/actions/runs/23218451571/job/67485162871?pr=7762
2. After I build new container images, the playwright tests are broken and need fixing. I suppose this has to do with https://github.com/nextcloud/all-in-one/pull/7006. 
3.

### Expected behavior 
They should still work

### Actual behavior 
They dont

szaimen:
cc @pabzm @Zoey2936

pabzm:
From my research this is unlikely to be caused by our latest changes, since it's coming from the network stack.

pabzm:
https://issues.chromium.org/issues/41465264 has a lot of people complaining about it

pabzm:
https://github.com/docker/for-linux/issues/914

Maybe github is playing with IPv6 (finally!)?

pabzm:
The playwright-report from the CI shows that the error occurs after the initial login, so maybe the cause is our changing the networks when starting up additional containers.

szaimen:
> The playwright-report from the CI shows that the error occurs after the initial login, so maybe the cause is our changing the networks when starting up additional containers.

Yeah sounds like it. However weird that this was not a problem before... Anyway I have a possible fix in https://github.com/nextcloud/all-in-one/pull/7769

PR #7779, opened 2026-03-19
mastercontainer: use remote-host caddy plugin only for login endpoints

szaimen:
- Fix https://github.com/nextcloud/all-in-one/issues/7778
- Fix part of https://github.com/nextcloud/all-in-one/issues/7765


szaimen:

PR #7806, opened 2026-03-23
Fix log viewer on small screens

### LINKED ISSUE #7758: log viewer still not working 100% correctly yet on mobile
### Steps to reproduce
1. open log viewer on mobile
2. See it is not 100% responsive yet

### Expected behavior 
It should be responisve

### Actual behavior 
It is not yet

See short video:

https://github.com/user-attachments/assets/5a744223-e6d1-4479-8e97-9f95932f942c

pabzm:
Previously it showed a lot of white space at the bottom, and the floating box didn't stick. Both is now fixed.

- Closes #7758

### DIFF
--- a/php/templates/log.twig
+++ b/php/templates/log.twig
@@ -53,7 +53,7 @@
                 transition: opacity 1s, display 1s allow-discrete;
             }
         </style>
-        <script src="log-view.js"></script>
+        <script src="log-view.js?v1"></script>
     </head>
     <body data-container-id="{{ id }}">
         <div id="floating-box">


php/public/log-view.js
```diff
@@ -96,7 +96,7 @@ class LogViewer {
     }
 
     scrollToBottom() {
-        window.scrollTo(0, document.body.scrollHeight);
+        this.logElem.scrollTop = this.logElem.scrollHeight;
```
szaimen, line 99: I guess we need to increase the version of the file here to make it work for existing instances: https://github.com/nextcloud/all-in-one/blob/7c8cabdb2ddfd3760abf67ca6fe3b0ef204e0746/php/templates/log.twig#L46

szaimen:
This is now released with v12.9.1 Beta. Testing and feedback is welcome! See https://github.com/nextcloud/all-in-one#how-to-switch-the-channel

PR #7841, opened 2026-04-02
aio-interface: throttle login attempts to 5 failures per 5 minutes

pabzm:
Currently there's no throttling at all, which is not a good idea. This introduces a limit of 5 attempts in 5 minutes.

### DIFF
--- a/php/src/Controller/LoginController.php
+++ b/php/src/Controller/LoginController.php
@@ -11,10 +11,31 @@ use Psr\Http\Message\ResponseInterface as Response;
 use Psr\Http\Message\ServerRequestInterface as Request;
 
 readonly class LoginController {
+    private const int MAX_LOGIN_ATTEMPTS_PER_TTL = 5;
+    private const int LOGIN_COUNTER_TTL = 300;
+    private const string RATE_LIMIT_CACHE_KEY = 'login_failed_attempts';
+
     public function __construct(
         private AuthManager $authManager,
         private DockerActionManager $dockerActionManager,
     ) {
+    }
+
+    private function getFailedLoginCount() : int {
+        $count = apcu_fetch(self::RATE_LIMIT_CACHE_KEY);
+        return $count !== false ? (int)$count : 0;
+    }
+
+    private function incrementFailedLoginCount() : void {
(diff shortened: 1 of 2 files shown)


php/src/Controller/LoginController.php
```diff
@@ -11,23 +11,52 @@
 use Psr\Http\Message\ServerRequestInterface as Request;
 
 readonly class LoginController {
+    private const int MAX_LOGIN_ATTEMPTS_PER_TTL = 5;
+    private const int LOGIN_COUNTER_TTL = 300;
+    private const string RATE_LIMIT_CACHE_KEY = 'login_failed_attempts';
```
szaimen, line 16: shouldnt we also store the ip-address or someting like that? Otherwise an external brute force attempt might block login even for correct login attempts, no?

pabzm: I supposed that this approach would be good enough, but maybe you're right.

But I don't like using IP-addresses for these purposes, because they make things complicated and brittle with CGNAT, optional reverse proxies, etc. (should we also block the IP if it's part of an ISP's CGNAT? How should we determine when to trust the `X-Forwarded-For` HTTP header?). Also, serious attackers send their requests from a lot of hosts or proxies anyway.

I'd alternatively suggest to extend the response times in case of failed password validation, so that even the error is only shown to the user (or bot) after an increasing amount of time (e.g. 30s or 1min). Then an attacking bot might slow down a legitime user's login (who e.g. has to try several passwords from a list), but won't block it completely.

szaimen: Maybe a simple start would be to slow down login attempts per session to one every 5s or something like that?

But maybe lets have a call about the general idea soon :)

pabzm: I created a new PR to get the 5 second penalty into the code base until we figured out a more complex solution that suits us: https://github.com/nextcloud/all-in-one/pull/7863

It's a step forward, which is progress, too.

This PR I'll close, because it's not the solution that suits us.

pabzm:
Closing in favour of https://github.com/nextcloud/all-in-one/pull/7863 and possibly a different, more complex solution in the future

PR #7845, opened 2026-04-02
aio-interface: fix page not automatic reloading after container starting progress spinner

szaimen:


php/public/automatic_reload.js
```diff
@@ -1,4 +1,4 @@
-document.addEventListener("DOMContentLoaded", function(event) {
+window.addEventListener("load", function(event) {
     if (document.hasFocus()) {
```
szaimen, line 2: I think this is the main problem. The iframe is a different document. So if it is still focused after the page loaded, the page does not have the focus yet. Needs testing.

Issue #7833, opened 2026-03-30
Wrong TOKEN on "Open Nextcloud AIO Interface" button


gruffatremi:
### Steps to reproduce
1. Connect to nextcloud at https://server.ip/settings/admin/overview
2. Click on [Open Nextcloud AIO Interface] 
3. Get the page with "The login is blocked since Nextcloud is running.
Please use the [automatic login](https://github.com/nextcloud/all-in-one#how-to-easily-log-in-to-the-aio-interface) from your Nextcloud." 

### Expected behavior 
1. Connect to nextcloud at https://server.ip/settings/admin/overview
2. Click on [Open Nextcloud AIO Interface] 
3. Arrive on Nextcloud AIO admin interface

### Actual behavior 

The URL in the nextcloud button have the wrong token in the URL, I found the good token by typing sudo docker exec nextcloud-aio-mastercontainer cat /mnt/docker-aio-config/data/configuration.json, it gives me AIO_TOKEN, if I replace it with the wrong one, I am able to connect to the admin interface, but this is not really easy to do always. 

#### Output of `sudo docker info`
``` 
Client: Docker Engine - Community
 Version:    29.3.1
 Context:    default
 Debug Mode: false
 Plugins:
  buildx: Docker Buildx (Docker Inc.)
    Version:  v0.31.1
    Path:     /usr/libexec/docker/cli-plugins/docker-buildx
  compose: Docker Compose (Docker Inc.)
    Version:  v5.1.1
    Path:     /usr/libexec/docker/cli-plugins/docker-compose

Server:
 Containers: 23
  Running: 21
  Paused: 0
  Stopped: 2
 Images: 24
 Server Version: 29.3.1
 Storage Driver: overlay2
  Backing Filesystem: extfs
  Supports d_type: true
  Using metacopy: false
  Native Overlay Diff: true
  userxattr: false
 Logging Driver: json-file
 Cgroup Driver: cgroupfs
 Cgroup Version: 2
 Plugins:
  Volume: local
  Network: bridge host ipvlan macvlan null overlay
  Log: awslogs fluentd gcplogs gelf journald json-file local splunk syslog
 CDI spec directories:
  /etc/cdi
  /var/run/cdi
 Swarm: inactive
 Runtimes: io.containerd.runc.v2 nvidia runc
 Default Runtime: runc
 Init Binary: docker-init
 containerd version: 301b2dac98f15c27117da5c8af12118a041a31d9
[...]

PR #7851, opened 2026-04-02
aio-interface: avoid `Resend` popups on Firefox on start/stop containers

### LINKED ISSUE #7850: "Resend" popup after Stop/Start Containers
### Steps to reproduce
1. On the AIO page click "Stop Containers" or "Start Containers"
2. Wait for streaming of container Stop/Start to finish

### Expected behavior 
When the streaming is done. The page reloads or I'm showing a button or prompt to reload the page.

### Actual behavior 
I'm shown the following popup.

Image

Pressing "Resend" sends another `start` or `stop` POST and then the page reloads. 

I think we don't want to redo the start or stop on success, even though it is idempotent. We want to `GET /containers`

#### Other valuable info 

Image

asavageiv:
Use `location.reload(true)` instead of `reload` or `reload(1)` to ensure we use a GET request when reloading the page after streaming the `start` or `stop` request completes.

See also:
https://stackoverflow.com/a/41122753
https://developer.mozilla.org/en-US/docs/Web/API/Location/reload#forceget

- Fixes #7850
- Close https://github.com/nextcloud/all-in-one/issues/7833

### DIFF
--- a/php/public/forms.js
+++ b/php/public/forms.js
@@ -36,11 +36,11 @@ function showPassword(id) {
       showError("Server error. Please check the mastercontainer logs for details. This page will reload after 10s automatically. Then you can check the mastercontainer logs.");
       // Reload after 10s since it is expected that the updated view is shown (e.g. after starting containers)
       setTimeout(function(){
-        window.location.reload(1);
+        window.location.reload(true);
       }, 10000);
     } else {
       // If the responose is not one of the above, we should reload to show the latest content
-      window.location.reload(1);
+      window.location.reload(true);
     }
   }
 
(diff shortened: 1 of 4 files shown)


szaimen:
Thanks you for the fast fix @asavageiv 😊

Please increase the version here and then this should be ready to go from my pov :)
https://github.com/nextcloud/all-in-one/blob/1a7feba6bd60c7c5dc474c9ee3db54d2f2963225/php/templates/layout.twig#L6

php/public/forms.js
```diff
@@ -84,7 +84,7 @@ function showPassword(id) {
             document.getElementById('overlay-log')?.classList.add('visible');
             // Reload the page after the response was fully loaded into the iframe.
             document.querySelector('iframe[name="overlay-log"]').addEventListener('load', () => {
-                location.reload();
+                location.reload(true);
```
szaimen, line 87: So this is the actual fix, right?

asavageiv:
Updated the JS versions and signed the commits. Should be ready now. :)

szaimen:
This is now released with v12.9.2 Beta. Testing and feedback is welcome! See https://github.com/nextcloud/all-in-one#how-to-switch-the-channel

PR #7852, opened 2026-04-02
Add dev convenience script for building containers

asavageiv:
A small script to make building the containers for local dev testing a little easier:

```
Usage: ./scripts/build-containers.sh [--variant develop|beta] <container1> [container2] [container3] ...
Example: ./scripts/build-containers.sh --variant beta apache mastercontainer
```

Feel free to reject this if there's a better way.

### DIFF
--- a/scripts/build-containers.sh
+++ b/scripts/build-containers.sh
@@ -1,0 +1,47 @@
+#!/bin/bash
+
+VARIANT="develop"
+CONTAINERS=()
+
+# Parse flags first
+while [[ $# -gt 0 && $1 == --* ]]; do
+  case $1 in
+    --variant)
+      VARIANT="$2"
+      shift 2
+      ;;
+    *)
+      echo "Unknown option: $1"
+      exit 1
+      ;;
+  esac
+done
+
+# Remaining arguments are containers
+CONTAINERS=("$@")
+
+if [ ${#CONTAINERS[@]} -eq 0 ]; then
+    echo "Usage: $0 [--variant develop|beta] <container1> [container2] [container3] ..."
+    echo "Example: $0 --variant beta apache mastercontainer"
+    exit 1
+fi
+
+# Change to project root
+cd "$(dirname "$0")/.." || exit 1
+
+for container in "${CONTAINERS[@]}"; do
+  if [[ $container == "mastercontainer" ]]; then
+    TAG="all-in-one"
+  else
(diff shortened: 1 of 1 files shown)


szaimen:
Hi I am not sure about adding this to the repo as we would also need to maintain it then. Wdyt @pabzm?

pabzm:
I don't need this script, but also wouldn't mind it – but I'd also prefer to only include such scripts if multiple devs or contributors would like to use it.

asavageiv:
Out of curiousity, how do you all build the containers when doing local dev? Do you just type it out every time or use shell history? Maybe I am doing it wrong.

asavageiv:
The idea is loosely based on this:
https://github.blog/engineering/scripts-to-rule-them-all/

I use docker devcontainers heavily so I'm often spinning up a clean environment with no shell history.

pabzm:
> how do you all build the containers when doing local dev?

I don't do it often, usually work with the images built in the CI.

Why do you build them locally?

asavageiv:
I build them locally for faster iteration. Do you make a commit and push to CI before doing any testing or are you editing directly inside of running containers? My flow was:
1. Make edits to the code
2. Build appropriate containers with appropriate tags (sometimes push the image to a remote machine, depending on where I'm doing the editing vs where the containers are running)
3. Stop and start the containers with bypass_container_update/mastercontainer_update
4. Do testing

Sometimes I would make edits directly in the container then copy them back to my source code directory.
I tried out bind-mounting some directories into containers, but permissions got tricky.

scripts/build-containers.sh
```diff
@@ -0,0 +1,47 @@
...
+
+# Parse flags first
+while [[ $# -gt 0 && $1 == --* ]]; do
+  case $1 in
+    --variant)
+      VARIANT="$2"
+      shift 2
+      ;;
+    *)
+      echo "Unknown option: $1"
```
gggeek, line 14: errors and warnings should echo to >&2

scripts/build-containers.sh
```diff
@@ -0,0 +1,47 @@
+#!/bin/bash
+
+VARIANT="develop"
+CONTAINERS=()
+
+# Parse flags first
+while [[ $# -gt 0 && $1 == --* ]]; do
```
gggeek, line 7: what about adding a help text when users pass in `-h` ?

PR #7863, opened 2026-04-07
aio-interface: Punish failed login attempts with a delay

pabzm:
This is a very simple means against bots, until we find the time to discuss and implement something better.


php/src/Controller/LoginController.php
szaimen, line 44: we should proabably do the same here, right?

szaimen (commented):
Thanks! One comment below:

szaimen (approved):
LGTM, thanks! ☺️

szaimen:

PR #7864, opened 2026-04-07
apps.config.php: allow to disable the internal app store

szaimen:


pabzm (changes requested):
[Removed, was a misunderstanding]

Containers/nextcloud/config/apps.config.php
```diff
@@ -16,6 +16,12 @@
 if (getenv('APPS_ALLOWLIST')) {
     $CONFIG['appsallowlist'] = explode(" ", getenv('APPS_ALLOWLIST'));
 }
-if (getenv('NEXTCLOUD_APP_STORE_URL')) {
-    $CONFIG['appstoreurl'] = getenv('NEXTCLOUD_APP_STORE_URL');
+
+$appStoreUrl = getenv('NEXTCLOUD_APP_STORE_URL');
+if ($appStoreUrl) {
+    if ($appStoreUrl === 'no') {
```
pabzm, line 22: Maybe check for the value being blank (unset or `""`) to disable the appstore, too?

szaimen: I would only disable it if explicitely set to no. Blank or unset will not set the value iirc...

pabzm: We use blank as "no", and unset as "use default" for other options, too IIRC, that's why I think it would make the config more consistent.

But this is just a recommendation, not a veto.

pabzm (approved):
I would prefer to (also) use a blank string as negative, but I'm still approving.

Thank you for developing this!

szaimen:

PR #7890, opened 2026-04-10
apache: Add an explanation why we disable the HTTP challenge in that Caddy instance

pabzm:
It confused people before, maybe this helps a little bit.

### DIFF
--- a/Containers/apache/Caddyfile
+++ b/Containers/apache/Caddyfile
@@ -75,6 +75,7 @@ http://{$APACHE_HOST}.nextcloud-aio:23973, # For Collabora callback and WOPI req
     tls {
         issuer acme {
             profile shortlived
+            # Disable HTTP challenge because that would require port 80, which we don't get (it's exposed to the mastercontainer).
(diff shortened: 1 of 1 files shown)


Containers/apache/Caddyfile
```diff
@@ -75,6 +75,7 @@ http://{$APACHE_HOST}.nextcloud-aio:23973, # For Collabora callback and WOPI req
     tls {
         issuer acme {
             profile shortlived
+            # Disable HTTP challenge because that would require port 80, which we don't get (it's exposed to the mastercontainer).
```
szaimen, line 78:
```suggestion
            # Disable HTTP challenge because that would require port 80, which we don't get (it's exposed to the mastercontainer).
            # This container by default only exposes port 443 if not configured otherwise via APACHE_PORT.
```

szaimen (commented):
Nice idea but one suggestion below

szaimen (approved):
LGTM :)

PR #7897, opened 2026-04-13
Let manual-install use WOPI URLs internal to the Docker network

Fs00:
I've realized that my AIO manual installation was still using the public domain as the WOPI URL for Collabora instead of the Docker internal hostname of the Apache container (what the mastercontainer setup does since #6676).
Two changes were needed to make this work on the manual setup:
- the `occ richdocuments:activate-config` command differed between the mastercontainer and manual install: the former used the one defined in containers.json (which specified the internal WOPI URLs), while the latter used the one found in run-exec-commands.sh script of the NextCloud AIO container.
I've moved the command with the correct parameters in the latter script file, to make it consistent with both setups and to prevent potential future differences.
- the Apache container internal hostname wasn't being resolved in the manual-install because the name of the Docker network was being [automatically determined by Docker Compose](https://docs.docker.com/compose/how-tos/networking/) using the parent folder name. By specifying the expected name of the network in the compose file, the behavior is now consistent with the mastercontainer setup.


szaimen (changes requested):
Hi, this change will unfortunately not work in the case of the helm chart deployment: https://github.com/nextcloud/all-in-one/tree/main/nextcloud-aio-helm-chart which is the main reason why we kept the behaviour like that for the manual-install...

manual-install/update-yaml.sh
```diff
@@ -159,6 +159,7 @@ cat << NETWORK >> containers.yml
 
 networks:
   default:
+    name: nextcloud-aio
```
szaimen, line 162: This single change would be fine from my pov

Fs00:
@szaimen Can you elaborate on why this change would not work for the Helm chart deployment? I'm not familiar with Kubernetes, but I guess it doesn't support containers talking to each other in the same Docker network like in Docker Compose?
Would it be possible to find a way to make this an exception for the Helm chart then? I'm thinking of a container env variable for the Collabora WOPI URL that is set differently for the Helm chart, but let me know your opinion 🙏

szaimen:
AFAIK helm does not have the concept of docker networks, that is why it won't work...

Fs00:
I have reworked my approach: now the Collabora WOPI URL and callback URL are configurable via environment variables of the NextCloud container, so that they can be easily configured for the manual-install and excluded in the Helm chart (let me know if I did that correctly, I wasn't able to test the Helm chart update script on my PC due to an error with Kompose).
To make the latter scenario work as it did before, the Collabora WOPI URL falls back to `https://{COLLABORA_DOMAIN}` when it's unset.
@szaimen Please let me know your feedback!

Containers/nextcloud/entrypoint.sh
```diff
@@ -754,10 +749,13 @@ if [ "$COLLABORA_ENABLED" = 'yes' ]; then
...
         php /var/www/html/occ app:update richdocuments
     fi
-    php /var/www/html/occ config:app:set richdocuments wopi_url --value="https://$COLLABORA_HOST/"
-    # Make collabora more save
-    COLLABORA_IPv4_ADDRESS="$(dig "$COLLABORA_HOST" A +short +search | grep '^[0-9.]\+$' | sort | head -n1)"
-    COLLABORA_IPv6_ADDRESS="$(dig "$COLLABORA_HOST" AAAA +short +search | grep '^[0-9a-f:]\+$' | sort | head -n1)"
+
+    COLLABORA_WOPI_URL=${COLLABORA_WOPI_URL:=https://$COLLABORA_DOMAIN}
+    php /var/www/html/occ config:app:set richdocuments wopi_url --value="$COLLABORA_WOPI_URL"
+
```
Fs00, line 755: I really would have liked to configure the callback URL here too, but [the `richdocuments:activate-config` command resets the callback URL](https://github.com/nextcloud/richdocuments/blob/11789de8c5a5f606761ed5663c0eeaad2fdbfbcc/lib/Command/ActivateConfig.php#L50) when it's called without the `callback-url` parameter...

Containers/nextcloud/entrypoint.sh
```diff
@@ -738,12 +738,7 @@ php /var/www/html/occ config:app:set notify_push base_endpoint --value="https://
 
 # Collabora
 if [ "$COLLABORA_ENABLED" = 'yes' ]; then
-    set -x
-    if echo "$COLLABORA_HOST" | grep -q "nextcloud-.*-collabora"; then
-        COLLABORA_HOST="$NC_DOMAIN"
-    fi
-    set +x
-    # Remove richdcoumentscode if it should be incorrectly installed
+    # Remove richdocumentscode should it be incorrectly installed
```
szaimen, line 741: Please keep this logic. Maybe you need to rename COLLABORA_DOMAIN to COLLABORA_HOST further below

Containers/nextcloud/run-exec-commands.sh
```diff
@@ -14,10 +14,15 @@ if [ -n "$NEXTCLOUD_EXEC_COMMANDS" ]; then
...
+    # exec-commands for Collabora are specified here since Collabora must work also if using manual-install 
     if [ "$COLLABORA_ENABLED" = yes ]; then
         echo "Activating Collabora config..."
-        php /var/www/html/occ richdocuments:activate-config
+        ACTIVATE_CONFIG_COMMAND="php /var/www/html/occ richdocuments:activate-config"
+        if [ -n "$COLLABORA_WOPI_CALLBACK_URL" ]; then
+            $ACTIVATE_CONFIG_COMMAND --callback-url="$COLLABORA_WOPI_CALLBACK_URL"
+        else
+            $ACTIVATE_CONFIG_COMMAND
+        fi
```
szaimen, line 25: probably best to restucture this a bit. Something like:
```suggestion
        COLLABORA_ACTIVATE_CONFIG_ARGS=()
        if [ -n "$COLLABORA_WOPI_CALLBACK_URL" ]; then
            COLLABORA_ACTIVATE_CONFIG_ARGS=(--callback-url="$COLLABORA_WOPI_CALLBACK_URL")
        fi
        php /var/www/html/occ richdocuments:activate-config "${COLLABORA_ACTIVATE_CONFIG_ARGS[@]}"
```

php/containers.json
```diff
@@ -228,7 +228,9 @@
         "CLAMAV_HOST=nextcloud-aio-clamav",
         "ONLYOFFICE_ENABLED=%ONLYOFFICE_ENABLED%",
         "COLLABORA_ENABLED=%COLLABORA_ENABLED%",
-        "COLLABORA_HOST=nextcloud-aio-collabora",
+        "COLLABORA_DOMAIN=%NC_DOMAIN%",
```
szaimen, line 231:
```suggestion
        "COLLABORA_HOST=nextcloud-aio-collabora",
```

Containers/nextcloud/run-exec-commands.sh
Fs00: Yeah yours looks definitely better, I'm not very proficient in Bash scripting 😅

Containers/nextcloud/entrypoint.sh
Fs00: I've renamed the `COLLABORA_HOST` variable because I think that the previous name doesn't reflect its usage anymore.
Now the variable it's used merely to construct the fallback for the WOPI URL when the latter is unset (mainly for compatibility with the Helm chart) and for determining the IP address of the Collabora domain, and thus the name `COLLABORA_DOMAIN`.

Given that now we have a dedicated env variable for the WOPI URL, why do we need the extra logic? It seems unnecessary and potentially confusing to me, but maybe there is some edge case I'm not seeing.

Fs00:
It's been a while, but before getting back to work to get this in a mergeable state I have a few questions:
- since Collabora has been updated to a version that includes a fix for the root cause of #7757, the workaround introduced in #7762 can be reverted: will this make the changes in this PR work on the Helm Chart like on the other deployments?
- regarding the `COLLABORA_HOST` env variable: is the rename/logic change problematic? As I wrote in the [comment above](https://github.com/nextcloud/all-in-one/pull/7897#discussion_r3164175126), I do think that its "magical" behavior can create confusion when paired with the two new environment variables, but I can live with it if you want to keep it as-is.

@szaimen let me know :)

szaimen:
Hey @Fs00, please check my suggested changes and commit them if they look good to you. Also unfortunately there are some conflicts now that need to be resolved. 

Sorry for my late reply. I was very busy the recent weeks. 

Maybe @pabzm can further handle this going forward?

PR #7901, opened 2026-04-13
nextcloud-entrypoint.sh: make log-type configurable

szaimen:


Containers/nextcloud/entrypoint.sh
```diff
@@ -438,11 +438,19 @@ EOF
...
+                php /var/www/html/occ config:system:set log_type --value="errorlog"
+                php /var/www/html/occ config:system:set log_type_audit --value="errorlog"
+                php /var/www/html/occ app:disable logreader
+            else
+                php /var/www/html/occ config:system:set log_type --value="file"
+                php /var/www/html/occ config:system:set log_type_audit --value="file"
+                php /var/www/html/occ app:enable logreader
+            fi
             php /var/www/html/occ config:system:set logfile --value="/var/www/html/data/nextcloud.log"
+            php /var/www/html/occ config:system:set logfile_audit --value="/var/www/html/data/audit.log"
```
pabzm: I think these two should move into the `else`-branch above. As an admin of an automated installation like AIO I think it's pretty confusing to see a config option being set in the config file which contradicts another option (and to have to find out that one of the options doesn't apply).


```suggestion
                php /var/www/html/occ config:system:set logfile --value="/var/www/html/data/nextcloud.log"
                php /var/www/html/occ config:system:set logfile_audit --value="/var/www/html/data/audit.log"
            fi

```

Containers/nextcloud/entrypoint.sh
```diff
@@ -640,8 +648,17 @@ fi
...
+    php /var/www/html/occ config:system:set log_type_audit --value="errorlog"
+    php /var/www/html/occ app:disable logreader
+else
+    php /var/www/html/occ config:system:set log_type --value="file"
+    php /var/www/html/occ config:system:set log_type_audit --value="file"
+    php /var/www/html/occ app:enable logreader
+fi
 php /var/www/html/occ config:system:set logfile --value="/var/www/html/data/nextcloud.log"
-php /var/www/html/occ config:app:set admin_audit logfile --value="/var/www/html/data/audit.log"
+php /var/www/html/occ config:system:set logfile_audit --value="/var/www/html/data/audit.log"
```
pabzm: Same as above


```suggestion
    php /var/www/html/occ config:system:set logfile --value="/var/www/html/data/nextcloud.log"
    php /var/www/html/occ config:system:set logfile_audit --value="/var/www/html/data/audit.log"
fi
```

Containers/nextcloud/entrypoint.sh
pabzm: Also I think these two sections should be refactored. Maybe not in this PR, but soon. The entrypoint script is already over 1000 lines long. From my point of view it should be maximum half that long.

pabzm (changes requested):
Thank you for the work!

I have only minor nitpicks, and one suggestion.

Containers/nextcloud/entrypoint.sh
szaimen:
```suggestion
                php /var/www/html/occ config:system:set logfile --value="/var/www/html/data/nextcloud.log"
                php /var/www/html/occ config:system:set logfile_audit --value="/var/www/html/data/audit.log"
            fi
```

szaimen: makes sense, thanks!

Containers/nextcloud/entrypoint.sh

pabzm (approved):
Looks good, thank you! (Didn't test yet, though)

szaimen:

PR #7902, opened 2026-04-13
allow to adjust the log level globally

szaimen:
This is best reviewed commit by commit


.github/workflows/sync-normalize-aio-log-level.yml
```diff
@@ -0,0 +1,34 @@
...
+        uses: peter-evans/create-pull-request@c0f553fe549906ede9cf27b5156039d195d2ece0 # v7
+        with:
+          commit-message: Sync normalize_aio_log_level helper
+          signoff: true
+          title: Sync normalize_aio_log_level helper
+          body: Automated sync of the shared normalize_aio_log_level helper into container directories that rely on local build contexts.
+          labels: dependencies, 3. to review
+          milestone: next
+          branch: aio-sync-normalize-aio-log-level
+          token: ${{ secrets.GITHUB_TOKEN }}
```
szaimen: @pabzm do you have a better idea how to sync this script over to all necessary containers?

szaimen: resolved

manual-install/sample.conf
```diff
@@ -25,7 +25,10 @@ APACHE_IP_BINDING=0.0.0.0          # This can be changed to e.g. 127.0.0.1 if yo
 APACHE_MAX_SIZE=17179869184          # This needs to be an integer and in sync with NEXTCLOUD_UPLOAD_LIMIT
 APACHE_PORT=443          # Changing this to a different value than 443 will allow you to run it behind a web server or reverse proxy (like Apache, Nginx, Caddy, Cloudflare Tunnel and else).
 ADDITIONAL_COLLABORA_OPTIONS=['--o:security.seccomp=true']          # You can add additional collabora options here by using the array syntax.
+AIO_LOG_LEVEL=warn          # Allows to adjust the global AIO log level. Valid values are debug, info, warning and error.
```
szaimen: `# Allows to adjust the global AIO log level. Valid values are debug, info, warn and error.`

manual-install/update-yaml.sh
```diff
@@ -100,6 +100,9 @@ sed -i 's|NC_DOMAIN=|NC_DOMAIN=yourdomain.com          # TODO! Needs to be chang
 sed -i 's|NEXTCLOUD_PASSWORD=|NEXTCLOUD_PASSWORD=          # TODO! This is the password of the initially created Nextcloud admin with username "admin".|' sample.conf
 sed -i 's|TIMEZONE=|TIMEZONE=Europe/Berlin          # TODO! This is the timezone that your containers will use.|' sample.conf
 sed -i 's|COLLABORA_SECCOMP_POLICY=|COLLABORA_SECCOMP_POLICY=--o:security.seccomp=true          # Changing the value to false allows to disable the seccomp feature of the Collabora container.|' sample.conf
+sed -i 's|AIO_LOG_LEVEL=|AIO_LOG_LEVEL=warning          # Allows to adjust the global AIO log level. Valid values are debug, info, warning and error.|' sample.conf
```
szaimen: `sed -i 's|AIO_LOG_LEVEL=|AIO_LOG_LEVEL=warning          # Allows to adjust the global AIO log level. Valid values are debug, info, warn and error.|' sample.conf`

Containers/mastercontainer/start.sh
```diff
@@ -333,6 +337,20 @@ else
     export NEXTCLOUD_DRI_GID=""
 fi
 
+# Log level logics
+if [ -n "$AIO_LOG_LEVEL" ] && ! grep -q "^debug$\|^info$\|^warn$\|^error$"; then
+    print_red "AIO_LOG_LEVEL must be one of 'debug', 'info', 'warn' or 'error'.
+It is set to '$AIO_LOG_LEVEL'".
+    exit 1
+fi
+CADDY_LOG_LEVEL="$(echo AIO_LOG_LEVEL | tr '[:lower:]' '[:upper:]')"
```
szaimen: Need to add a default if `AIO_LOG_LEVEL` is not set so that caddy in the mastercontainer still works

Also
`CADDY_LOG_LEVEL="$(echo "$AIO_LOG_LEVEL" | tr '[:lower:]' '[:upper:]')"`

szaimen (commented):
Playwright test is expected to fail at this point as the log level is not set in the mastercontainer. We could change the test though to mount start.sh. Then it should work...

Containers/mastercontainer/start.sh
szaimen: fixed

manual-install/update-yaml.sh
szaimen: fixed

manual-install/sample.conf
szaimen: fixed

.github/workflows/playwright-on-push.yml
szaimen: Also need to chmod +x of the start.sh

szaimen: done

Containers/collabora/healthcheck.sh
```diff
@@ -1,5 +1,9 @@
 #!/bin/bash
 
+if [ "$AIO_LOG_LEVEL" = 'debug' ]; then
+    set -x
+fi
+
```
pabzm: Isn't this pretty useless, since the scripts just exits cleanly in the next line?

Containers/collabora-online/healthcheck.sh
```diff
@@ -1,5 +1,9 @@
 #!/bin/bash
 
+if [ "$AIO_LOG_LEVEL" = 'debug' ]; then
+    set -x
+fi
+
```
pabzm: Same question for usefulness here

Containers/imaginary/start.sh
```diff
@@ -1,5 +1,20 @@
...
+if [ "$AIO_LOG_LEVEL" = 'debug' ]; then
+    export DEBUG='*'
+    export GOLANG_LOG='info'
+elif [ "$AIO_LOG_LEVEL" = 'info' ]; then
+    export GOLANG_LOG='info'
+elif [ "$AIO_LOG_LEVEL" = 'warn' ]; then
+    export GOLANG_LOG='warning'
+else
+    export GOLANG_LOG='error'
+fi
```
pabzm: A `case` statement would be more readable, I think:

```
case "$AIO_LOG_LEVEL" in
    debug)
        export DEBUG='*'
        export GOLANG_LOG='info'
        ;;
    info)
        export GOLANG_LOG='info'
        ;;
    warn)
        export GOLANG_LOG='warning'
        ;;
    *)
        export GOLANG_LOG='error'
esac
```

But this is not a blocker, just a suggestion.

(For some reason I can't insert a code suggestion here.)

php/src/Data/ConfigurationManager.php
```diff
@@ -1112,6 +1120,22 @@ private function getPlaceholderValue(string $placeholder) : string {
...
+    }
+
+    private function getElasticLogLevel() : string {
+        return match ($this->aioLogLevel) {
+            'debug' => 'DEBUG',
+            'info' => 'INFO',
+            'error' => 'ERROR',
+            default => 'WARN',
+        };
+    }
```
pabzm: I don't like to have these custom "translations" in here, because they are very specific to their containers and should happen inside the container logic, as you implemented it for all other containers.

I guess you chose this way because those two containers don't have an entrypoint yet, but I'd strongly prefer to create one, then, instead of having this logic as a special case in our PHP code base.

(Special cases like these make it hard(er) to comprehend how things work for someone not yet intimate with all details (or us at a later point in time, having forgotten some details), and I care a lot about comprehensibility, since it's a very helpful criterion to help avoiding building up technical debt.)

pabzm (changes requested):
Generally looks good, just the custom "translations" in PHP don't sit with me well.

I didn't test it, though.

szaimen:
Todo: imginary, talk and talk-recording need defaults in their dockerfile for the aio log level

Containers/collabora/healthcheck.sh
szaimen: Right, will remove this, thanks!

Containers/collabora-online/healthcheck.sh
szaimen: Right, will remove this, thanks!

Containers/imaginary/start.sh
szaimen: Good suggestion, thanks!

php/src/Data/ConfigurationManager.php
szaimen: Hm, honestly creating our own entrypoint means more effort keeping it up-to-date with the upstream one. As I consider collabora and elasticsearch as 3rd party services I wanna keep them as close to upstream as possible to keep effort on our side as low as possible. 

Btw the string replacement happens on docker run level, so it should not cause any tech debt going foward. 

But lets talk about this soon :)

szaimen: And in the case of elasticsearch we could probably simplify the function by using a to upper dunction instead...

szaimen:
```suggestion
    private function getCollaboraLogLevel() : string {
        return match ($this->aioLogLevel) {
            'warn' => 'warning',
            default => $this->aioLogLevel,
        };
    }

    private function getElasticLogLevel() : string {
        strtoupper($this->aioLogLevel);
    }
```

Containers/imaginary/start.sh
szaimen:
```suggestion
case "$AIO_LOG_LEVEL" in
    debug)
        export DEBUG='*'
        export GOLANG_LOG='info'
        ;;
    info)
        export GOLANG_LOG='info'
        ;;
    warn)
        export GOLANG_LOG='warning'
        ;;
    *)
        export GOLANG_LOG='error'
esac
```

Containers/collabora-online/healthcheck.sh
szaimen:
```suggestion
```

Containers/collabora/healthcheck.sh
szaimen:
```suggestion
```

pabzm (approved):
Looks good, thank you!

(I didn't test it yet, though.)

szaimen:

Issue #7622, opened 2026-02-18
Update all instances to nc33


szaimen:
## Tasks
 
TODO:
- [x] Wait for at least Nextcloud 33.0.1
 
Issues:
- [x] https://github.com/nextcloud/server/issues/59941
- [ ] https://github.com/nextcloud/server/issues?q=is%3Aopen%20is%3Aissue%20label%3A33-feedback%20label%3Ahigh
- [ ] https://github.com/nextcloud/server/issues?q=is%3Aopen+is%3Aissue++label%3A33-feedback+

Apps not compatible:
- [x] increase aio version to Nextcloud 33 https://github.com/nextcloud/all-in-one/pull/7876
- [x] fulltextsearch
- [x] fulltextsearch_elasticsearch
- [x] onlyoffice
- [x] files_antivirus
- [x] theming_customcss
- [x] twofactor_admin
- [x] [other bundled apps](https://cloud.nextcloud.com/apps/tables/#/table/182)
- [x] drawio
- [x] files_linkeditor
- [x] memories
- [ ] metadata
- [x] music
- [x] passwords
- [ ] further apps: https://help.nextcloud.com/t/apps-with-missing-support-for-nc33/239672

Test:
- [x] check if all included features work
    - [x] ClamAV 
    - [x] Collabora
    - [x] OnlyOffice
    - [x] Fulltextsearch 
    - [x] Imaginary (for previews of heic, heif, illustrator, pdf, svg, tiff and webp.
    - [x] Talk
    - [x] talk recording
    - [x] ~~docker-socket-proxy~~ (deprecated)
    - [x] whiteboard
    - [x] HARP
- [x] check admin overview
    - [x] Error: Configuration server ID; Server identifier isn’t configured. It is recommended if your Nextcloud instance is running on several PHP servers. Add a server ID in your configuration. https://github.com/nextcloud/all-in-one/pull/7910
- [x] check each other overview
- [x] check if nextcloud-aio app works
- [x] add other issues/features that should get in for the major release
- [x] https://github.com/nextcloud/all-in-one/pull/7877
- [x] https://github.com/nextcloud/all-in-one/pull/7879
- [x] https://github.com/nextcloud/all-in-one/issues/7609
- [x] https://github.com/nextcloud/all-in-one/issues/7407
- [ ] increase docker API_VERSION ? (check latest beta or stable release logs of docker)
[...]

PR #7910, opened 2026-04-15
nextcloud: add `server.config.php` that adds the serverid by using the hostname as key

szaimen:
- For https://github.com/nextcloud/all-in-one/issues/7622

### DIFF
--- a/Containers/nextcloud/config/server.config.php
+++ b/Containers/nextcloud/config/server.config.php
@@ -1,0 +1,4 @@
+<?php
+$CONFIG = array (
+  'serverid' => crc32(gethostname()) % 512,
+);


Containers/nextcloud/config/server.config.php
```diff
@@ -0,0 +1,4 @@
+<?php
+$CONFIG = array (
+  'serverid' => crc32(gethostname()) % 1024,
```
Altahrim: To match usage: 
https://github.com/nextcloud/server/blob/master/lib/private/Snowflake/SnowflakeGenerator.php#L37
```suggestion
  'serverid' => crc32(gethostname()) & 0x1FF,
```
(also fine with `% 512`)

Altahrim (approved):

Containers/nextcloud/config/server.config.php
szaimen: > (also fine with `% 512`)

lets use this then :)

szaimen:
```suggestion
  'serverid' => crc32(gethostname()) % 512,
```

szaimen: done

szaimen:

PR #7912, opened 2026-04-15
aio-interface: set cookie prefix to __Host-Http-

Zoey2936:
- Follow-up to https://github.com/nextcloud/all-in-one/pull/7689
- Address https://github.com/nextcloud/all-in-one/pull/7689#issuecomment-4252309050

### DIFF
--- a/php/public/index.php
+++ b/php/public/index.php
@@ -38,6 +38,7 @@ $container->set(Guard::class, function () use ($responseFactory) {
 
 // Register Middleware To Be Executed On All Routes
 session_start([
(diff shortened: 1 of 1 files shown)


php/public/index.php
```diff
@@ -38,6 +38,7 @@
 
 // Register Middleware To Be Executed On All Routes
 session_start([
+    "name" => "__Host-Http-PHPSESSID", // set cookie prefix to prevent other pages from overwriting this cookie. See https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Set-Cookie#cookie_prefixes
```
szaimen:
```suggestion
    "name" => "__Host-Http-PHPSESSID", // Set cookie prefix to prevent other pages from overwriting this cookie. See https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Set-Cookie#cookie_prefixes
```

szaimen (approved):
LGTM, thanks! ☺️

szaimen:

Issue #7893, opened 2026-04-11
Talk HPB: Janus advertises Docker bridge IP as ICE candidate, breaking direct WebRTC connections


JoachimSchurig:
## Summary

When running Nextcloud AIO with the Talk High Performance Backend in Docker bridge network mode (the default), Janus advertises its Docker-internal IP (e.g. `172.21.0.4`) as ICE host candidate. This prevents direct WebRTC connections between clients and forces all media through the internal eturnal relay. External TURN servers are rendered useless, and direct IPv6 P2P connections are impossible.

## Root Cause

In `start.sh`, the Talk host IP is resolved via:

```bash
IPv4_ADDRESS_TALK="$(dig "$TALK_HOST" IN A +short +search | grep '^[0-9.]\+$' | sort | head -n1)"
```

Since `TALK_HOST=nextcloud-aio-talk`, this resolves to the Docker-internal IP (`172.21.0.4`), not the host's actual network address. This value is never used to set Janus's `nat_1_1_mapping` in `janus.jcfg`, and the setting remains at its default (unset).

As a result, Janus has no knowledge of the host's real IP and advertises its container IP as ICE candidate — which is unreachable for any client outside the Docker bridge network.

## Impact

- **No direct P2P connections**: Clients cannot connect to Janus directly. All media must be relayed through the internal eturnal, even for clients on the same LAN as the host.
- **External TURN servers are useless**: Any externally configured TURN server (e.g. coturn on a VPS) allocates relay addresses correctly, but the relay cannot reach Janus at `172.21.0.4`. TURN relay candidates are generated but never succeed.
- **IPv6 broken**: Even when both client and host have global IPv6 addresses, the direct IPv6 path is blocked because Janus only offers its Docker-internal candidate.
- **Fragile connections**: The only working media path is through the internal eturnal within the Docker network, which is a single point of failure and adds unnecessary latency.

## Evidence

ICE candidate pair from `about:webrtc` (Firefox) during a call with HPB enabled:

| Status | Local Candidate | Remote Candidate (Janus) |
|---|---|---|
| succeeded | `172.21.0.4:62765/udp(relay-tcp)` | `172.21.0.4:59149/udp(host)` |

Both sides show the Docker bridge IP. The connection only works because the internal eturnal is on the same Docker network.

After **disabling HPB**, clients connect directly via IPv6 P2P:

| Status | Local Candidate | Remote Candidate |
|---|---|---|
[...]

PR #7926, opened 2026-04-16
nextcloud-entrypoint: rewrite turn and stun logic to always add turn and stun server

szaimen:
- The default Stun and Turn servers are required when using the bundled talk conainer
- Linking back to to https://github.com/nextcloud/all-in-one/issues/7893#issuecomment-4260768752


szaimen:

PR #7981, opened 2026-04-22
nextcloud: enable `taskprocessing:worker` as standard service

szaimen:
- Close https://github.com/nextcloud/all-in-one/discussions/5430


szaimen:

PR #7984, opened 2026-04-23
nextcloud: allow to configure `memcache_customprefix` for redis

szaimen:


szaimen:

PR #7985, opened 2026-04-23
nextcloud: adjust phpredis session timeouts to match Nextclouds config

szaimen:
- Follow-up to https://github.com/nextcloud/all-in-one/pull/7980


szaimen:

PR #7999, opened 2026-04-25
aio-interface: only set the placeholder for timezone and not the value automatically because it is confusing

szaimen:


szaimen:

PR #8055, opened 2026-05-04
home-assistant: Update readme.md

alexkannes222:
Add information to use the Caddy community container as reverse proxy and not only the external reverse proxy option to make Home Assistant accessible remotely over https, and add the documentation that needs to be done in Home Assistant to allow reverse proxy traffic.

Signed-off-by: Tim Alexander Neuenbauer <alex.kannes222@gmail.com>


alexkannes222:
Note, that the commit contains a "Signed-off-by:" that was automatically created by Github, yet the upper one is the correct one.

community-containers/home-assistant/readme.md
```diff
@@ -6,6 +6,8 @@ This container bundles Home Assistant and auto-configures it for you.
 - After adding and starting the container, you can visit `http://ip.address.of.this.server:8123` in order to set up your Home Assistant instance.
 - The data of Home Assistant will be automatically included in AIOs backup solution!
 - In order to access your Home Assistant outside the local network, you have to set up your own reverse proxy. You can set up a reverse proxy following [these instructions](https://github.com/nextcloud/all-in-one/blob/main/reverse-proxy.md).
+- And to allow the traffic from the reverse proxy to be accepted by Home Assistant, follow [these instructions](https://www.home-assistant.io/integrations/http/#reverse-proxies) from the Home Assistant documentation. 
+- Or, to use the Caddy with geoblocking community container, follow the following instruction to add your own Caddyfile, to use it for Home Assistant: https://github.com/nextcloud/all-in-one/blob/99ea91c5ef5faa6ae6d26c118d8fbb20d3d67cc5/community-containers/caddy/readme.md?plain=1#L19
```
szaimen: I am not sure about hardcoding a link to a specific commit here. Maybe we can simply describe in words here what to do and link to the normal readme of the community container instead?

szaimen (commented):
Hey, sorry for the late reply! One comment below:

community-containers/home-assistant/readme.md
alexkannes222: Because that way, it shows the exact line, as there isn’t a "permalink"/go to bulletpoint function in that, and in the md view, it shows that line directly. it has also been used elsewhere, that is how i found that exact line like that

alexkannes222: > Because that way, it shows the exact line, as there isn’t a "permalink"/go to bulletpoint function in that, and in the md view, it shows that line directly. it has also been used elsewhere, that is how i found that exact line like that

seems like i mistook something between issues and markdown files

alexkannes222: > I am not sure about hardcoding a link to a specific commit here. Maybe we can simply describe in words here what to do and link to the normal readme of the community container instead?

The thing is, that it would be redundant, to restate what already is said there.

szaimen (approved):
LGTM, thanks! :)

PR #7983, opened 2026-04-23
[Community Containers] Adding joplin-server

lonode:
- Needs https://github.com/szaimen/aio-caddy/pull/142

This PR adds [Joplin Server](https://hub.docker.com/r/joplin/server) as a community container.

Lots of users use Joplin with WebDAV configured on nextcloud, but joplin server is a dedicated tool made the authors that is superior in both benchmark and reliability. 

I have [wrapped ](https://github.com/lonode/joplin-server-standalone)their official docker image with an embedded postegreSQL database, as the community container mechanism of Nextcloud AIO does not allow for multi-container (docker compose type) setup.

PR #8091, opened 2026-05-13
Fix docker registry and community containers

lonode:
Changes : 

- Docker Hub registry : official images manifest are now correctly prepended with library/ . Detection of such official images is done by simply looking if "/" is present or not
- ContainerDefintionFetcher :  While definition of multiple container is possible, only the first one is truly picked up for download. This is fixed.

This has been tested and validated using the [Joplin Server PR](https://github.com/nextcloud/all-in-one/pull/7983)

### DIFF
--- a/php/src/Docker/DockerHubManager.php
+++ b/php/src/Docker/DockerHubManager.php
@@ -3,8 +3,6 @@ declare(strict_types=1);
 
 namespace AIO\Docker;
 
-use AIO\ContainerDefinitionFetcher;
-use AIO\Data\ConfigurationManager;
 use GuzzleHttp\Client;
 
 readonly class DockerHubManager {
@@ -15,6 +13,16 @@ readonly class DockerHubManager {
         $this->guzzleClient = new Client();
     }
 
+
+    // Official Docker Hub images need the library/ prefix when using the registry API directly.
+    private function NormalizeImageName(string $name): string {
+        if (!str_contains($name, '/')) {
+            return 'library/' . $name;
+        }
+        return $name;
+    }
+
+
     public function GetLatestDigestOfTag(string $name, string $tag) : ?string {
         $cacheKey = 'dockerhub-manifest-' . $name . $tag;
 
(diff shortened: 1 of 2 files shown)


php/src/Docker/DockerHubManager.php
```diff
@@ -24,19 +32,20 @@ public function GetLatestDigestOfTag(string $name, string $tag) : ?string {
         }
 
         // If one of the links below should ever become outdated, we can still upgrade the mastercontainer via the webinterface manually by opening '/api/docker/getwatchtower'
+        $normalizedName = $this->NormalizeImageName($name);
```
szaimen:
```suggestion
        // Normalize the name, important for docker hub library images
        $name = $this->NormalizeImageName($name);
```

szaimen (dismissed):
Hi, thanks a lot for the PR! :)

Just one suggestion below:

php/src/Docker/DockerHubManager.php
lonode: Fine by me ! But i guess then you want to replace all the next occurences of the variable :)

szaimen: Yes :)

lonode: Done !

php/src/Docker/DockerHubManager.php
```diff
@@ -15,6 +13,16 @@ public function __construct(
         $this->guzzleClient = new Client();
     }
 
+
+    // Official Docker Hub images need the library/ prefix when using the registry API directly.
+    private function NormalizeImageName(string $name): string {
```
pabzm: Please use a lower case first character for the function name.

(The function names starting with upper case are a legacy and shall be changed, too, one day.)

pabzm (changes requested):
Thank you for the contribution!

It looks good to me, I just have a minor nitpick.

php/src/Docker/DockerHubManager.php
lonode: Done.

pabzm (approved):
Looks good, thank you!

szaimen:

PR #8121, opened 2026-05-16
Remove HARP_HOST environment variable from manual-install YAML file

Fs00:
The `HARP_HOST` env variable seems to be an unintended leftover in the manual install YAML file, since all other HARP-related variables are removed. I can confirm that no issues occur when launching the container without the aforementioned variable.


szaimen (approved):
LGTM, thanks!

PR #8145, opened 2026-05-18
fix(ci): skip helm-release workflow on forks

### LINKED ISSUE #8128: Helm chart publication gha workflow should not run on forks by default
Today I pulled the main branch from this repo, and pushed it into my own fork.
After a while, I got an error notification from a failed gha workflow, which tried to publish the helm charts.
The error happened because in my own fork I do not have a `gh-pages` branch, which is where the helm charts get published to.

I find this error annoying, and think that there are many ways to fix it:
1. create a gh-pages branch in my repo
2. patch the `main` branch in my own repo and disable that workflow
3. send a PR which disables the running of that workflow for all forks

I'm not a fan of option 1, as I do not need to have those gh pages up, and also if I ever decide to have them I might want to have different content in them.

Option 2 seems hackish and prone to future merge conflicts / bad PRs.

I might send a PR for option 3, but I am not sure about all other forked repos owners would feel about this...

algojogacor:
## Summary of Changes

Skip the `helm-release` GitHub Actions workflow when running on forks. The workflow publishes Helm charts to the `gh-pages` branch, which doesn't exist on forks, causing failed workflow runs for contributors who sync their forks.

## Root Cause / What Was Fixed

The `helm-release.yml` workflow triggers on every push to `main` when files under `nextcloud-aio-helm-chart/**` change. On forks, the `helm/chart-releaser-action` fails because:
1. Forks don't have a `gh-pages` branch configured
2. The chart-releaser action tries to create a GitHub Release and publish to `gh-pages`

Added `if: github.repository == 'nextcloud/all-in-one'` condition to the `release` job so it only runs on the upstream repository.

## Testing Done

- Verified the workflow YAML syntax is valid
- The conditional check uses the standard GitHub Actions `github.repository` context, which is well-tested and reliable
- No other workflows are affected by this change

## Related Issue

Fixes #8128

### DIFF
--- a/.github/workflows/helm-release.yml
+++ b/.github/workflows/helm-release.yml
@@ -10,6 +10,7 @@ on:
 
 jobs:
   release:
+    if: github.repository == 'nextcloud/all-in-one'
     runs-on: ubuntu-latest
     steps:
       - name: Checkout


szaimen (approved):
LGTM, thanks!

szaimen:

PR #8148, opened 2026-05-18
aio-interface: make sure ublock does not break the log viewer

### LINKED ISSUE #8144: AIO log-display page fails silently when uBlock Origin is enabled in the browser
To reproduce:
- install and enable uBlock Origin in the browser, in its standard configuration
- open the AIO page on port 8080
- click on the link to display the logs of any running container

Result:
the page opens, but it stays blank (except for the control box in the top-right corner). No logs are shown, no error message is shown

How I figured out the error: opening the browser dev console, looking at error message stating "Loading failed for the <script> ..."

Suggested fixes:
1. rename (or alter?) log-view.js so that it does not get blocked by default by uBlock and similar tools
2. change the way the page initialises: make an error message appear if the js log-loading script fails to load

gggeek:
Fixes #8144


php/templates/log.twig
```diff
@@ -17,5 +18,6 @@
             </div>
         </div>
         <pre>{{ logContent }}</pre>
+        <script src="log-load.js?v1" onerror="document.getElementById('autoloading-box').textContent='Error loading the JS script. Check if your ad-blocker is interfering';"></script>
```
Zoey2936: this will break the CSP, which blocks all inline js and css

szaimen:
As addition to @Zoey2936's comment: @gggeek please fix DCO. We need this in order to be able to merge the PR. Thanks im advance! :)

gggeek:
@szaimen what is DCO?

php/templates/log.twig
gggeek: OK. I'll test with an 'alert', which might work before the document body has been rendered

gggeek: Change pushed

gggeek:
As a side note: I see that PR #8149 has landed, clearly AI-created, to fix the same bug, while this one was pending...

php/templates/log.twig
Zoey2936: The CSP will block all js which is not inside a .js file, so onerror will never work. The only way would be addEventListener or similiar in a always laoded js file which checks for this file load

szaimen:
> @szaimen what is DCO?

See https://github.com/nextcloud/all-in-one/pull/8148/checks?check_run_id=76526611631

php/templates/log.twig
gggeek: Ok, my bad for not having checked what the CSP was actually set to.

I will just go for doing without the error handling then - having an "always loaded" js script which checks for the loading of a 2nd one seems like an invitation to recursive failure. 

I might send a 2nd PR later, where the error text is the default in the page's html, and it gets changed to the non-error version by the loaded script.

Btw, it seems quite a strict CSP, if I may :-D

gggeek:
> https://github.com/nextcloud/all-in-one/pull/8148/checks?check_run_id=76526611631

Aha! One of the things which I did not find in either the CoC or develop.md.

I guess I will have to follow up with a PR to add that and the other bits of missing info to develop.md 🤣

php/templates/log.twig
gggeek: 3rd time a charm?

gggeek:
OK, DCO is fine (and hopefully all my future commits to AIO will be both signed-off and signed).

The Playwright test is now failing, but i think it might be for an unrelated cause...

Zoey2936 (approved):

gggeek:
The Playwright test re-run. As expected, all green now

szaimen (approved):
LGTM, thanks a lot! :)

szaimen:

PR #8149, opened 2026-05-18
fix: prevent uBlock Origin from blocking container log viewer

### LINKED ISSUE #8144: AIO log-display page fails silently when uBlock Origin is enabled in the browser

algojogacor:
## Summary

Renamed `log-view.js` to `container-log.js` to avoid being blocked by uBlock Origin and similar ad-blocking browser extensions. Added a fallback error message that detects when the script fails to load and guides users to disable ad-blocking extensions or use a different browser.

## Root Cause

uBlock Origin's default filter lists match script filenames containing patterns like `log-view`, causing the container log viewer page to render blank with no error message when an ad blocker is enabled. The user sees only an empty page with the control box.

## Changes

1. **Renamed** `php/public/log-view.js` → `php/public/container-log.js` — the filename `container-log.js` is less likely to match ad-blocker filter patterns.

2. **Updated** `php/templates/log.twig`:
   - Changed the `<script src>` reference to point to the new filename
   - Added an inline detection script that checks whether `LogViewer` is defined after a 2-second grace period
   - If the script failed to load (e.g., blocked by an extension), a red error banner appears inside the floating box explaining the likely cause and suggesting workarounds

## Testing

- Verified the file rename doesn't break the existing functionality (the file content is unchanged)
- The error detection script is passive — it only activates if `LogViewer` is undefined after DOM load, so normal operation is unaffected
- The error message appears within the existing `#floating-box` container, maintaining visual consistency

Fixes #8144

### DIFF
--- a/php/public/container-log.js
+++ b/php/public/container-log.js
@@ -1,0 +1,142 @@
+class LogViewer {
+    // Configure the interval in seconds for autoloading log data.
+    autoloadIntervalSec = 5;
+    // Set to true to see some debug log statements in the browser console.
+    debugLog = false;
+
+    // Don't touch these, please.
+    containerId;
+    apiBaseUrl = 'api/docker/logs';
+    autoloadIntervalId = null;
+    logElem;
+    lastLogTimestamp = '';
+    autoloadingDisabledFromButton = false;
+    loaderElem;
+    dataLoadingLock;
+
+    constructor() {
+        const id = document.body.dataset.containerId;
+        if (typeof(id) !== 'string' || !id.startsWith('nextcloud-aio-')) {
+            throw new Exception('Invalid container ID');
+        }
+        this.containerId = id;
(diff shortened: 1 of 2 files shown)


gggeek:
FYI, there is already PR #8148 pending to fix this issue...

PR #8161, opened 2026-05-19
aio-interface: do not cache the containers, logs and setup screen as it shows credentials

szaimen:
- Split-up of https://github.com/nextcloud/all-in-one/pull/8049


pabzm (commented):
Technically the code looks ok, but I don't know what the purpose of these change is, and thus can't comment on their appropriateness.

szaimen:

PR #8165, opened 2026-05-19
docs(reverse-proxy): document X-Forwarded-For requirement for notify_push

RobinAngele:
## What this PR does

Adds two small documentation additions to `reverse-proxy.md` explaining that reverse proxies must not strip the `X-Forwarded-For` header, because `notify_push` relies on it for its startup self-test.

### Context

When `notify_push` starts it performs a self-test: it makes a request through the Nextcloud URL and checks that the `X-Forwarded-For` header it sent arrives at Nextcloud unchanged. If a reverse proxy strips or overwrites this header (e.g. a hardened config using `proxy_set_header X-Forwarded-For ""`), the self-test fails silently. Clients then fall back to slow polling-based sync instead of real-time push notifications — a confusing symptom that's hard to diagnose.

This came up in #8108 (where I originally tried to add a workaround env var). @szaimen correctly pointed out that the right fix is to configure the reverse proxy properly, so this PR just documents that requirement.

### Changes

1. **Note in step 1** (before the proxy-specific configs): explains the header requirement and points users to ensure their proxy passes it through.
2. **New debug step in step 7**: tells users to run `docker logs nextcloud-aio-notify-push` and look for "self-test failed" if push notifications are slow.

The sample configs already do the right thing (Nginx has `proxy_set_header X-Forwarded-For $proxy_add_x_forwarded_for;`, Caddy passes it automatically) — this just makes the requirement explicit so users don't inadvertently strip it.

### DIFF
--- a/reverse-proxy.md
+++ b/reverse-proxy.md
@@ -188,6 +188,9 @@ The process to run Nextcloud AIO behind a reverse proxy has three required steps
 
     </details>
 
+> [!NOTE]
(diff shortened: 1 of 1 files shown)


Zoey2936:
I don't understand this change. All containers require the header to get the real ip, also all example configs correctly set this header. Like they also require other header, which we also don't mention directly

szaimen (changes requested):
See https://github.com/nextcloud/all-in-one/pull/8165#issuecomment-4490143519

RobinAngele:
Fair point — closing this. The sample configs already handle it correctly, and the real issue in my case was AIO running in normal mode behind an external nginx, which is better addressed by switching to reverse-proxy mode rather than by documenting header forwarding.

PR #8166, opened 2026-05-19
manual-upgrade.md: Move md files from CRLF to LF line endings; instruct to use :ro flag when mounting the docker socket

gggeek:


pabzm (changes requested):
@gggeek I can't tell what the actual text changes are, and thus can't review those. Could you split the line-ending changes of from the changes to the content, please?

gggeek:
It's 99% line ending changes, really... apart from that there are 2 cases of having added `:ro` to the example docker command lines args used to mount the docker daemon socket.

Sorry for having conflated 2 changes in a single PR.

But your request made me discover something new, and quite useful: you can append `?w=1` to the GH URLs that show PRs and have the diffs shown ignoring whitespace (which includes changes to line endings).

So https://github.com/nextcloud/all-in-one/pull/8166/changes?w=1 could maybe be enough for safe reviewing?

pabzm:
Yes, that helps, thank you for that hint!

I won't block this, but I'm wondering why we should introduce this change, since mounting the socket suggests an security advantage but actually gains almost none, which I don't like very much.

@gggeek Could you explain what your motivation or idea for this change is?

Zoey2936:
it is already done for the compose.yaml and I think also every else: https://github.com/nextcloud/all-in-one/blob/54b9817061a7b02fe5fe680bdae6b09c0654a405/compose.yaml#L10

szaimen (approved):
The change LGTM, thanks!

gggeek:
> mounting the socket RO suggests an security advantage but actually gains almost none, which I don't like very much.

I fully agree with that. It was just done like that in other places/docs, and it also seems to be the most common way to do it, in other blog posts and similar...

Btw, about this topic: I have started working on a new project dedicated exactly to that issue: https://github.com/gggeek/yet-another-docker-socket-proxy. It aims to be a more-flexible and more-accurate replacement for the docker-socket-proxy container based in HAProxy used within NC AIO...

PR #8176, opened 2026-05-20
Add AzuraCast community container

biguenique:
﻿## AzuraCast community container

This PR adds [AzuraCast](https://www.azuracast.com/) as a community container for Nextcloud AIO.

AzuraCast is a self-hosted, open source web radio management suite. It includes a web interface, Icecast streaming, a built-in SFTP server (SFTPGo), and automatic TLS.

### Features
- Web interface at `https://radio.$NC_DOMAIN` (via Caddy community container — see related PR)
- Radio streaming ports for up to 5 stations (Icecast: 8000-8040, source: 8005-8045, SHOUTcast: 8006-8046)
- SFTP uploads on port 2022 (SFTPGo built-in)
- Nextcloud file integration via External Storage (SFTP) — no bind-mounts required
- Dedicated shared volume (`/var/azuracast/shared`) for advanced storage scenarios
- Optional install volumes for SHOUTcast 2, Stereo Tool, RSAS, and GeoLite2

### Testing
Tested on Nextcloud AIO v12.9 with Caddy community container on Ubuntu 24.04.

### Related PR
Caddy integration: https://github.com/szaimen/aio-caddy/pull/144 (adds `radio.$NC_DOMAIN` reverse proxy with `flush_interval -1` for SSE/WebSocket)


community-containers/azuracast/azuracast.json
```diff
@@ -0,0 +1,108 @@
...
+                { "ip_binding": "", "port_number": "8020", "protocol": "tcp" },
+                { "ip_binding": "", "port_number": "8025", "protocol": "tcp" },
+                { "ip_binding": "", "port_number": "8026", "protocol": "tcp" },
+                { "ip_binding": "", "port_number": "8030", "protocol": "tcp" },
+                { "ip_binding": "", "port_number": "8035", "protocol": "tcp" },
+                { "ip_binding": "", "port_number": "8036", "protocol": "tcp" },
+                { "ip_binding": "", "port_number": "8040", "protocol": "tcp" },
+                { "ip_binding": "", "port_number": "8045", "protocol": "tcp" },
+                { "ip_binding": "", "port_number": "8046", "protocol": "tcp" }
+            ],
```
szaimen: Hi, with that amount of ports I wonder if it woulnd't be better to use the host network?

```suggestion
            "internal_port": "host",
            "restart": "unless-stopped",
```

community-containers/azuracast/azuracast.json
```diff
@@ -0,0 +1,108 @@
...
+                    "writeable": true
+                },
+                {
+                    "source": "nextcloud_aio_azuracast_rsas_install",
+                    "destination": "/var/azuracast/storage/rsas",
+                    "writeable": true
+                },
+                {
+                    "source": "nextcloud_aio_azuracast_geolite_install",
+                    "destination": "/var/azuracast/storage/geoip",
```
szaimen: Woulnd't it be easier to simply mount the parent folder instead?
```suggestion
                    "source": "nextcloud_aio_azuracast_storage",
                    "destination": "/var/azuracast/storage",
                    "writeable": true
                },
                {
                    "source": "nextcloud_aio_azuracast_backups",
                    "destination": "/var/azuracast/backups",
                    "writeable": true
                }
```

szaimen (commented):
Hi, thanks a lot for your contribution! :)

For now I have two suggestions below:

community-containers/azuracast/azuracast.json
biguenique: Uh. Interesting suggestion! That would instantly raise any limit of stations AzuraCast could host. 

But what about container isolation? Wouldn't that expose all open ports on `nextcloud-aio-azuracast` on the host — even those that AzuraCast didn't intend to expose?

Also, `AZURACAST_HTTP_PORT=8080` would bind directly on the host — conflicting with the AIO master container, already listening on port 8080. That would likely require choosing a different port for AzuraCast.

community-containers/azuracast/azuracast.json
biguenique: That would make perfect sense to me. Leaner is better. But then `nextcloud_aio_azuracast_uploads` should also be stripped form `backup_volumes` — which is fine: from what I get, it turns out `uploads` is only used to stores files temporarily. We should have excluded it from backups in the first place.

I'm on it.

community-containers/azuracast/azuracast.json
szaimen: You are right. Are there additional ports that are exposed by the container that should not be exposed on the host? 
Maybe the project has some docs on how to run the container in host network mode?

Also, AZURACAST_HTTP_PORT should be set to a different value anyway? Otherwise how would it be exposed on the host?

biguenique: I couldn't find anything about running AzuraCast in host network mode.

Here is the result I got from Nmap from within the AzuraCast container : 
```
# nmap -sT 127.0.0.1
Starting Nmap 7.95 ( https://nmap.org ) at 2026-05-22 01:10 EDT
Nmap scan report for localhost (127.0.0.1)
Host is up (0.00030s latency).
Not shown: 994 closed tcp ports (conn-refused)
PORT     STATE SERVICE
2022/tcp open  down
3306/tcp open  mysql
6025/tcp open  x11
8080/tcp open  http-proxy
8090/tcp open  opsmessaging
8443/tcp open  https-alt

Nmap done: 1 IP address (1 host up) scanned in 0.10 seconds
```

3306 (mysql) is a problem, 6025 (x11? maybe *SupervisorD*) might be. Would need to be firewalled on the host — an extra requirement that seems contrary to the philosophy of community containers (that they should be readily deployable without extra configurations). 8090 (opsmessaging) is a test station (internal Icecast port — see note below).

But my main concern is that Docker being the official deployment method for AzuraCast, it's designed to be insulated from the host, and any changes in the distribution might introduce vulnerabilities that were not foreseen because of this. So my take is that explicitly binding ports is the simplest *and* the safest bet.

Is there any example of a community container in host network mode?

As for the port 8080 — it's solely used by the http reverse proxy (Caddy) and doesn't need to be directly accessed. 8443 (https) is useless in this context. I must admit these arbitrary replacements for 80 and 443 were badly chosen, in retrospect.

The [official docs](https://www.azuracast.com/docs/administration/multi-site-installation/#edit-azuracast-web-serving-ports) suggest using `10080` and `10443` as port replacements for http and https. I could reconfigure the container to use these instead. That would require also committing changes accordingly to my PR on aio-caddy.
I would favor switching to theses ports to match the docs, if you agree — host network or not.

A note on the limited available stations ports: these ports are only used for direct http (no TLS) access to Icecast/Shoutcast/live source listener port (managed by Liquidsoap, I believe). For actual stream broadcasting, they're essentially useless here, since AzuraCast allows proxying the streams through the http-proxy port (secured through Caddy https). The settings for this are documented in the `readme.md`. This means you can still virtually run as many stations as you want through internal ports that get proxied through the web port (I tested it — that's port 8090 you see in the ports list above). You really only need the bindings for live remote source clients, and you can easily assign ports to stations that actually need them from within AzuraCast.

szaimen: All right, you convinced me that host network mode is probably not doable for this contianer. Lets leave it like that then. 

However lets also change AZURACAST_HTTP_PORT to something not used yet and also expose that directly on the server. See e.g. https://github.com/nextcloud/all-in-one/blob/2b1ec334e8cd6031ec994f5d005095791375d136/community-containers/vaultwarden/vaultwarden.json#L13-L15

biguenique: Good. I changed ports to 10080 and 10443 (as the AC docs example), both bound to the host.
I've committed here and accordingly to [aio-caddy#144](https://github.com/szaimen/aio-caddy/pull/144).

I'm not sure about the ` %APACHE_IP_BINDINGS%` bit though.
I put `"ip_binding": "%APACHE_IP_BINDINGS%"` for port 10080 (http proxied by Caddy) and left it empty for 10443.

I tested that both ports are accessible directly, but: 
- Direct access to ports using the main $NC_DOMAIN or subdomain.$NC_DOMAIN fails because of HSTS rules (that's expected)
- Direct access to port 10080 using IP (or some other domain) fails in the browser with `ERR_UNSAFE_PORT`
- Direct access to port 10443 (IP or other domain) warns about the invalid certificate (also expected)

I'll wait for a confirmation before resolving this proposal.
Edit : I'll actually mark it as resolved, and wait for an infirmation instead!

biguenique:
Commit d47feb5ea607e28218f39eef5c172b9f1e48c164 brings changes in shared external storage after extensive testing and much consideration.

**Shared storage: `nextcloud_aio_azuracast_shared` → `%NEXTCLOUD_MOUNT%`**
Replaced my original named volume workaround with the AIO-native `%NEXTCLOUD_MOUNT%` mechanism (same pattern as Jellyfin). The volume is silently skipped when `NEXTCLOUD_MOUNT` is not configured, so existing setups without it are unaffected.

https://github.com/nextcloud/all-in-one/blob/d47feb5ea607e28218f39eef5c172b9f1e48c164/community-containers/azuracast/azuracast.json#L74-L78

I considered `%NEXTCLOUD_DATADIR%` for a minute (as also used by Jellyfin) but ruled it out: it would give AzuraCast read access to all Nextcloud user data, which is inappropriate for a radio container.

The `readme.md` was updated accordingly.

https://github.com/nextcloud/all-in-one/blob/d47feb5ea607e28218f39eef5c172b9f1e48c164/community-containers/azuracast/readme.md#L78-L81

**File permissions: documented in readme**
With `NEXTCLOUD_MOUNT`, the two processes (Nextcloud as `www-data`/UID 33, AzuraCast as `azuracast`/UID 1000) share a directory with no group membership in common. The default behavior is:
- Files (`644`) are world-readable by both — no intervention needed for read access
- Some AzuraCast directories are created with mode `700`, blocking Nextcloud traversal

Write access and access to `700` directories require POSIX ACLs on the host. The readme now documents targeted `setfacl` commands for read-only and bidirectional read-write access, with a note that they must be re-run after AzuraCast creates new subdirectories.

**UMASK: no effect**
Testing on a live instance confirmed that the `UMASK` environment variable is ignored by the AzuraCast image — files are always created at `644`/`755` regardless. No change to the JSON is relevant here.

I'm now pretty confident about the current state of this PR.

community-containers/azuracast/azuracast.json
```diff
@@ -0,0 +1,87 @@
...
+                { "ip_binding": "", "port_number": "8035", "protocol": "tcp" },
+                { "ip_binding": "", "port_number": "8036", "protocol": "tcp" },
+                { "ip_binding": "", "port_number": "8040", "protocol": "tcp" },
+                { "ip_binding": "", "port_number": "8045", "protocol": "tcp" },
+                { "ip_binding": "", "port_number": "8046", "protocol": "tcp" }
+            ],
+            "environment": [
+                "TZ=%TIMEZONE%",
+                "AZURACAST_HTTP_PORT=10080",
+                "AZURACAST_HTTPS_PORT=10443",
```
szaimen, line 41: Ah I didnt know AZURACAST also comes with an HTTPS port. In that case the HTTP_PORT should not be exposed

community-containers/azuracast/azuracast.json
```diff
@@ -0,0 +1,87 @@
...
+                {
+                    "ip_binding": "%APACHE_IP_BINDING%",
+                    "port_number": "10080",
+                    "protocol": "tcp"
+                },
+                {
+                    "ip_binding": "",
+                    "port_number": "10443",
+                    "protocol": "tcp"
+                },
```
szaimen, line 16:
```suggestion
                {
                    "ip_binding": "%APACHE_IP_BINDING%",
                    "port_number": "10443",
                    "protocol": "tcp"
                },
```

community-containers/azuracast/azuracast.json
```diff
@@ -0,0 +1,87 @@
+{
+    "aio_services_v1": [
+        {
+            "container_name": "nextcloud-aio-azuracast",
+            "display_name": "AzuraCast",
+            "documentation": "https://github.com/nextcloud/all-in-one/tree/main/community-containers/azuracast",
+            "image": "ghcr.io/azuracast/azuracast",
+            "image_tag": "stable",
+            "internal_port": "10080",
```
szaimen, line 9:
```suggestion
            "internal_port": "10443",
```

community-containers/azuracast/azuracast.json
```diff
@@ -0,0 +1,87 @@
...
+                },
+                {
+                    "source": "%NEXTCLOUD_MOUNT%",
+                    "destination": "%NEXTCLOUD_MOUNT%",
+                    "writeable": true
+                }
+            ],
+            "backup_volumes": [
+                "nextcloud_aio_azuracast_stations",
+                "nextcloud_aio_azuracast_db",
```
szaimen, line 77: shouldnt the storage also be included?
```suggestion
                "nextcloud_aio_azuracast_storage",
                "nextcloud_aio_azuracast_db",
```

szaimen (commented):
Some remaining comments and I think then this should be ready to go :)

community-containers/azuracast/azuracast.json
biguenique: The storage volume only stores optional software and the geolite database that can easily be reinstalled if needed.
I see no need to bloat backups with it.

https://github.com/nextcloud/all-in-one/blob/d47feb5ea607e28218f39eef5c172b9f1e48c164/community-containers/azuracast/readme.md#L51

szaimen: all right, fine by me then

community-containers/azuracast/azuracast.json
biguenique: Not sure.
My PR on Caddy will reverse proxy the HTTP port 10080 on radio.{$NC_DOMAIN}. 
My guess is that the `internal_port` should remain 10080.
I suppose it doesn't need to be exposed on the host — but will it be reachable by Caddy then?

And now that I think of it, I'm not even sure how relevant it is to expose port 10443 at all, since all HTTPS traffic would go through the Caddy reverse proxy, and I see no easy way to issue valid SSL certificates from within AzuraCast anyway.

szaimen: > I suppose it doesn't need to be exposed on the host — but will it be reachable by Caddy then?

Yes it would

> And now that I think of it, I'm not even sure how relevant it is to expose port 10443 at all, since all HTTPS traffic would go through the Caddy reverse proxy, and I see no easy way to issue valid SSL certificates from within AzuraCast anyway.

I would still expose it

biguenique: Let's go then!

community-containers/azuracast/azuracast.json
biguenique: What about the internal port?

szaimen: Yeah, we can keep it like it is. Fine by me 👍

szaimen:
@biguenique thanks a lot! ☺️

Can you please also add it to https://github.com/nextcloud/all-in-one/blob/7e6571cfcc82826b35692bb373565bb7a0585be7/community-containers/readme.md?plain=1#L9-L60 I think then we can merge this if CI is green :)

community-containers/readme.md
```diff
@@ -40,6 +40,7 @@ flowchart TD
         DLNA(["📡  DLNA\nMedia Streaming"]):::community
         MEMTR(["📸  Memories\nVideo Transcoder"]):::community
         MKV(["💿  MakeMKV\nBlu-ray / DVD rip"]):::community
+        AZURA(["📻  AzuraCast\nRadio Station Server"]):::community
```
biguenique, line 43: Just did!

@szaimen Thanks to you! 🐰

community-containers/azuracast/azuracast.json
```diff
@@ -0,0 +1,83 @@
...
+                { "ip_binding": "", "port_number": "8020", "protocol": "tcp" },
+                { "ip_binding": "", "port_number": "8025", "protocol": "tcp" },
+                { "ip_binding": "", "port_number": "8026", "protocol": "tcp" },
+                { "ip_binding": "", "port_number": "8030", "protocol": "tcp" },
+                { "ip_binding": "", "port_number": "8035", "protocol": "tcp" },
+                { "ip_binding": "", "port_number": "8036", "protocol": "tcp" },
+                { "ip_binding": "", "port_number": "8040", "protocol": "tcp" },
+                { "ip_binding": "", "port_number": "8045", "protocol": "tcp" },
+                { "ip_binding": "", "port_number": "8046", "protocol": "tcp" }
+            ],
```
szaimen, line 37:
```suggestion
            ]
```

community-containers/azuracast/azuracast.json
```diff
@@ -0,0 +1,83 @@
...
+                { "ip_binding": "", "port_number": "8020", "protocol": "tcp" },
+                { "ip_binding": "", "port_number": "8025", "protocol": "tcp" },
+                { "ip_binding": "", "port_number": "8026", "protocol": "tcp" },
+                { "ip_binding": "", "port_number": "8030", "protocol": "tcp" },
+                { "ip_binding": "", "port_number": "8035", "protocol": "tcp" },
+                { "ip_binding": "", "port_number": "8036", "protocol": "tcp" },
+                { "ip_binding": "", "port_number": "8040", "protocol": "tcp" },
+                { "ip_binding": "", "port_number": "8045", "protocol": "tcp" },
+                { "ip_binding": "", "port_number": "8046", "protocol": "tcp" }
+            ]
```
szaimen:
```suggestion
            ],
```

community-containers/azuracast/azuracast.json
```diff
@@ -0,0 +1,83 @@
...
+            "container_name": "nextcloud-aio-azuracast",
+            "display_name": "AzuraCast",
+            "documentation": "https://github.com/nextcloud/all-in-one/tree/main/community-containers/azuracast",
+            "image": "ghcr.io/azuracast/azuracast",
+            "image_tag": "stable",
+            "internal_port": "10080",
+            "restart": "unless-stopped",
+            "ports": [
+                {
+                {
```
szaimen, line 12:
```suggestion
```

szaimen (approved):
LGTM. Thanks a lot! ☺️

biguenique:
Thanks @szaimen and everyone involved for creating opportunities to contribute and providing the actual support for it.
This PR is my very first time — in many regards!

szaimen:

PR #8178, opened 2026-05-20
Add top-level .editorconfig; improve app/.editorconfig

gggeek:
Implements feat. req. #8167

While at it, I modified one file which was using tabs for indentation and moved it to using spaces, which I deemed better than enforcing tab indentation for all xml files

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,12 +1,18 @@
 # https://editorconfig.org
 
-# note: the files in ./composer actually use 4 spaces instead of tabs
-
 root = true
 
 [*]
 charset = utf-8
 end_of_line = lf
-indent_style = tab
+indent_size = 4
+indent_style = space
 insert_final_newline = true
 trim_trailing_whitespace = true
+
+[*.yaml]
+indent_size = 2
+
+
+[*.yml]
+indent_size = 2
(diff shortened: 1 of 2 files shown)


.editorconfig
```diff
@@ -0,0 +1,20 @@
+# https://editorconfig.org
+
+# Tip: to find files violating the rules set out here, run `docker run --rm --volume=$PWD:/check mstruebing/editorconfig-checker`
```
szaimen, line 3: Should we add a workflow that checks this automatically?

szaimen (commented):
Thanks! One question below:

.editorconfig
gggeek: We "might" (TM) 😄 

Atm there are dozens of files failing, so first we would need a PR with massive reformatting.

After that, we have decide how anal we want to be about this.

In an open, inclusive project being too strict about formatting might be off-putting for contributors who do not use an IDE with built-in support for .editorconfig. One more hop to jump through.

If we otoh leave it as a non-blocking workflow, it just means more logs which no-one will read at best, and find an annoyance at worst...

szaimen (approved):
LGTM, thanks!

szaimen:

PR #8211, opened 2026-05-28
update tags of playwright workflows

szaimen:


php/README.md
```diff
@@ -1,5 +1,7 @@
 # PHP Docker Controller
 
+<!--test-->
```
szaimen: remove

pabzm (approved):
👍

php/README.md
szaimen: done

PR #7823, opened 2026-03-25
caddy: use acme shortlived profile

Zoey2936:
- change default acme profile to shortlived 
- Close #7776

PR #8235, opened 2026-06-01
revert: caddy: use acme shortlived profile

szaimen:
- Fixes https://help.nextcloud.com/t/aio-update-to-nextcloud-aio-v13-1-0-downgrades-lets-encrypt-certificates-to-one-week/245206
- reverts https://github.com/nextcloud/all-in-one/pull/7823
- reason: currently talk will show a warning regularly if the cert expires in a few days: https://github.com/nextcloud/all-in-one/pull/7823#issuecomment-4518082615
- Needs https://github.com/nextcloud/spreed/issues/17306

### DIFF
--- a/Containers/apache/Caddyfile
+++ b/Containers/apache/Caddyfile
@@ -78,7 +78,6 @@ http://{$APACHE_HOST}.nextcloud-aio:23973, # For Collabora callback and WOPI req
     # TLS options
     tls {
         issuer acme {
-            profile shortlived
             # Disable HTTP challenge because that would require port 80, which we don't get (it's exposed to the mastercontainer).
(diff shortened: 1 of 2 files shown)


Zoey2936:
I understamd the apache container, but why for the mastercontainer?
Also unsetting it mean using the classic profile. If only the lifetime is an Issue we just could switch to the tlsserver profile

szaimen:
> I understamd the apache container, but why for the mastercontainer?
> Also unsetting it mean using the classic profile. If only the lifetime is an Issue we just could switch to the tlsserver profile

Yeah you are right. I didnt think this more through. Let me adjust things again :)

szaimen:
Adjustments in https://github.com/nextcloud/all-in-one/pull/8237. @Zoey2936 can you please review? :)

PR #7995, opened 2026-04-24
add desec

szaimen:
- This PR is best reviewable commit by commit
- [x] The PR was tested and verified that it works locally
- [x] The PR was completely or partially created with AI

## Summary
- Integrates Desec into the AIO interface which simplifies the domain registration and domain setup.
- You can either create a new account directly or use an existing account. 
- You have a button that you can open where the process will be done. After the process is done, the whole page reloads and you can continue with the setup.
- Caddy and dnsmasq community containers will be activated during the process which will take over the cert generation and enable split-brain dns.
- Community containers will work automatically for example Vaultwarden as the subdomains of the desec domain point automatically via a wildcard dns entry towards the main desec domain.
- Playwright tests are added to ensure that the system works as expected in the future
- the desec api is mocked via a small node js based file.
- the qa and other docs were updated to mention desec

PR #8247, opened 2026-06-03
Run playwright tests via compose setup, locally and in CI

pabzm:
The playwright tests now are running via a docker compose setup, which can be executed via ./php/tests/run.sh locally, and also gets called from the github workflow.


pabzm:
Rebased to resolve conflicts

szaimen:
Playwright seems to fail unfortunately :/

pabzm:
@szaimen It works now, please have another look

szaimen:
Thanks a lot for the PR @pabzm :)

However I would postpone my review of this PR until we merged https://github.com/nextcloud/all-in-one/pull/7995 as this will conflict with this PR here and I would like to merge the desec PR as son as possible. Hope that is fine for you? :)

pabzm:
I'd be happy to not postpone it too long, since I already had to resolve conflicts twice. but for desec it's ok.

szaimen:
> I'd be happy to not postpone it too long, since I already had to resolve conflicts twice. but for desec it's ok.

All right! Will try to review it after merging desec. Thanks for all your effort!

szaimen:
This conflicts now after https://github.com/nextcloud/all-in-one/pull/7995 unfortunately but I can finally review it after another rebase :)

pabzm:
Rebased onto "main" and included the desec tests in the compose setup (which required some code cleanup)

pabzm:
I'll have a look at the CI failures tomorrow

php/src/Data/ConfigurationManager.php
```diff
@@ -231,15 +231,15 @@ class ConfigurationManager
      * stray env var redirecting it in production.
      */
     public string $desecApiBase {
-        get => $this->get('desec_api_base', 'https://desec.io/api/v1');
+        get => $this->getEnvironmentalVariableOrConfig('DESEC_API_BASE', 'desec_api_base', 'https://desec.io/api/v1');
```
szaimen: I am not sure about this change. This means anyone can set and adjust this. We should not expose such a testing config to the users imho.

php/src/Data/ConfigurationManager.php
```diff
@@ -231,15 +231,15 @@ class ConfigurationManager
...
     }
 
     /**
      * Base URL of the deSEC dynamic-DNS update endpoint. Configurable via the
      * 'desec_update_url' config key (configuration.json) only — see desecApiBase.
      */
     public string $desecUpdateUrl {
-        get => $this->get('desec_update_url', 'https://update.dedyn.io/');
+        get => $this->getEnvironmentalVariableOrConfig('DESEC_UPDATE_URL', 'desec_update_url', 'https://update.dedyn.io/');
```
szaimen: Same

php/tests/compose.yaml
```diff
@@ -0,0 +1,123 @@
...
+    init: true
+    restart: always
+    # links:
+    #   - desec-mock
+    ports:
+      - "8080:8080"
+    volumes:
+      - nextcloud_aio_mastercontainer:/mnt/docker-aio-config
+      - /var/run/docker.sock:/var/run/docker.sock:ro
+      - backup_vol:/mnt/test
```
szaimen: This is not how you configure backup volumes for aio

php/tests/compose.yaml
```diff
@@ -0,0 +1,123 @@
...
+    profiles:
+      - none
+    environment:
+      SKIP_DOMAIN_VALIDATION: ${SKIP_DOMAIN_VALIDATION-true}
+      APACHE_PORT: 11000
+      DESEC_API_BASE: http://desec-mock:8090/api/v1 
+      DESEC_UPDATE_URL: http://desec-mock:8090/update
+    entrypoint: bash /start.sh
```
szaimen: This now changes the default also for the on-workflow-dispatch trigger. This should only be set for the app-local-code run imho

php/tests/compose.yaml
```diff
@@ -0,0 +1,123 @@
...
+
+  app-code-from-image:
+    extends: app-base
+    container_name: nextcloud-aio-mastercontainer
```
szaimen, line 43: Container_name can be set in the base afaics?

php/tests/compose.yaml
```diff
@@ -0,0 +1,123 @@
...
+    volumes:
+      - ..:/var/www/docker-aio/php
+      - ../../Containers/mastercontainer/internal.Caddyfile:/internal.Caddyfile
+      - ../../Containers/mastercontainer/headers.Caddyfile:/headers.Caddyfile
+      - ../../Containers/mastercontainer/start.sh:/start.sh
+    profiles:
+      - local-code
+
+  desec-mock:
+    image: docker.io/library/node:latest
```
szaimen: Should we pin the version and update it regularly?

php/tests/compose.yaml
```diff
@@ -0,0 +1,123 @@
...
+    volumes:
+      - ..:/app/php
+    working_dir: /app
+    command: node php/tests/desec-mock.mjs 8090 2>&1
+          
+  test-runner-base:
+    image: mcr.microsoft.com/playwright:v1.56.1
```
szaimen: Should the image version be updated regularly?

php/tests/compose.yaml
```diff
@@ -0,0 +1,123 @@
...
+      BASE_URL: "https://nextcloud-aio-mastercontainer:8080"
+      DEBUG: "pw:api"
+      CI: "true"
+      DESEC_MOCK_URL: "http://desec-mock:8090"
+      #AIO_PASSWORD: correct horse battery staple aircraft
+    command: |-
+      bash -c "
+        cd tests
+        # Install dependencies
+        npm ci --omit=optional
```
szaimen: Shouldnt npm ci run in a different container?

szaimen (changes requested):
Thanks a lot! Some comments from my side already below

php/src/Data/ConfigurationManager.php
pabzm: I still don't understand your position and all other approaches I tried are too patchy or don't work. Let's talk about this again, please.

php/tests/compose.yaml
pabzm: Why not? Isn't that just the saving directory for the backup files? We don't need those files after the tests finished.
How *should* the backup volumes be configured in your opinion?

php/tests/compose.yaml
pabzm: I just introduced this because `/start.sh` isn't executable in the image for some reason. How is the entrypoint different for the on-workflow-dispatch trigger?

php/tests/compose.yaml
pabzm: I tried that, but it apparently gets overwritten by the service's name. So I'd keep it this way, since it doesn't hurt much.

php/tests/compose.yaml
szaimen: https://github.com/nextcloud/all-in-one#how-to-create-the-backup-volume-on-windows

php/tests/compose.yaml
szaimen: Hm this is weird. In that case the image would be broken so we should fix the image instead of overwriting this here

php/tests/compose.yaml
szaimen: Ok

php/src/Data/ConfigurationManager.php
szaimen: as discussed, I am fine to have this exposed as a variable if we hide this behind anoher env for example TESTING_INSTANCE=true or something alike.

php/tests/compose.yaml
pabzm: Do you mean a more concrete tag like `26.4.0` or even an image ID?

php/tests/compose.yaml
pabzm: Yes, probably, but the version needs to stay aligned with the version of [the referenced NPM package](https://github.com/nextcloud/all-in-one/blob/main/php/tests/package.json#L6).

php/tests/compose.yaml
szaimen: 26.4.0 would be fine from my pov

php/tests/compose.yaml
pabzm: Good idea, that reduces the `command` attributes, too!

php/tests/compose.yaml
szaimen: I guess we should add a workflow then that updates the version

php/tests/compose.yaml
```diff
@@ -68,35 +68,34 @@ services:
     working_dir: /app
     command: node php/tests/desec-mock.mjs 8090 2>&1
           
+  npm-installer:
+    image: docker.io/library/node:latest
```
szaimen: we should pin that version as well and update it regularly

php/tests/compose.yaml
pabzm: The image appears to be fine. I can start the service `app-base` without specifying the entrypoint, too. Just `app-local-code` runs into `exec /start.sh failed: Permission denied`. I'll investigate this.

php/src/Data/ConfigurationManager.php
szaimen: Another option would be to name it `TESTING__DESEC_API_BASE` for example to make it very clear that this is only meant for testing.

php/tests/compose.yaml
pabzm: FTR: we figured out in a verbal discussion that the volume isn't actually needed at all, so this is solved.

php/src/Data/ConfigurationManager.php
pabzm: Did that (with three underscores!)

php/tests/compose.yaml
pabzm: Fixed

php/tests/compose.yaml
pabzm: Done

php/tests/compose.yaml
pabzm: Done

php/tests/compose.yaml
pabzm: The file now references the tag `26`, but pins it to a concrete digest. A CI workflow checks daily for new digests for the tag and opens a pull request if it finds an update. This way we can use a "floating" tag but have the benefits of a pinned image, too.

szaimen:
@pabzm kaybe it would make sense to suqash before merging?

szaimen (approved):
Generally LGTM

pabzm:
I'll clean up the commit history once the CI is green and I had a last look on everything.

pabzm:
Cleaned up commit history

PR #8279, opened 2026-06-08
Feat/eurooffice default

moodyjmz:


Containers/apache/Caddyfile
```diff
@@ -54,7 +54,8 @@ http://{$APACHE_HOST}.nextcloud-aio:23973, # For Collabora callback and WOPI req
     route /eurooffice/* {
         uri strip_prefix /eurooffice
         reverse_proxy {$EUROOFFICE_HOST}:80 {
-            header_up X-Forwarded-Host {http.request.hostport}/eurooffice
+            header_up X-Forwarded-Host {http.request.hostport}
+            header_up X-Forwarded-Prefix /eurooffice
```
Zoey2936, line 58: can you link the commit where this header was added? I can't find it

moodyjmz: The original `header_up X-Forwarded-Host {http.request.hostport}/eurooffice` was introduced in https://github.com/nextcloud/all-in-one/commit/6375921282200d4c0a80233494c76b800b7d3859 (PR #8052), copied from the OnlyOffice block. That pattern doesn't actually work for EuroOffice though — the server doesn't read a path-suffixed `X-Forwarded-Host`.

EuroOffice nginx reads `X-Forwarded-Prefix` as a dedicated header to know its subpath. Support was added by @rikled in https://github.com/euro-office/documentserver/commit/5972e51 (`http-common.conf`):

```nginx
map $http_x_forwarded_prefix $the_prefix {
    default $http_x_forwarded_prefix;
}
```

`$the_prefix` is then used in `ds-docservice.conf` to construct redirect URLs:

```nginx
location = / { return 302 $the_scheme://$the_host$the_prefix/welcome/; }
return 302 $the_scheme://$the_host$the_prefix/<version>-...$cache...;
```

Without `X-Forwarded-Prefix`, `$the_prefix` is empty and the browser fetches `/sdkjs/...` instead of `/eurooffice/sdkjs/...`, which Caddy routes to Nextcloud → 404.

Zoey2936: https://github.com/euro-office/documentserver/blob/main/develop/setup/eo-proxy.conf returns 404

szaimen: https://github.com/Euro-Office/DocumentServer/issues/121

Zoey2936: then the `header_up X-Forwarded-Host {http.request.hostport}` line should be removed since it is the default in caddy: https://caddyserver.com/docs/caddyfile/directives/reverse_proxy#defaults

szaimen: Hm. Honestly I am still not convinced yet like you. Hopefully @moodyjmz can point us at the correct code in their repo.

szaimen:
Btw @moodyjmz I see 3 changes in this PR. Can you please split them up into separate PRs? Thanks in advance!

One for the caddyfile change, one for the entrypoint.sh change, one for the migration logic.

Containers/apache/Caddyfile
moodyjmz: Sorry for the confusion — I had a local file that documented the `X-Forwarded-Prefix` pattern but it was never pushed upstream.

The header support lives in [`Euro-Office/document-server-package` — `http-common.conf.m4` L35](https://github.com/Euro-Office/document-server-package/blob/main/common/documentserver/nginx/includes/http-common.conf.m4#L35), inherited from upstream ONLYOFFICE via https://github.com/Euro-Office/document-server-package/commit/294f9d6.

And I think you're right about the `X-Forwarded-Host` line — it's redundant since Caddy sets it by default. The block should be:

```
route /eurooffice/* {
    uri strip_prefix /eurooffice
    reverse_proxy {$EUROOFFICE_HOST}:80 {
        header_up X-Forwarded-Prefix /eurooffice
    }
}
```

moodyjmz: The `X-Forwarded-Prefix` support is in [`Euro-Office/document-server-package` — `http-common.conf.m4` L35](https://github.com/Euro-Office/document-server-package/blob/main/common/documentserver/nginx/includes/http-common.conf.m4#L35) — EuroOffice maps it to `$the_prefix` which is used to construct internal redirect URLs under the subpath.

moodyjmz:
> Btw @moodyjmz I see 3 changes in this PR. Can you please split them up into separate PRs? Thanks in advance!
> 
> One for the caddyfile change, one for the entrypoint.sh change, one for the migration logic.

Fair point, I was looking to get EO working with AIO and not paying enough attention to the different concerns my changes were cutting across. I will split into those PRs.

PR #8288, opened 2026-06-09
fix(nextcloud): register EuroOffice preview provider

moodyjmz:

---

## Summary

Registers the EuroOffice preview provider in `enabledPreviewProviders` so that office document thumbnails are generated correctly.

**Change:**
- Adds `OCA\Eurooffice\Preview` at index 50 to avoid colliding with AIO's seeded indices (1–7, 23). The check is idempotent.

## Test plan

- [ ] Office document thumbnails appear in the Files list
- [ ] Office document thumbnails appear in the recommendation widget
- [ ] Re-running the entrypoint is a no-op (idempotency check)

## Related PRs

- **#8290** — Caddyfile `X-Forwarded-Prefix` header fix
- **#8289** — Make EuroOffice the default editor for new installs

### DIFF
--- a/Containers/nextcloud/entrypoint.sh
+++ b/Containers/nextcloud/entrypoint.sh
@@ -908,6 +908,12 @@ if [ "$EUROOFFICE_ENABLED" = 'yes' ]; then
         fi
 
         php /var/www/html/occ config:app:set eurooffice DocumentServerUrl --value="https://$EUROOFFICE_HOST"
+
+        # Register EuroOffice preview provider in the explicit allowlist.
+        # Use a high fixed index (50) to avoid colliding with AIO's seeded indices (1-7, 23).
+        if ! php /var/www/html/occ config:system:get enabledPreviewProviders | grep -q "Eurooffice"; then
+            php /var/www/html/occ config:system:set enabledPreviewProviders 50 --value="OCA\Eurooffice\Preview"
+        fi
     fi
 else
     # Remove EuroOffice app if disabled and removal is requested
--- a/Containers/nextcloud/entrypoint.sh
+++ b/Containers/nextcloud/entrypoint.sh
@@ -912,7 +912,7 @@ if [ "$EUROOFFICE_ENABLED" = 'yes' ]; then
         # Register EuroOffice preview provider in the explicit allowlist.
         # Use a high fixed index (50) to avoid colliding with AIO's seeded indices (1-7, 23).
         if ! php /var/www/html/occ config:system:get enabledPreviewProviders | grep -q "Eurooffice"; then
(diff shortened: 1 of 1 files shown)


Containers/nextcloud/entrypoint.sh
```diff
@@ -903,11 +903,24 @@ if [ "$EUROOFFICE_ENABLED" = 'yes' ]; then
 
         # Adjust the EuroOffice host if using internal pattern
         if echo "$EUROOFFICE_HOST" | grep -q "nextcloud-.*-eurooffice"; then
+            # Set internal URLs for server-to-server calls via Docker network,
+            # bypassing the public domain which may not resolve inside containers.
```
szaimen: Actually this is a key requirement for AIO (that the domain resolves also inside the containers). We even check that this works during the initialy setup. So I am not sure about this change. Does it come with other benefits than that?

szaimen (changes requested):

Containers/nextcloud/entrypoint.sh
moodyjmz: You're right — these settings were added during local testing where DNS wasn't resolving inside containers. I didn't fully understand the setup and thought it was connected to how EuroOffice communicates internally. Dropping `DocumentServerInternalUrl` and `StorageUrl`; keeping only the preview provider registration.

szaimen:
Thanks @moodyjmz! Can you please squash your changes and rename the title and descirption accordingly or create a new PR? Thanks! :)

moodyjmz:
Thanks for the review! Squashed, updated the title and description, and rebased onto main.

Containers/nextcloud/entrypoint.sh
```diff
@@ -908,6 +908,12 @@ if [ "$EUROOFFICE_ENABLED" = 'yes' ]; then
         fi
 
         php /var/www/html/occ config:app:set eurooffice DocumentServerUrl --value="https://$EUROOFFICE_HOST"
+
+        # Register EuroOffice preview provider in the explicit allowlist.
+        # Use a high fixed index (50) to avoid colliding with AIO's seeded indices (1-7, 23).
+        if ! php /var/www/html/occ config:system:get enabledPreviewProviders | grep -q "Eurooffice"; then
+            php /var/www/html/occ config:system:set enabledPreviewProviders 50 --value="OCA\Eurooffice\Preview"
```
szaimen: I'd say one higher than 23 should be safe enough :)
```suggestion
            php /var/www/html/occ config:system:set enabledPreviewProviders 24 --value="OCA\Eurooffice\Preview"
```

Containers/nextcloud/entrypoint.sh
moodyjmz: No worries :)

szaimen (approved):
LGTM, thanks! :)

szaimen:

PR #8289, opened 2026-06-09
feat: enable EuroOffice by default for new installs

moodyjmz:

---

## Summary

~~Makes EuroOffice the default office suite for AIO, replacing Collabora as the out-of-the-box choice. Includes a one-time migration that runs on mastercontainer startup to update existing installs.~~
Sets EuroOffice as the default enabled office suite for new AIO installs.

**Changes:**

- `ConfigurationManager.php`: flips defaults — EuroOffice `true`, ~~Collabora `false`, adds `eurooffice` to the default installed-apps list~~
- ~~`php/public/index.php`: calls `performMigrations()` on startup~~
- ~~`ConfigurationManager::performMigrations()`: a guarded one-time migration (`eurooffice_default_migration_v1`) that enables EuroOffice and disables Collabora/OnlyOffice on existing AIO instances that haven't explicitly chosen an office suite~~
- ~~`readme.md`: updates description to reflect EuroOffice as default~~

~~**Migration safety:** The migration is gated by a config flag. It runs exactly once per instance. If a user has already explicitly enabled Collabora or OnlyOffice, the flag will not yet be set, so the migration will run — but they can re-enable their preferred suite immediately afterwards from the AIO interface.~~

## Test plan

- [ ] Fresh AIO install: EuroOffice is enabled by default
- ~~[ ] Existing AIO install with Collabora: after mastercontainer restart, migration runs once and switches to EuroOffice; Collabora can be re-enabled from settings~~
- ~~[ ] Running migration a second time is a no-op (flag already set)~~
- ~~[ ] `eurooffice` appears in the default installed apps list~~

## Related PRs

- **#8290** — Caddyfile `X-Forwarded-Prefix` header fix
- **#8288** — Register EuroOffice preview provider
- **This PR (#8289)** — ~~Make EuroOffice the default editor for new and existing installs~~ Enable EuroOffice by default for new installs

### DIFF
--- a/php/src/Data/ConfigurationManager.php
+++ b/php/src/Data/ConfigurationManager.php
@@ -100,13 +100,13 @@ class ConfigurationManager
     }
 
     public bool $isEuroofficeEnabled {
-        get => $this->get('isEuroofficeEnabled', false);
+        get => $this->get('isEuroofficeEnabled', true);
         set { $this->set('isEuroofficeEnabled', $value); }
     }
 
     public bool $isCollaboraEnabled {
         // Type-cast because old configs could have 1/0 for this key.
-        get => (bool) $this->get('isCollaboraEnabled', true);
+        get => (bool) $this->get('isCollaboraEnabled', false);
         set { $this->set('isCollaboraEnabled', $value); }
     }
 


php/src/Data/ConfigurationManager.php
```diff
@@ -100,13 +100,14 @@ class ConfigurationManager
     }
 
     public bool $isEuroofficeEnabled {
-        get => $this->get('isEuroofficeEnabled', false);
+        // Type-cast because old configs could have 1/0 for this key.
+        get => (bool) $this->get('isEuroofficeEnabled', true);
```
szaimen:
```suggestion
        get => $this->get('isEuroofficeEnabled', true);
```

php/src/Data/ConfigurationManager.php
```diff
@@ -927,7 +928,16 @@ public function getNextcloudStartupApps() : string {
         if (is_string($apps)) {
             return trim($apps);
         }
-        return 'deck twofactor_totp tasks calendar contacts notes';
+        return 'deck twofactor_totp tasks calendar contacts notes eurooffice';
+    }
+
+    public function performMigrations(): void {
```
szaimen: I am against such a migration. Please remove this

php/public/index.php
```diff
@@ -25,6 +25,7 @@
 require __DIR__ . '/../vendor/autoload.php';
 
 $container = \AIO\DependencyInjection::GetContainer();
+$container->get(\AIO\Data\ConfigurationManager::class)->performMigrations();
```
szaimen: please remove this

szaimen (changes requested):
Hi, thanks!

A few comments below

php/src/Data/ConfigurationManager.php
```diff
@@ -927,7 +928,16 @@ public function getNextcloudStartupApps() : string {
         if (is_string($apps)) {
             return trim($apps);
         }
-        return 'deck twofactor_totp tasks calendar contacts notes';
+        return 'deck twofactor_totp tasks calendar contacts notes eurooffice';
```
szaimen: This change is also not needed. It will be installed anyway.
```suggestion
        return 'deck twofactor_totp tasks calendar contacts notes';
```

szaimen:
Thank you @moodyjmz! :)

Can you please squash and adjust the PR description? Then I guess we can merge this :)

moodyjmz:
Thanks for the review! All addressed — removed the migration and the index.php call, reverted the startup apps change, and dropped the cast (no legacy configs to worry about here). Also pre-emptively squashed and updated the PR title.

szaimen (approved):
LGTM, thanks! :)

szaimen:

PR #8315, opened 2026-06-12
allow install_latest_major to be set to "no" 

tobiasKaminsky:
Previously the mapping was just $this->installLatestMajor ? 'yes' : '', so a stored value of 'no' was truthy and would have been turned into INSTALL_LATEST_MAJOR=yes in the Nextcloud container — the opposite of what the user intended. Now both ''

  and 'no' result in an empty env value, which the entrypoint's [ "$INSTALL_LATEST_MAJOR" = yes ] check treats as disabled, while any other value (e.g. the '34' set by the start button or upgrade button) still maps to yes.

  Assisted-by: Claude


tobiasKaminsky:
Manually tested:
"no" -> ""
"34" -> "yes"
"abc" -> "yes"
""  -> ""

szaimen (approved):
LGTM, thanks! :)

szaimen:

PR #8326, opened 2026-06-16
DockerActionManager: Fix fetching logs initially on podman

pabzm:
Podman doesn't like an empty `since` URL parameter, so the initial log fetching fails.

This change sends the parameter only if it has a non-empty value and thus fixes that problem.


szaimen (approved):

szaimen:

PR #8379, opened 2026-06-24
apache: re-introduce `X-Forwarded-Proto https` so that onlyoffice and eurooffice always work no matter the actual reverse proxy config in front of it

szaimen:

### DIFF
--- a/Containers/apache/Caddyfile
+++ b/Containers/apache/Caddyfile
@@ -55,7 +55,6 @@ http://{$APACHE_HOST}.nextcloud-aio:23973, # For Collabora callback and WOPI req
     route /eurooffice/* {
         uri strip_prefix /eurooffice
         reverse_proxy {$EUROOFFICE_HOST}:80 {
-            header_up X-Forwarded-Host {http.request.hostport}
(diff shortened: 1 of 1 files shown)


Containers/apache/Caddyfile
```diff
@@ -47,14 +47,17 @@ http://{$APACHE_HOST}.nextcloud-aio:23973, # For Collabora callback and WOPI req
...
             header_up X-Forwarded-Host {http.request.hostport}/onlyoffice
+            header_up X-Forwarded-Proto https
         }
     }
 
     # EuroOffice
     route /eurooffice/* {
         uri strip_prefix /eurooffice
         reverse_proxy {$EUROOFFICE_HOST}:80 {
+            header_up X-Forwarded-Host {http.request.hostport}
```
Zoey2936: this line should still not be needed

szaimen: you are right
```suggestion
```

szaimen:
This is now released with v13.3.1 Beta. Testing and feedback is welcome! See https://github.com/nextcloud/all-in-one#how-to-switch-the-channel

PR #8391, opened 2026-06-26
use `--sensitive` option for configuring tokens, secrets and passwords via occ

### LINKED ISSUE #8390: Use occ sensitive setting for encryption keys
See https://github.com/nextcloud/server/pull/61608

szaimen:
- Resolves: https://github.com/nextcloud/all-in-one/issues/8390

## Summary

## AI (if applicable)


pabzm (approved):
From the code it looks good to me (didn't test it since you didn't test, yet, neither).

I'm wondering if there's any code that reads those settings and needs the flag to its `occ` command, too? But if that's not the case I see no reason not to merge.

(The title says `--secret` while the code uses `--sensitive` – I guess the code is right?)

szaimen:
Ok seems like I should have tested this. I'll close this as it does not work iin its current form.

Issue #8414, opened 2026-07-03
AIO Restore - interrupted by new backup starting


steveroot:
**Short version**:
I’m a silly billy! I deleted my Nextcloud disk when migrating the VM to a new server.

Not a problem, I can restore from the offsite 4TB backup… which should take a couple of days.

It restores enough for nextcloud to work but files are missing. It stops the restore overnight.

I think it may be the next backup routine is started before the restore has finished by I’m not certain.

Originally shared in the nextcloud forum and I was asked to post this as a bug too

I haven’t found any other references to restores failing overnight and when I first started my test restore worked fine but it was over a local network and a much smaller set of files.

**Long version with more details:**

I’m hoping my solution will be to extract the backup from borg on the remote server (`borg export-tar /path_to/borgnextcloud::20260630_230253-nextcloud-aio recover_from_this.tar.gz`) , then directly access the volumes to access my files and photos, and copy them back into Nextcloud.

I also have a point in time backup of the VM which I could recover but it’s about 30 days old. At this point I’m wishing the Borg backup could send to two destinations as then I’d have and an on site and off site copy :smiley:

_The Basics_
Nextcloud Server version (e.g., 29.x.x):
33.0.6 AIO 13.3.1
_Operating system and version (e.g., Ubuntu 24.04):_
26.04 (nb: backup was probably taken on a 24.04 install)
_Web server and version (e.g, Apache 2.4.25):_
AIO
_Reverse proxy and version _(e.g. nginx 1.27.2)_
AIO
_PHP version (e.g, 8.3):_
AIO
_Is this the first time you’ve seen this error? (Yes / No):_
Yes
_When did this problem seem to first start?_
When I was an idiot and pressed delete disk
_Installation method (e.g. AlO, NCP, Bare Metal/Archive, etc.)_
AIO
_Are you using CloudfIare, mod_security, or similar? (Yes / No)_
No
_Summary of the issue you are facing:_
Nextcloud AIO restore appears to completes but it is incomplete. Enough of the files were present to make me think it had worked quicker than I expected but when looking for files I noticed and folders they’d appear to be there for a few seconds then vanish. I suspect the database expected them to be there, then found they were not so updated.

It appears to stop overnight around 23:50 hours (GMT+1)

_Steps to replicate it (hint: details matter!):_
Delete a large disk

Recover large data set using AIO restore, where recovery is expected to be greater than 24hours

_Log entries_
This is from the AIO Interface, ‘logs’ link

[...]

PR #8421, opened 2026-07-05
Doc update - borg restore for long restore processes

steveroot:
Long running restores can be interrupted by new backups starting.

- Resolves: # nothing
- Mitigates # https://github.com/nextcloud/all-in-one/issues/8414

## Summary
- [ n/a - it's a doc update] The PR was tested and verified that it works locally
- [ ] [Sign-off message](https://github.com/src-d/guide/blob/master/developer-community/fix-DCO.md) is added to all commits
- [ n/a ] Tests (playwright if possible) are included
- [ n/a] Screenshots before/after for front-end changes
- [Y ] Documentation has been updated or is not required
- [ n/a] [Milestone next added](https://github.com/nextcloud/all-in-one/milestones)

## AI (if applicable)

- [ No, all my own work] The content of this PR was partly or fully generated using AI

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -344,6 +344,7 @@ https://your-domain-that-points-to-this-server.tld:8443
     - [How to adjust borgs retention policy?](#how-to-adjust-borgs-retention-policy)
     - [How to migrate from AIO to AIO?](#how-to-migrate-from-aio-to-aio)
     - [Are remote borg backups supported?](#are-remote-borg-backups-supported)
+    - [Restoring a borg backup](#Restoring-a-borg-backup)
     - [Failure of the backup container in LXC containers](#failure-of-the-backup-container-in-lxc-containers)
     - [How to create the backup volume on Windows?](#how-to-create-the-backup-volume-on-windows)
     - [Pro-tip: Backup archives access](#pro-tip-backup-archives-access)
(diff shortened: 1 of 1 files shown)


steveroot:
I'm new to this, I'm not sure how to add labels. 
Probably got something else wrong too - I hope it's useful!

readme.md
```diff
@@ -1048,6 +1049,12 @@ Some alternatives, which do not have all the above benefits:
 - Here is another one that utilizes borgmatic and BorgBase for remote backups: https://github.com/nextcloud/all-in-one/discussions/4391
 - create your own backup solution using a script and borg, borgmatic or any other to backup tool for backing up to a remote target (make sure to stop and start the AIO containers correctly following https://github.com/nextcloud/all-in-one#how-to-enable-automatic-updates-without-creating-a-backup-beforehand)
 
+---
+### Restoring a borg backup ###
+Beware that a long running restore process may be interuprted by the start of a new backup process.
+See [https://github.com/nextcloud/all-in-one/blob/main/docker-ipv6-support.md](https://github.com/nextcloud/all-in-one/issues/8414))
```
pabzm, line 1055: Showing one URL but linking to another smells a little bit strange. Is that on purpose? If so, why?

readme.md
```diff
@@ -1048,6 +1049,12 @@ Some alternatives, which do not have all the above benefits:
 - Here is another one that utilizes borgmatic and BorgBase for remote backups: https://github.com/nextcloud/all-in-one/discussions/4391
 - create your own backup solution using a script and borg, borgmatic or any other to backup tool for backing up to a remote target (make sure to stop and start the AIO containers correctly following https://github.com/nextcloud/all-in-one#how-to-enable-automatic-updates-without-creating-a-backup-beforehand)
 
+---
+### Restoring a borg backup ###
+Beware that a long running restore process may be interuprted by the start of a new backup process.
```
pabzm, line 1054: There's a typo here:


```suggestion
Beware that a long running restore process may be interrupted by the start of a new backup process.
```

pabzm (changes requested):
Thank you very much for the contribution, it definitely is useful!

I've got two comments, please have a look at them.

readme.md
steveroot: Not on purpose, that would be me struggling to work out how to do this :-)
Thanks for spotting, leave it with me and I'll resubmit and learn from doing it.
Thanks for reviewing

szaimen (changes requested):
Instead of documenting the workaround we should rather try to fix the underlying problem imho.

steveroot:
I'm on holiday next week so plan to fix the pull request and resubmit.
I agree on the fix rather than just document work around, but that's probably beyond my skill set right now :-)

szaimen:
> I'm on holiday next week so plan to fix the pull request and resubmit. I agree on the fix rather than just document work around, but that's probably beyond my skill set right now :-)

All good! There is an open PR for this now here: https://github.com/nextcloud/all-in-one/pull/8514

Issue #8445, opened 2026-07-13
Hardware Acceleration: Allow to select which GPUs shall be used (by "memories" container)


ZambeiCloud:
I recently changed my server hardware and now I have two GPUs running with Ubuntu 26.04. The Intel iGPU and a dedicated Nvidia GPU. I'd like to use the iGPU for hardware acceleration for the memories community container through /dev/dri and noticed a strange behaviour. Sometimes after a reboot, the hardware acceleration doesn't work. I first narrowed the problem down to the fact that it's possible that after a reboot the Nvidia GPU can be the renderD128 device. So I came up with the idea to explicity map the iGPU into the AIO mastercontainer by path. I attached a screenshot from my docker compose file. This works perfect, as I only can see one device in the mastercontainer. But after the last reboot, I had again problems with the hardware acceleration. The kernel mounted the Nvidia GPU on the host to renderD128. The mastercontainer still had the iGPU on renderD128 because of the explicit mapping but the memories container had both GPUs and renderD128 was the Nvidia GPU.
It can be that the issue lies within the memories container but because AIO does the heavy lifting stuff, I assume the issue is on the AIO side.

### Steps to reproduce
1. Use hardware with more than 1 GPU and set up instance with a container which can use hardware acceleration
2. Map only one GPU into the mastercontainer
3. Check /dev/dri on the mastercontainer and the childcontainer

### Expected behavior 
The childcontainer (memories in this case) only has the explicit mapped /dev/dri device 

### Actual behavior 
The childcontainer got all /dev/dri devices

#### Other valuable info 

<img width="921" height="827" alt="Image" src="https://github.com/user-attachments/assets/e205ab57-0a22-48e0-ab0b-a4a8fb9d3c5e" />

<img width="921" height="824" alt="Image" src="https://github.com/user-attachments/assets/2f4b9e3b-9036-43b9-ba69-6f084df7d5c9" />

<img width="920" height="711" alt="Image" src="https://github.com/user-attachments/assets/9439b5ee-139e-433b-ab58-5ad6724f929e" />

pabzm:
The presence of `/dev/dri` is only used as a indicator to enable the checks for file system permissions, etc. Mounting only specific devices is not supported by AIO currently. (E.g. [the "memories" community container is mounting the whole `/dev/dri` unconditionally](https://github.com/nextcloud/all-in-one/blob/main/community-containers/memories/memories.json#L28).

[...]

PR #8471, opened 2026-07-20
feat(dri): allow selecting a specific /dev/dri render node via NEXTCLOUD_DRI_DEVICE

ecgang:
On multi-GPU hosts the kernel can assign the render node number (e.g. renderD128) to a different GPU across reboots, so passing the whole /dev/dri directory into containers can make hardware transcoding target the wrong GPU (issue https://github.com/nextcloud/all-in-one/discussions/8505).

This adds a NEXTCLOUD_DRI_DEVICE env var to pin a single render node:
- Unset (default): unchanged — the whole /dev/dri is passed through.
- Set to e.g. /dev/dri/renderD129: only that node is passed, exposed inside the container at the canonical /dev/dri/renderD128 so VA-API / go-vod work regardless of host numbering.
- Invalid/missing: logs a warning and falls back to the whole /dev/dri.

Render-group GID detection in the mastercontainer start script honours the selection too.

Scope note: this lives in the shared device-mount path, so it applies to every container that declares /dev/dri (nextcloud, talk-recording, memories, jellyfin, plex, local-ai), not just memories — intentional, since a user pinning a GPU wants it used everywhere. The renderD128 remap is deliberate so no in-container reconfig is needed; I'm open to a transparent same-path mount instead if you'd prefer.

- Resolves:  https://github.com/nextcloud/all-in-one/discussions/8505

## Summary

Testing done: composer psalm and composer lint are clean. I verified the input validation rejects path traversal, injection attempts, a trailing newline, and wrong devices (all fall back to /dev/dri), and checked the start.sh node-selection + GID logic on Alpine/busybox (the mastercontainer runtime) with a simulated multi-GPU /dev/dri. I don't have multi-GPU hardware to test end-to-end — could someone with that setup (e.g. @ZambeiCloud) confirm on real hardware?

## AI (if applicable)

-  [x]  The content of this PR was partly or fully generated using AI
  I used Claude Code Opus 4.8, and personally reviewed the diff, ran checks, and made the design calls


szaimen (changes requested):
Hi, before continuing with this, can you please clarify why you would need such a capability?

ecgang:
Hey @szaimen, the case is in the original report (#8445 / now #8505): on a multi-GPU host (Intel iGPU + dedicated Nvidia, in @ZambeiCloud's setup) the kernel doesn't guarantee which GPU lands on renderD128 across reboots. AIO mounts the whole /dev/dri directory into containers like memories, so there's no way today to pin hardware transcoding to one GPU. As a result, after a reboot it can silently switch to the wrong device. That's what broke for @ZambeiCloud in production.

NEXTCLOUD_DRI_DEVICE pins a single render node and remaps it to the canonical /dev/dri/renderD128 inside the container, so transcoding stays on the chosen GPU regardless of host renumbering. Unset, behavior is identical to today.

This is the PR @pabzm asked for on the issue (votes + implementation). Happy to adjust the approach if you'd rather mount at the real device path instead of remapping.

szaimen:
> Hey @szaimen, the case is in the original report (#8445 / now #8505): on a multi-GPU host (Intel iGPU + dedicated Nvidia, in @ZambeiCloud's setup) the kernel doesn't guarantee which GPU lands on renderD128 across reboots. AIO mounts the whole /dev/dri directory into containers like memories, so there's no way today to pin hardware transcoding to one GPU. As a result, after a reboot it can silently switch to the wrong device. That's what broke for @ZambeiCloud in production.
> 
> NEXTCLOUD_DRI_DEVICE pins a single render node and remaps it to the canonical /dev/dri/renderD128 inside the container, so transcoding stays on the chosen GPU regardless of host renumbering. Unset, behavior is identical to today.

I see but wouldnt it be possible to simply disable the iGPU in that case which should allow to pin the nvidia gpu to /dev/dri/renderD128?

ecgang:
@szaimen In this case it would be the wrong way round: @ZambeiCloud wants the **iGPU** for transcoding and keeps the Nvidia card for other workloads (#8445). Disabling the iGPU would remove the GPU they actually want to use.

The underlying cause is that the kernel assigns render nodes in driver registration order, so which GPU lands on `renderD128` is not stable across boots. FreeRDP hit the same thing with a hardcoded `renderD128` on hybrid-GPU hosts (FreeRDP/FreeRDP#12779).

There is a host-side workaround — a modprobe `softdep` forcing `i915` to register before `nvidia`. But it is a host-wide kernel change, it does not apply when the drivers are built into the kernel, and it does not address the part that actually broke here: @ZambeiCloud had already pinned the iGPU by path into the mastercontainer, and that worked. AIO then mounted the whole `/dev/dri` directory into the memories container, so that container saw both nodes and used the wrong one.

Unset, `NEXTCLOUD_DRI_DEVICE` changes nothing. If an env var is not the direction you want, the stable identifier for this is `/dev/dri/by-path/pci-…-render`, and I could accept that instead. Tell me which you prefer and I will adjust.

php/src/Docker/DockerActionManager.php
```diff
@@ -337,19 +337,26 @@ public function CreateContainer(Container $container): void {
...
-            }
-            $devices[] = ["PathOnHost" => $device, "PathInContainer" => $device, "CgroupPermissions" => "rwm"];
             if ($device === '/dev/dri') {
+                if (!$this->configurationManager->nextcloudEnableDriDevice) {
+                    continue;
+                }
+                // Allow pinning a single render node on multi-GPU hosts (issue #8445).
+                // The selected host node is remapped to the canonical /dev/dri/renderD128
+                // inside the container so VAAPI/go-vod find it regardless of host numbering.
+                $hostDevice = $this->configurationManager->driDevice; // '/dev/dri' or '/dev/dri/renderD12x'
```
pabzm, line 347: The comment is a little bit misleading, I think, since the host-path could also be `/dev/dri/card*`, etc.

pabzm (changes requested):
Thank you for this clean and well-commented pull request!

I have just a minor nitpick about a comment, the rest looks good to me!

Did you test the changes locally? (I don't have a system to do so.)

Issue #8437, opened 2026-07-10
aio-talk unhealthy — eturnal crash-loops with "Failed to set alternate signal stack" / SIGABRT, seemingly on newer CPUs with AVX-512


Eagle-Admin:
Since deploying the Talk HPB container fresh yesterday, aio-talk never goes healthy. Everything in the container looks fine except eturnal (the bundled TURN server), which crashes on every single startup attempt and gets restarted in a loop by supervisord, forever, a few seconds apart.

**Setup:**
Standalone docker host (not swarm), running as a VM under VMware
OS: Ubuntu Server 24.04 LTS (fully patched)
Host CPU: Intel Xeon 6517P (Granite Rapids)
Image: ghcr.io/nextcloud-releases/aio-talk:20260702_083546 — confirmed this is the current latest/beta tag too, built 2026-06-26
docker info confirms native linux/amd64, no emulation involved
Container ulimits are default (ulimit -s 8192, nothing custom), no custom seccomp, AppArmor is the default docker-default profile, no daemon.json overrides, no systemd drop-ins on the docker service — checked all of this trying to rule out something on my end

**Relevant log lines:**
```
Exec: /opt/eturnal/erts-15.2.6/bin/erlexec -noinput +Bd -boot /opt/eturnal/releases/1.12.2/start -mode embedded -boot_var SYSTEM_LIB_DIR /opt/eturnal/lib -config /opt/eturnal/releases/1.12.2/sys.config -args_file /opt/eturnal/releases/1.12.2/vm.args -erl_epmd_port 3470 -start_epmd false -- foreground
Root: /opt/eturnal
      /opt/eturnal
sys/unix/sys_signal_stack.c:101:sys_sigaltstack(): Internal error: Failed to set alternate signal stack
2026-07-09 21:40:40,922 WARN exited: eturnal (terminated by SIGABRT (core dumped); not expected)
```
This repeats every ~2-3 seconds indefinitely. Janus, the NATS server, and the Go signaling server all start up fine and the signaling server does report listening on [::]:8081 — it's specifically eturnal that never comes up, which I assume is why the healthcheck fails.

[...]

PR #8495, opened 2026-07-29
Add Uptime kuma community container

derStephan:
This is my very first attempt for creating a community container. I will add another PR to integrate that into caddy. 

## Summary

## AI (if applicable)


szaimen (dismissed):
Hi, I am not really sure about this. See https://github.com/nextcloud/all-in-one/discussions/7082#discussioncomment-14966370

derStephan:
I answered there.

pabzm:
@derStephan Thank you for your first PR! 👏 

I think it's a nice idea to integrate a full, standalone monitoring tool like this, and would prefer it over tools like "upptime" that rely on github.com.

My main problem with this is the integration of the docker socket, because it is such a gaping security boundary violation waiting to be exploited. I know some other community containers are using it, too, and I'm unhappy with that, and at least would like to not allow even more containers to use it.

From my perspective this would also be useful without the docker socket. It would require more configuration work, and provide less details automatically, but that's the price for better security protection, I'd say.

@derStephan Do you think you could change the integration to work like that? For Nextcloud there's e.g. `/status.php` as easy monitoring target.

Apart from this fundamental issue I'd have a few smaller remarks, which I'll hold back until we found a common ground.

derStephan:
@pabzm Thank you for your thoughts. 

I generally agree that this can be a security issue. However, it only allows read-only access to the socket. And without that, it is still useful for remote servers but not for the internal health documentation. Glances container is using this socket as well because without it, it will be useless. 

The status.php requires a running nextcloud and will not alert in cases like #8437. 

I can put a fat warning in the readme - as I did for the downtime during backups. Then the actual user can decide.

derStephan:
done.

Zoey2936:
> However, it only allows read-only access to the socket.

Just to let you know, even with a read-only socket it is still possible to send data over it, readonly, only means that the socket can't be deleted

The docker socket is also mounted ro into the masterconatiner: https://github.com/nextcloud/all-in-one/blob/1c14b27090be1c22cc99a24cdd83dadb777e7004/compose.yaml#L10

still the mastercontainer can access the docker API with write access

derStephan:
Ok. I am a bit lost here. @szaimen said, that it is too little related to nextcloud AIO.  Seeing the state of aio containers is the answer to that. 

@pabzm said, that this should be avoided. But then it is too little related to nextcloud AIO. 

I personally can live with or without container access. If this is out of the question, this is fine for me. I can run this thing on my own. I just wanted to give something back.  I see a great benefit in having such thing to monitor other systems - not only for me but for at least some people as the discussion shows. 

How can we proceed? Should I change something? Should I close my PR? I can live with anything. No hard feelings. Promise.

szaimen:
Hey @derStephan so I just reflected about this PR again and I would be fine to include this if you remove the docker socket from the container. Additionally, please fix the red CI.

szaimen:
Thanks @derStephan! Unfortunately DCO is red now. Please fix it. See https://github.com/nextcloud/all-in-one/pull/8495/checks?check_run_id=102763795278

community-containers/uptime-kuma/readme.md
```diff
@@ -0,0 +1,40 @@
...
+[Linkboard App](https://apps.nextcloud.com/apps/linkboard) does not provide real-time data but can monitor remote servers by pinging https:// addresses. It has a historic view of those pings over 7 days. If this is fine for you, then this is your tool.
+
+Uptime Kuma is the right choice if you want to have more than that, e.g.
+- unlimited historic data
+- publicly accessible status pages
+- monitoring of DNS/certificate/port/database/SMTP/...
+- get notifications to talk/mail/signal/matrix/...
+- faster checks than once per minute
+- ...
+
```
szaimen, line 40: Also please add the link to the github repo and a link to you as maintainer as done with all the other CCs

PR #8523, opened 2026-08-02
Fix dockerlint

Zoey2936:
- Resolves: # 

## Summary

## AI (if applicable)


Containers/notify-push/Dockerfile
```diff
@@ -15,12 +16,15 @@ RUN set -ex; \
 # Give root a random password
     echo "root:$(openssl rand -base64 12)" | chpasswd; \
     apk del --no-cache \
-        openssl;
+        openssl; \
+    delgroup www-data; \
+    addgroup -g 33 -S www-data; \
+    adduser -u 33 -S -D -G www-data -H -s /sbin/nologin www-data;
```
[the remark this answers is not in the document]
Zoey2936: the group exists by default, comment makes no sense

.github/workflows/docker-lint.yml
```diff
@@ -34,13 +34,12 @@ jobs:
...
+          for file in $(find -name "*Dockerfile"); do
             # DL3018 warning: Pin versions in apk add. Instead of `apk add <package>` use `apk add <package>=<version>`
-            # DL4006 warning: Set the SHELL option -o pipefail before RUN with a pipe in it. If you are using /bin/sh in an alpine image or if your shell is symlinked to busybox then consider explicitly setting your SHELL to /bin/ash, or disable this check
-            hadolint "$file" --ignore DL3018 --ignore DL4006 | tee -a ./hadolint.log
+            # DL3041 warning: Specify version with `dnf install -y <package>-<version>`.
+            # DL3066 info: Non-numeric user-id may not be resolvable by host system
+            hadolint "$file" --ignore DL3018 --ignore DL3041 --ignore DL3066 | tee -a hadolint.log
           done
-          if grep -q "DL[0-9]\+\|SC[0-9]\+" ./hadolint.log; then
+          if [ -n "$(cat hadolint.log)" ]; then
```
Zoey2936:
```suggestion
          if [ -s hadolint.log ]; then
```

.github/workflows/docker-lint.yml
```diff
@@ -34,13 +34,12 @@ jobs:
 
       - name: run lint
         run: |
-          DOCKERFILES="$(find ./Containers -name Dockerfile)"
-          mapfile -t DOCKERFILES <<< "$DOCKERFILES"
-          for file in "${DOCKERFILES[@]}"; do
+          for file in $(find -name "*Dockerfile"); do
```
Zoey2936:
```suggestion
          for file in $(find Containers -name "*Dockerfile"); do
```

szaimen:
Hey, thanks a lot for the PR! I'll have a look soon! :)

szaimen:
@Zoey2936 thanks a lot for the PR! Can you please split the change of line endings into a different PR so that we can focus with this one on fixing the dockerlint? Thanks in advance! :)

Zoey2936:
> @Zoey2936 thanks a lot for the PR! Can you please split the change of line endings into a different PR so that we can focus with this one on fixing the dockerlint? Thanks in advance! :)

done here and #8542

szaimen:
Please rebase this branch

Zoey2936:
done

szaimen:
Please rebase instead of merging in this case

Zoey2936:
done

.github/workflows/docker-lint.yml
```diff
@@ -34,13 +34,12 @@ jobs:
...
         run: |
-          DOCKERFILES="$(find ./Containers -name Dockerfile)"
-          mapfile -t DOCKERFILES <<< "$DOCKERFILES"
-          for file in "${DOCKERFILES[@]}"; do
+          for file in $(find Containers -name "*Dockerfile"); do
             # DL3018 warning: Pin versions in apk add. Instead of `apk add <package>` use `apk add <package>=<version>`
-            # DL4006 warning: Set the SHELL option -o pipefail before RUN with a pipe in it. If you are using /bin/sh in an alpine image or if your shell is symlinked to busybox then consider explicitly setting your SHELL to /bin/ash, or disable this check
-            hadolint "$file" --ignore DL3018 --ignore DL4006 | tee -a ./hadolint.log
+            # DL3041 warning: Specify version with `dnf install -y <package>-<version>`.
+            # DL3066 info: Non-numeric user-id may not be resolvable by host system
```
pabzm, line 40: Do we still need this, despite the user-IDs being replaced by usernames in this same PR?

.github/workflows/docker-lint.yml
```diff
@@ -34,13 +34,12 @@ jobs:
...
+            hadolint "$file" --ignore DL3018 --ignore DL3041 --ignore DL3066 | tee -a hadolint.log
           done
-          if grep -q "DL[0-9]\+\|SC[0-9]\+" ./hadolint.log; then
+          if [ -s hadolint.log ]; then
```
pabzm, line 43: This isn't quite the same. Any output of executing hadolint (even coming from the shell, not even hadolint itself) would now abort the further execution of the script.

Can you explain why you changed this?

Containers/apache/Dockerfile
```diff
@@ -14,6 +14,7 @@ COPY --chmod=775 healthcheck.sh /healthcheck.sh
 
 VOLUME /mnt/data
 
+SHELL ["/bin/ash", "-o", "pipefail", "-c"]
```
pabzm, line 17: Why `ash` instead of the default `sh`?

Containers/nextcloud/Dockerfile
```diff
@@ -9,6 +9,7 @@ ENV REDIS_DB_INDEX=0
 
 # AIO settings start # Do not remove or change this line!
 ENV NEXTCLOUD_VERSION=33.0.7
+# hadolint ignore=DL3064
```
pabzm, line 12: Please add a comment what this is required for.

pabzm (changes requested):
Thank you for the PR! I've got a few questions and one request for a comment.

Containers/apache/Dockerfile
Zoey2936: https://github.com/hadolint/hadolint/wiki/DL4006

.github/workflows/docker-lint.yml
Zoey2936: because hadolint only prints to console if it finds issues and output/no-outpout is easier to detect than reading the file for a regex which could break easier

.github/workflows/docker-lint.yml
Zoey2936: yes since the dockerlint rules wants to use the id numbers instead of usernames, but I think using IDs will break faster then usernames

Containers/nextcloud/Dockerfile
Zoey2936: https://github.com/hadolint/hadolint/wiki/DL3064 I'm cutrently on mobile, can you please create a suggestion?

.github/workflows/docker-lint.yml
pabzm: Is the switch to usernames instead of UIDs motivated by the assumption that they might break faster? Or by something else?

I don't understand why we should go against the good practice here, if we already have it, and then need to silence the linter.

.github/workflows/docker-lint.yml
pabzm: Ok, thank you! I'd like to read that explanation as a comment in the code, since without it, to me the new code feels "lazy".

Containers/apache/Dockerfile
pabzm: Thank you. I'd prefer to have it mentioned in a comment, that this explicit shell choice is necessary for the combination of Alpine and hadolint only.

Containers/nextcloud/Dockerfile
pabzm: > Required to stop hadolint from complaining about potentially sensitive data in an environment variable. It might be a good idea to reconsider using and env var for this at one point, but the actual value here is just a dummy and not relevant, so we silence the warning for now.

.github/workflows/docker-lint.yml
Zoey2936: currently we have both, IDs and Names, this could stay like it, but from my opinion I think a username is stabler than IDs since user accounts created with `useradd`/`adduser` and not id set as command arg could get a different id anytime, while a username change would need active changes.

.github/workflows/docker-lint.yml
Zoey2936: done

Containers/nextcloud/Dockerfile
Zoey2936: added

Containers/apache/Dockerfile
Zoey2936: is it enough to just mention the github wiki page as comment?

szaimen (approved):
LGTM, thanks!

szaimen:

PR #8533, opened 2026-08-03
replace supervisor with dinit

Zoey2936:
- Resolves: #8522

## Summary

## AI (if applicable)


Containers/nextcloud/dinit.d/cron
```diff
@@ -0,0 +1,6 @@
+type = process
+command = /cron.sh
+run-as = www-data
+term-signal = TERM
+restart = yes
+options = shares-console signal-process-only
```
Zoey2936, line 6: `signal-process-only` should keep the `TRAP` inside the cron.sh working

pabzm:
I don't like this very much, since it basically locks us to Alpine as base image. Debian, Ubuntu, Centos/RHEL, etc. all don't support dinit as init system and don't provide packages for it, as far as I can tell. Please correct me if that's wrong.

And reducing the images by 25-30MB in size is nice, but I doubt that it's worth replacing an established and proven central building block we have experience with by something we have no experience with (AFAIK) and generally rather little known and little used. (Please also correct me here, if the "little known and little used" is only due to my limited personal view).

Sorry to argue about this after you put the effort into this PR, I wasn't aware of it until yesterday.

Zoey2936:
Yes it is not packaged by debian etc. But I don't see how this is an issue. AIO currently uses alpine where possible, so why should it be from instrest if other distros package it? The whole container system is focused on alpine and from what I see, the usage of alpine in containers is still growing and I don't think this will change. Also dinit is so small, it took 30 seconds to build and package with tests if distros are switched: https://build.alpinelinux.org/buildlogs/build-edge-x86_64/community/dinit/dinit-0.21.0-r0.log, supervisor is just a few seconds faster. Replacing alpine with a diffrent distro would currently already mean rewriting nearly all Dockerfiles and adjusting the start scripts, moving the debian images to alpine a few years ago was nothing else.

Also it is self contained, so CVEs in dependencies or python itself don't matter. And it will still work if python includes breaking changes.

And for being little and unused, I would say the opposite dinit is over ten years old, unlike supervisor activly maintained and even used as PID 1 in some distros (chimera, artix and more, yes not the big distros but still a lot of people using it in total including me). Also excluding AIO, I don't know a single project using supervisor.

Zoey2936:
> Sorry to argue about this after you put the effort into this PR, I wasn't aware of it until yesterday.

no problem, this PR didn't took long to write

szaimen (changes requested):
Hi, thanks a lot for the PR @Zoey2936! :)

Generally I like this change and I just let Claude Opus 5 review this PR and here are their findings:

---
Reviewed the dinit migration. Four things, roughly in order of severity:

### 1. `dinit -s <names>` starts nothing when dinit is PID 1 — `Containers/mastercontainer/start.sh:452`

In dinit 0.21, `process_sys_args` is derived from `getpid()==1 && getuid()==0` *before* argv parsing, and `-s` (unlike `-m`/`-o`) never clears it. Every bare service name gets discarded, only the literal word `single` survives, and `services_to_start` falls back to a non-existent `boot` service. So a mastercontainer started **without `--init`** (common in Unraid/Portainer/Synology/Proxmox templates — the readme presents `--init` purely as zombie-reaping) boots zero services and the UI never comes up. Same latent trap in `Containers/nextcloud/Dockerfile:282`, which runs as `USER root:root`.

`-o`/`--container` is the flag that actually clears `process_sys_args`, and it's the semantically correct one here anyway:

```diff
-exec dinit -s php-fpm caddy-internal caddy-acme cron backup-time-file-watcher session-deduplicator domain-validator
+exec dinit --container php-fpm caddy-internal caddy-acme cron backup-time-file-watcher session-deduplicator domain-validator
```

Same change for the nextcloud Dockerfile's dinit invocation. (`-t` before each name works too, but is noisier.)

### 2. Dead config, no graceful drain — `Containers/apache/dinit.d/apache`

`stop-command` on a `process` service replaces signalling entirely, so `term-signal = TERM` never applies. `httpd -k graceful-stop` then waits indefinitely for in-flight requests while dinit's 10s stop-timeout and docker's 10s grace period both SIGKILL it — long uploads are severed anyway. Net effect is a slower shutdown with no drain gained over plain SIGTERM.

Simplest fix is to drop `stop-command` and keep the signal, which is what supervisord effectively did:

```diff
 type = process
 command = httpd -DFOREGROUND
-stop-command = httpd -k graceful-stop
 term-signal = TERM
 restart = yes
 options = shares-console
```

If you do want a real drain, `stop-command` has to stay *and* get room to finish — `stop-timeout = 0` (unlimited) on the service plus a longer `--stop-timeout`/`stopGracePeriod` on the container. Otherwise the current config is the worst of both.

### 3. Log-level switching became one-way — `Containers/mastercontainer/start.sh:351`

```sh
sed -i 's|access.log = /proc/self/fd/1|access.log = /dev/null|' /usr/local/etc/php-fpm.d/docker.conf
sed -i 's|^options = shares-console|#options = shares-console|' /etc/dinit.d/domain-validator
```

Both edits mutate container files with no inverse. The old `SUPERVISORD_STDOUT` indirection was stateless; now a mastercontainer that was started once with `AIO_LOG_LEVEL=warn` keeps logging suppressed even after being restarted with `AIO_LOG_LEVEL=debug`, because the source strings no longer match. `tests/QA/060-environmental-variables.md:10` covers exactly this flow.

Make the branch total so every start rewrites both files to a known state:

```diff
-if [ "$AIO_LOG_LEVEL" != 'debug' ]; then
-    sed -i 's|access.log = /proc/self/fd/1|access.log = /dev/null|' /usr/local/etc/php-fpm.d/docker.conf
-    sed -i 's|^options = shares-console|#options = shares-console|' /etc/dinit.d/domain-validator
-fi
+if [ "$AIO_LOG_LEVEL" = 'debug' ]; then
+    sed -i 's|^access.log = .*|access.log = /proc/self/fd/1|' /usr/local/etc/php-fpm.d/docker.conf
+    sed -i 's|^#\s*options = shares-console|options = shares-console|' /etc/dinit.d/domain-validator
+else
+    sed -i 's|^access.log = .*|access.log = /dev/null|' /usr/local/etc/php-fpm.d/docker.conf
+    sed -i 's|^options = shares-console|#options = shares-console|' /etc/dinit.d/domain-validator
+fi
```

(The patterns are loosened to `^access.log = .*` and `^#\s*options` so they're idempotent from either state — worth confirming `docker.conf` has no second `access.log` line that would now also match.)

### 4. Missing `signal-process-only` — `Containers/nextcloud/dinit.d/run-exec-commands`

`run-exec-commands.sh` ends in the same `trap signal_handler SIGINT SIGTERM; sleep inf & wait $!` shape as `cron.sh`, which d4b302f6 had to fix with exactly this option. Without it dinit signals the whole process group, the `sleep` child dies first, `wait` returns, and the script exits 143 instead of running the trap's `exit 0`.

```diff
-options = shares-console
+options = shares-console signal-process-only
```

Worth a scan of the other service files for the same pattern — anything whose command traps signals and waits on a backgrounded child needs it.

Zoey2936:
For (1): Yes this is true, if tini is not pid 1 this will break. But the solution is not to replace `-s` with `--container`, since without `-s` dinit will not search /etc/dinit.d, but the `.config` folder of the user it is running as (if it does not run as root). If running without tini should be supported then the correct way would be appending `--container`, not replacing `-s`. In that case I would recommend to replace `-s` with `--system --container` or `--container --system`

For (2): This is also true, the question is if TERM or graceful-stop is preffered.

For (3): Also true, but does this matter? Changing the envs requires recreating the container, so I would say it does not matter

For (4): Yes the script uses a TRAP, but the trap just does `exit 0`, so there is no waiting needed. Also trap will still run after sleep is killed

szaimen:
> For (1): [...] `--system --container` or `--container --system`

Sounds good :)

> For (2): This is also true, the question is if TERM or graceful-stop is preffered.

Generally I'd say a graceful stop is preferred :)



> For (3): Also true, but does this matter? Changing the envs requires recreating the container, so I would say it does not matter

True, however I was wondering if we can use envs instead of patching files during runtime?



> For (4): Yes the script uses a TRAP, but the trap just does `exit 0`, so there is no waiting needed. Also trap will still run after sleep is killed

good point 👍

Zoey2936:
> > For (2): This is also true, the question is if TERM or graceful-stop is preffered.
> 
> Generally I'd say a graceful stop is preferred :)

The only issue could be that if apache doesn't gracefully stopped in time (which could be because of any open connection), it will be killed instead of terminated

> > For (3): Also true, but does this matter? Changing the envs requires recreating the container, so I would say it does not matter
> 
> True, however I was wondering if we can use envs instead of patching files during runtime?

From what I checked, sadly no

Zoey2936:
pushed

szaimen (approved):
LGTM. Thanks a lot! ☺️

szaimen:

PR #8545, opened 2026-08-06
enable TURNS over port 443 in talk container

derStephan:
- Needs https://github.com/szaimen/aio-caddy/pull/174
- Resolves: https://github.com/szaimen/aio-caddy/issues/166

## Summary

## AI (if applicable)


pabzm:
Can you please summarise what this does and why? Thanks!

derStephan:
Sure. This PR changes caddy configuration to allow for TLS encrypted video conferencing. This makes changes necessary in Talk Admin settings. 

The goal is to get talk working for clients that are behind deep-packet inspection firewalls which block unencrypted traffic over port 443. 

This makes talk as reliable as zoom and teams and raise security by enabling encryption for video conferences.

Containers/nextcloud/entrypoint.sh
```diff
@@ -950,7 +950,8 @@ if [ "$TALK_ENABLED" = 'yes' ]; then
     # Add turn server
     # shellcheck disable=SC2153
     if ! php /var/www/html/occ talk:turn:list --output="plain" | grep server | grep -q " $TURN_DOMAIN:$TALK_PORT"; then
-        php /var/www/html/occ talk:turn:add turn "$TURN_DOMAIN:$TALK_PORT" "udp,tcp" --secret="$TURN_SECRET"
+        php /var/www/html/occ talk:turn:add turns "$TURN_DOMAIN:$TALK_PORT" "udp,tcp" --secret="$TURN_SECRET"
```
winterrific: From my point of view the protocols should be reversed. Since Caddyl4 is just working on TCP, the tls termination just can take place there. UDP port 443 is bound directly to eturnal which does not have any tls cert for turns. 

turns -> tcp only
turn -> tcp and udp

derStephan: But my tests revealed otherwise and CaddyL4 names capabilities for UDP and TCP. Please have a look at my test results here: 

https://github.com/szaimen/aio-caddy/issues/166#issuecomment-5201358093

TLS seems to be terminated fine over UDP and this can only be done by Caddy L4.

winterrific: This is strange then, since caddy community container is only bound to 443/tcp; see https://github.com/nextcloud/all-in-one/blob/main/community-containers/caddy/caddy.json

```
            "ports": [
                {
                  "ip_binding": "",
                  "port_number": "443",
                  "protocol": "tcp"
                }
            ],
```

When I test it with turns and only tcp it works for me as well. When I change the setting of tuns to udp only, it doesn't.

derStephan: strange indeed. Would you like a test call?

derStephan: did a test. confirmed it works.

Containers/nextcloud/entrypoint.sh
```diff
@@ -950,7 +950,7 @@ if [ "$TALK_ENABLED" = 'yes' ]; then
     # Add turn server
     # shellcheck disable=SC2153
     if ! php /var/www/html/occ talk:turn:list --output="plain" | grep server | grep -q " $TURN_DOMAIN:$TALK_PORT"; then
-        php /var/www/html/occ talk:turn:add turn "$TURN_DOMAIN:$TALK_PORT" "udp,tcp" --secret="$TURN_SECRET"
+        php /var/www/html/occ talk:turn:add turn,turns "$TURN_DOMAIN:$TALK_PORT" "udp,tcp" --secret="$TURN_SECRET"
```
szaimen, line 953: Unfortunately this will break for other installations for turns is not enabled so we would need to find a different solution

szaimen (changes requested):
Hi, thanks a lot for the PR and sorry that it took so long to review it! 

One comment below:

Containers/nextcloud/entrypoint.sh
derStephan: As we simply add TURNS, the clients will try but fall back to TURN in this case. I don't see any other solution. Do you?

derStephan: What do you think? Can you think of a better solution? I am biting my nails here.

derStephan: we could add a occ call in caddy container's start.sh to add turns to the settings. But do we need to handle duplicates?

szaimen: Yes, simply separate the settings into two configurations. One with turn and one with turns. However this needs then some identifier somewhere so that we know shen to use this and not the current one as some instancea might not use the aio-caddy container and would then be left in broken state...

derStephan: But it is not broken if turns is set - even without aio-caddy.

PR #8557, opened 2026-08-08
precompress static files

Zoey2936:
- Resolves: https://github.com/nextcloud/all-in-one/pull/8521#issuecomment-5189898081

## Summary

## AI (if applicable)

After making some size comparisons, the results speak for brotli, it is the smallest when compressing nextcloud and the support in browser is perfect since many many years. So zstd (a little bit bigger on the highest level compared to brotlis highest level and not supported in older browsers) and gzip (bad compression) are not needed.

But I must correct myself in one important point: integrity checks break...
I see two options how this could be fixed:
1. compressed files are saved in a diffrent folder and apache rewrites to this folder if brotli is used, works fine, but skips the check for all precompressed files so they could be modified without the check noticing it
2. resign all files again after compressing them with a selfsigned key

I think option 2 is the best

### DIFF
--- a/Containers/apache/nextcloud.conf
+++ b/Containers/apache/nextcloud.conf
@@ -17,6 +17,28 @@ Listen 8000
     <Proxy "fcgi://${NEXTCLOUD_HOST}:9000" flushpackets=on>
     </Proxy>
 
+    <IfModule mod_rewrite.c>
+        RewriteEngine On
+        RewriteCond %{HTTP:Accept-Encoding} br
+        RewriteCond %{DOCUMENT_ROOT}%{REQUEST_URI}.br -s
+        RewriteRule ^ %{REQUEST_URI}.br [END]
+    </IfModule>
+
+    <IfModule mod_mime.c>
+        AddEncoding br .br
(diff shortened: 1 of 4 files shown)


szaimen:
> After making some size comparisons, the results speak for brotli, it is the smallest when compressing nextcloud and the support in browser is perfect since many many years. So zstd (a little bit bigger on the highest level compared to brotlis highest level and not supported in older browsers) and gzip (bad compression) are not needed.
> 
> But I must correct myself in one important point: integrity checks break... I see two options how this could be fixed:
> 
> 1. compressed files are saved in a diffrent folder and apache rewrites to this folder if brotli is used, works fine, but skips the check for all precompressed files so they could be modified without the check noticing it
> 2. resign all files again after compressing them with a selfsigned key

Honestly, I'd favour Nextcloud Zip archive shipping the already compressed files. 

Can you please create a FR for this here? https://github.com/nextcloud/server/issues 

Thanks in advance! :)

Zoey2936:
see https://github.com/nextcloud/server/issues/63324

PR #8562, opened 2026-08-10
nextcloud: fix ftps support disabled due to missing openssl-dev at ftp configur…

yGhazi-LUSEO:
FTPS mounts fail with `SSL/TLS required on the control channel` even though
the Dockerfile already passes `--with-openssl-dir=/usr` to
`docker-php-ext-configure ftp`. Root cause: `openssl-dev` was not installed
yet when that configure call ran, so PHP's `ftp` extension silently built
without SSL support (`ftp_ssl_connect()` undefined at runtime).

On top of that, `ftp` was being built a **second** time later in the same
`RUN` (in the imap/samba block), without re-passing the openssl flag — this
second build overwrote the first, correctly-configured one.

### Fix
- Added `openssl-dev` to the first `.build-deps` block (where `ftp` is
  configured with `--with-openssl-dir=/usr`)
- Moved `ftp` into the first `docker-php-ext-install` call, right after its
  `docker-php-ext-configure` call
- Removed the duplicate `ftp` install from the second block

### Verification
Before:
```
FTP support => enabled
FTPS support => disabled
```
After:
```
FTP support => enabled
FTPS support => enabled
```

Thanks in advance for reviewing this, happy to adjust based on feedback.

### DIFF
--- a/Containers/nextcloud/Dockerfile
+++ b/Containers/nextcloud/Dockerfile
@@ -60,6 +60,7 @@ RUN set -ex; \
         libxml2-dev \
         libzip-dev \
         openldap-dev \
+        openssl-dev \
         pcre-dev \
         postgresql-dev \
     ; \
@@ -70,6 +71,7 @@ RUN set -ex; \
     docker-php-ext-install -j "$(nproc)" \
         bcmath \
         exif \
+        ftp \
         gd \
         gmp \
         intl \
@@ -205,7 +207,6 @@ RUN set -ex; \
         bz2 \
         imap \
         pgsql \
-        ftp \
     ; \
     pecl install smbclient; \
     docker-php-ext-enable smbclient; \


Containers/nextcloud/Dockerfile
```diff
@@ -70,6 +71,7 @@ RUN set -ex; \
     docker-php-ext-install -j "$(nproc)" \
         bcmath \
         exif \
+        ftp \
```
szaimen, line 74: Hi, why did you put this here and did not leave it further down?

szaimen (commented):
Hi, thanks a lot but one comment below:

Containers/nextcloud/Dockerfile
yGhazi-LUSEO: Hi! I placed it in the first block because it is configured right there), and that block already has openssl-dev in its .build-deps and uses -j "$(nproc)" for parallel compilation.
This is actually my first pull request here, so apologies if my explanation isn't clear or if I got something wrong :)

PR #8563, opened 2026-08-11
localAI: Fix healthcheck pointing to wrong port (8080 instead of 10078)

MichelleAntunes:
Fix healthcheck pointing to wrong port (8080 instead of 10078)'

Problem: The nextcloud-aio-local-ai container is permanently reported as unhealthy, even though the LocalAI service itself works correctly.

Root cause: The base image's built-in HEALTHCHECK reads its target from the HEALTHCHECK_ENDPOINT environment variable, which defaults to http://localhost:8080/readyz. This community container configures LocalAI to listen on port 10078 instead (via LOCALAI_ADDRESS=:10078), but never overrides HEALTHCHECK_ENDPOINT to match — so the healthcheck always probes an empty port and fails.

Fix: Add HEALTHCHECK_ENDPOINT=http://localhost:10078/readyz to the environment array in local-ai.json.

Tested: Manually recreated the container on my own AIO instance with this environment variable added; it now reports healthy, and Context Chat / Nextcloud Assistant continued working normally throughout.

## Summary
- [x] Documentation has been updated or is not required - ("not required")

## AI (if applicable)

### DIFF
--- a/community-containers/local-ai/local-ai.json
+++ b/community-containers/local-ai/local-ai.json
@@ -14,7 +14,8 @@
                 "LOCALAI_ADDRESS=:10078",
                 "LOCALAI_CONFIG_DIR=/configuration",
                 "LOCALAI_MODEL_PATH=/models",
-                "LOCALAI_BACKEND_PATH=/backends"
+                "LOCALAI_BACKEND_PATH=/backends",
+                "HEALTHCHECK_ENDPOINT=http://localhost:10078/readyz"
             ],
             "ports": [
                 {
--- a/community-containers/local-ai/local-ai.json
+++ b/community-containers/local-ai/local-ai.json
@@ -15,7 +15,7 @@
                 "LOCALAI_CONFIG_DIR=/configuration",
                 "LOCALAI_MODEL_PATH=/models",
                 "LOCALAI_BACKEND_PATH=/backends",
-                "HEALTHCHECK_ENDPOINT=http://localhost:10078/readyz"
+                "HEALTHCHECK_ENDPOINT=http://127.0.0.1:10078/readyz"
             ],
             "ports": [
                 {


MichelleAntunes:
Additional context: found a related upstream issue after opening this PR — mudler/LocalAI#10987 (worker-mode containers permanently unhealthy because the healthcheck always probes the default frontend port). The fix there (#10999) made HEALTHCHECK_ENDPOINT empty-by-default and mode-aware via a healthcheck.sh script.

This PR addresses the same root cause — a hardcoded default that doesn't reflect the container's actual runtime configuration — but for the case of a custom port (10078) rather than worker mode. The Nextcloud community container is pinned to quay.io/go-skynet/local-ai:v3.12.1-gpu-vulkan, which predates that upstream fix.

Until the pinned base image is bumped, this explicit override remains necessary.

community-containers/local-ai/local-ai.json
```diff
@@ -14,7 +14,8 @@
                 "LOCALAI_ADDRESS=:10078",
                 "LOCALAI_CONFIG_DIR=/configuration",
                 "LOCALAI_MODEL_PATH=/models",
-                "LOCALAI_BACKEND_PATH=/backends"
+                "LOCALAI_BACKEND_PATH=/backends",
+                "HEALTHCHECK_ENDPOINT=http://localhost:10078/readyz"
```
szaimen, line 18:
```suggestion
                "HEALTHCHECK_ENDPOINT=http://127.0.0.1:10078/readyz"
```

szaimen (commented):
Hi, thanks a lot but one suggestion

szaimen (approved):
LGTM. Thanks!

szaimen:

PR #8597, opened 2026-08-20
Two factor authentication for both login variants

pabzm:
This adds optional two factor authentication based on shared TOTP-secrets to both login variants (password and AIO token authentication).

The 2FA can be enabled as soon as the containers are running in a section below the backup configuration. As long as it is not enabled, the AIO UI shows a warning at the top, and inside the Nextcloud a notification is shown to all Nextcloud admins, strongly recommending to enable it. The Nextcloud admin notification is sent via the mastercontainer's `cron.sh`, thus will be renewed whenever it is dismissed.

In order to show the notices, this PR includes a notification system for the AIO UI, providing two variants: notices, and warnings. Notices have a green background and border and vanish after 5 seconds. Warnings have a orange-leaning yellow background and border and don't vanish (and can't be dismissed manually, neither).

Another visual improvement is highlighted section headlines: If the URL hash matches an `h2` element's ID, the `h2` is highlighted and if the `h2` is followed by a `detail` element, that `detail` is opened. If effect, browsing to `#two-factor-auth` jumps to the section, which is highlighted and opened already (as shown in the second screenshot below).

The libraries are the same as used in https://github.com/nextcloud/twofactor_totp, to stay consistent with Nextcloud.

Screenshots are below.

## Summary
- [x] Tests (playwright if possible) are included
- [x] Screenshots before/after for front-end changes

## AI (if applicable)

## Screenshots

After the containers are up and running:

Screenshot 2026-08-20 at 09 58 25

-------------

After the link in the warning was clicked:

Screenshot 2026-08-20 at 09 58 53

-------------

The same section if the link wasn't clicked but the page was scrolled down:

Screenshot 2026-08-20 at 09 59 16

----------------------

After 2FA was enabled successfully:

Screenshot 2026-08-20 at 10 00 01

----------------------

The section's `detail` has a different summary and changed content now:

Screenshot 2026-08-20 at 10 00 36

-----------------

Logging in with the password (after stopping the containers) requires a 2FA code now:

Screenshot 2026-08-20 at 10 09 11

-------------------------

If logging in via Nextcloud's admin link (AIO token), the 2FA code is required, too:

Screenshot 2026-08-20 at 10 01 16

------------------

Inside Nextcloud a nagging notification is shown to all admins:

Screenshot 2026-08-20 at 11 52 04

### DIFF
--- a/Containers/mastercontainer/cron.sh
+++ b/Containers/mastercontainer/cron.sh
@@ -57,6 +57,9 @@ while true; do
     # Check if AIO is outdated
     su-exec www-data php /var/www/docker-aio/php/src/Cron/OutdatedNotification.php
 
+    # Nag admins to enable two-factor authentication if it is not set up yet
+    su-exec www-data php /var/www/docker-aio/php/src/Cron/TwoFactorAuthNotification.php
+
     # Update deSEC DNS IP record (no-op when IP is unchanged or deSEC is not configured)
     su-exec www-data php /var/www/docker-aio/php/src/Cron/UpdateDesecIp.php
 
--- a/php/composer.json
+++ b/php/composer.json
@@ -17,12 +17,15 @@
 		"slim/twig-view": "^3.3",
 		"slim/csrf": "^1.3",
 		"ext-apcu": "*",
-		"slim/psr7": "^1.8"
+		"slim/psr7": "^1.8",
+		"rullzer/easytotp": "^0.1.4",
+		"christian-riesen/base32": "^1.6"
 	},
 	"require-dev": {
 		"sserbin/twig-linter": "@dev",
 		"vimeo/psalm": "^6.0",
-		"wapmorgan/php-deprecation-detector": "dev-master"
+		"wapmorgan/php-deprecation-detector": "dev-master",
+		"phpunit/phpunit": "^12"
 	},
 	"scripts": {
 		"dev": [
@@ -34,6 +37,7 @@
 		"psalm:strict": "psalm --threads=1 --show-info=true",
 		"lint": "php -l src/*.php src/**/*.php public/index.php",
 		"lint:twig": "twig-linter lint ./templates",
+		"test": "phpunit --configuration phpunit.xml",
 		"php-deprecation-detector": "phpdd scan -n -t 8.5 src/*.php src/**/*.php public/index.php"
 	}
 }
--- a/php/composer.lock
+++ b/php/composer.lock
@@ -4,8 +4,67 @@
         "Read more about it at https://getcomposer.org/doc/01-basic-usage.md#installing-dependencies",
         "This file is @generated automatically"
     ],
-    "content-hash": "e49cef2ac29c2198aececcb842fce2fe",
+    "content-hash": "d97188b5d76a76a38e768193524db6a3",
     "packages": [
+        {
+            "name": "christian-riesen/base32",
+            "version": "1.6.1",
+            "source": {
+                "type": "git",
+                "url": "https://github.com/ChristianRiesen/base32.git",
+                "reference": "03082b946c111eb6ab41a98626fee33360c5232c"
+            },
+            "dist": {
+                "type": "zip",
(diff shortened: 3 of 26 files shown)


pabzm:
The "Docker lint" check is unrelated to the changes of this PR (will be improved or fixed in https://github.com/nextcloud/all-in-one/pull/8598)

szaimen (changes requested):
Hi, thanks a lot for this PR!

Also thanks a lot for working on tests for this!

I was wondering if the phpunit tests should also be tested in CI? (Currently they are not IIRC).

Additionally, the vendor-qrcode.js was included here. shouldnt we add a way to update it automatically?

szaimen (changes requested):
Apart from that, I just let Claude Opus 5.0 review this and here are its findings:

---

Reviewed the full diff against `main` (26 files), including the upstream sources of both new dependencies (`rullzer/easytotp` and `christian-riesen/base32` at the pinned revisions) to check the verify/decode contracts.

## Findings

### 1. Cross-site DoS on passphrase login — `php/src/Controller/LoginController.php:58`

`GetTryLogin` stores an *unvalidated* getlogin token in the session, and `TryLogin`'s pending-token branch then unconditionally preempts the password branch — nothing ever clears it, not even `Logout`.

`/api/auth/getlogin` is a public route, CSRF only guards POST/PUT/DELETE/PATCH, and `SameSite=Lax` sends the cookie on top-level navigation. So a single cross-site link plants a bogus token and permanently locks the admin out of passphrase login for that session (the password field is hidden under `two_factor_auth_only`).

This is the one I'd consider blocking: remotely triggerable and it locks out the admin.

### 2. Disk-full turns every 2FA login into a 500 — `php/src/Controller/LoginController.php:81`

`commitTwoFactorAuth()` writes `configuration.json` *before* `SetAuthState(true)`, and the `InvalidSettingConfigurationException` that `writeConfig()` throws on low disk is uncaught. Elsewhere the codebase deliberately only warns on this condition rather than failing the request.

### 3. TOTP replay window — `php/src/Auth/AuthManager.php:78`

Single-use protection is a read-then-write on `totp_last_counter` with no locking: two concurrent logins submitting the same code both read the stale counter and both authenticate.

### 4. Unvalidated base32 secret — `php/src/Data/ConfigurationManager.php:859`

`totp_secret` arrives from a hidden form field and is never validated as base32. base32 1.6.1's `decode()` *strips* invalid characters rather than rejecting them, so a secret that decodes to an empty HMAC key can be enabled while the UI reports 2FA as protecting the instance.

Also: `^1.6` in `composer.json` permits 1.6.2+, whose `decode()` throws an exception that isn't an `InvalidSettingConfigurationException` → 500.

### 5. False success message — `php/src/Data/ConfigurationManager.php:880`

`disableTwoFactorAuth()` early-returns when 2FA is already disabled, but the controller still emits "has been disabled" — so a blank code yields a success confirmation with nothing verified.

### 6. Missing `strval()` on request params — `php/src/Controller/LoginController.php:21`

`totp` / `totp_secret` / `totp_code` are read from `getParsedBody()` uncast. An array-valued param hits the `strict_types` string params and produces an uncaught `TypeError` (500 instead of 422). The codebase already casts elsewhere.

### 7. Nested transaction (latent) — `php/src/Data/ConfigurationManager.php:867`

`enableTwoFactorAuth`/`disableTwoFactorAuth` open a transaction inside `SetConfig`'s. The inner `commitTransaction()` clears the outer `noWrite` flag and blanks the config cache. Harmless today only because the TOTP blocks happen to run last.

## Checked and found sound

- The ±1 drift / counter-skip interaction with `easytotp`'s `verify()`
- `qrcode(0, "M")` auto-sizing
- CSP allowing `vendor-qrcode.js` under `script-src-elem 'self'`
- The flash-notification 201 → `Location` → reload round-trip consuming the queue
- `notifications.twig`'s `{link}` split staying auto-escaped
- The `readyState` guards in the new JS
- `notify.sh`'s admin-only targeting for the new cron nag

szaimen:
Closing as discussed. We are working on a different fix

PR #8598, opened 2026-08-20
Fix hadolint complaints

pabzm:
Fixes hadolint's complains as visible e.g. in https://github.com/nextcloud/all-in-one/actions/runs/32381178668/job/96464481888

1. Converts all shell-form HEALTHCHECK CMD instructions to exec (JSON
array) form. The domaincheck healthcheck relies on shell expansion and
the || operator, so it is wrapped as CMD ["sh", "-c", "..."] to preserve
behavior.
2. Makes hadolint ignore [DL3064](https://github.com/hadolint/hadolint/wiki/DL3064) in one case, where we assign a dummy value to the `AIO_TOKEN` environment variable. (I'm not sure why we do that but I don't dare to change that now.)

## Summary

## AI (if applicable)

### DIFF
--- a/Containers/apache/Dockerfile
+++ b/Containers/apache/Dockerfile
@@ -102,7 +102,7 @@ CMD ["/usr/bin/supervisord", "-c", "/supervisord.conf"]
 
 ENV AIO_LOG_LEVEL=warn
 
-HEALTHCHECK CMD /healthcheck.sh
+HEALTHCHECK CMD ["/healthcheck.sh"]
 LABEL com.centurylinklabs.watchtower.enable="false" \
     wud.watch="false" \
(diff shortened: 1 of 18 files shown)


pabzm:
The Docker lint check complains:

> DL3066 info: Non-numeric user-id may not be resolvable by host system

I'd prefer to fix this instead of ignoring it, but I'm not sure if using numeric UIDs could break something.

@szaimen What do you think?

Zoey2936:
Why a second PR? https://github.com/nextcloud/all-in-one/pull/8523

pabzm:
Because I forgot about yours and just worked on it spontaneously 

/close

PR #8610, opened 2026-08-23
Allow skipping Borg compacting during backups

cpf-sh:
- Resolves: no issue opened, I don't have issue creation perms!

## Summary
This PR adds a new variable, `BORG_SKIP_COMPACTING`, which allows for AIO users to tell AIO not to run `borg compact`. This is useful for all the reasons that Borg themselves discuss in [the Borg docs on separate compaction](https://borgbackup.readthedocs.io/en/stable/usage/notes.html#separate-compaction), but to specifically mention my use case - I rclone my Borg backups into an archival storage system that has a minimum charge of n days per file anyway, so doing `borg compact` more than once every n days might actually cost me _more_ money (because files could be overwritten, causing them to be charged twice) rather than being left alone :)

I don't think this is an area of the code where it's possible to write automated testing in the current setup, but please do correct me if I'm wrong.

- [X] [Sign-off message](https://github.com/src-d/guide/blob/master/developer-community/fix-DCO.md) is added to all commits
- [X] Documentation has been updated or is not required

## AI (if applicable)


szaimen (changes requested):
Hi, thanks for the PR!

I am unfortunately not sure if we should add this feature as it seems to be a very edge case related to the chosen backup target.

I'll discuss with @pabzm first if it makes sense for the project.

cpf-sh:
No worries. Just a note that this is the Borg default underlying, so if AIO decides not to support it at all, it should probably document that it differs from the default behaviour and does not support the default behaviour being restored.

cpf-sh:
Hi @szaimen - just wanted to follow up and check whether you've made a decision on this yet? Just want to know whether it's worth updating the branch or if I need to seek an alternative solution.

szaimen:
Hi, I discussed this with my colleague and as you seem to be the only one that is currently interested in something like this, I'd like to not merge this as of now. 

It probably makes sense to create a feature request for this here so that we can gather if more people are interested in something like this: https://github.com/nextcloud/all-in-one/discussions/new?category=ideas

PR #8633, opened 2026-08-27
feat: add oom_score_adj and cpu_shares to mastercontainer and sibling containers

szaimen:
- Supercedes and closes #8082 
- Supercedes and closes #8083 
- Supercedes and closes #8084

## Summary
- [x] Or will be tested after merge on a dedicated test instance (available for maintainers)

## AI (if applicable)

### DIFF
--- a/compose.yaml
+++ b/compose.yaml
@@ -5,6 +5,7 @@ services:
     init: true # This setting makes sure that signals from main process inside the container are correctly forwarded to children. See https://docs.docker.com/reference/compose-file/services/#init
     restart: always # This makes sure that the container starts always together with the host OS. See https://docs.docker.com/reference/compose-file/services/#restart
     oom_score_adj: -500 # This makes the Linux OOM killer strongly prefer other processes, keeping the AIO interface reachable under memory pressure. See https://docs.kernel.org/admin-guide/cgroup-v1/memory.html
+    cpu_shares: 2048 # This doubles the default CPU share weighting so the mastercontainer stays responsive under heavy load. See https://docs.docker.com/reference/compose-file/services/#cpu_shares
     container_name: nextcloud-aio-mastercontainer # This line is not allowed to be changed as otherwise AIO will not work correctly
     volumes:
       - nextcloud_aio_mastercontainer:/mnt/docker-aio-config # This line is not allowed to be changed as otherwise the built-in backup solution will not work
--- a/compose.yaml
+++ b/compose.yaml
@@ -4,6 +4,7 @@ services:
     image: ghcr.io/nextcloud-releases/all-in-one:latest # This is the container image used. You can switch to ghcr.io/nextcloud-releases/all-in-one:beta if you want to help testing new releases. See https://github.com/nextcloud/all-in-one#how-to-switch-the-channel
     init: true # This setting makes sure that signals from main process inside the container are correctly forwarded to children. See https://docs.docker.com/reference/compose-file/services/#init
     restart: always # This makes sure that the container starts always together with the host OS. See https://docs.docker.com/reference/compose-file/services/#restart
+    oom_score_adj: -500 # This makes the Linux OOM killer strongly prefer other processes, keeping the AIO interface reachable under memory pressure. See https://docs.kernel.org/admin-guide/cgroup-v1/memory.html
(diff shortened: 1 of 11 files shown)


szaimen:
## Review

The plumbing is correct: `Container.php` is instantiated from exactly one site, all 19 built-in containers declare both keys, the schema bounds them, and the docs update covers all five mastercontainer launch commands. The issues below are in the chosen fallback defaults, which are set to the *lowest* tier used in `containers.json` rather than to Docker-neutral values.

### 1. `oomScoreAdj` fallback of 500 makes undeclared containers the first OOM target

`php/src/ContainerDefinitionFetcher.php:368`

The fallback means every container without the key becomes the most OOM-killable process on the host. All 35 `community-containers/*/*.json` omit it, so Vaultwarden, Plex, Home Assistant etc. go from the host default of 0 to first-to-die under memory pressure. Nothing documents the new key for community-container authors either.

Suggest a sentinel (e.g. `null`/unset) so the field is simply omitted when undeclared.

### 2. `cpuShares` fallback of 512 halves CPU weight for undeclared containers

`php/src/ContainerDefinitionFetcher.php:321`

512 is half Docker's default weight of 1024, so the same 35 community containers (`local-ai`, `memories`, `azuracast`, …) silently lose half their CPU share on a saturated host.

`shm_size` in the same function already does this correctly with a `-1` sentinel, and the existing `if ($cpuShares > 0)` guard handles omission — so the fix is small.

### 3. Negative `oom_score_adj` will fail or be silently clamped under rootless Docker

`php/src/Docker/DockerActionManager.php:422`

Rootless Docker is officially supported (`docker-rootless.md`), but there dockerd/runc run without `CAP_SYS_RESOURCE` and cannot lower a child's `oom_score_adj`. The five containers at -100/-300 hit this on every `CreateContainer`: depending on engine version it's either an opaque container-creation failure with no fallback, or a silent clamp — meaning the advertised protection doesn't actually exist. The same applies to the documented `--oom-score-adj -500` for the mastercontainer.

Worth either detecting rootless and skipping the negative values, or documenting the limitation.

### 4. `!== 0` guard conflates an explicit `0` with "unset"

`php/src/Docker/DockerActionManager.php:422`

The schema permits `0` in its -500..500 range, but the guard drops it. Combined with finding 1, there's no way to opt out of the 500 default: omitting the key inherits it, and setting `0` explicitly is ignored.

### 5. `manual-install/latest.yml` not regenerated

`manual-install/latest.yml`

It's generated from `containers.json` by `update-yaml.sh` (which doesn't `del()` either new key), but wasn't regenerated here — so it lacks both settings while `shm_size`/`read_only`/`init` are present. This self-heals via the daily `update-yaml.yml` cron; same for the helm chart. Non-blocking, just noting it.

### Nits

- In `readme.md` and the other docs, `--oom-score-adj` was inserted *before* the `--publish` flags while `--cpu-shares` was inserted *after* them, which splits the three publish bullets in the explanation list.
- Siblings are deleted and recreated on each AIO start, so they pick up the new settings automatically — but the mastercontainer does not. That requires a manual recreate, which isn't mentioned anywhere in the PR or docs.

derStephan:
I would like to repeat what I stated in the other PR. The Talk container - especially the TURN component in it - is a real CPU eater but for good reason. It makes video conferencing for the most users possible in the first place. So this is a very important feature of talk. 
But this is a feature with real-time requirements. So lowering cpu shares for talk to 512 will make video conferencing worse in heavy load situations. 
From my point of view, cpu priority for the talk container should be raised to the very top. Talk - together with caddy - is the only (!) container with real-time needs. Everything else can just wait.

Zoey2936:
Also some things:
- Will this break rootless mode? From my understanding setting these values requires root
- Same issue still vor non-systemd hosts: they don't set this option https://github.com/moby/moby/blob/5062ed7823ba3db32c939a7fd93e0454568a93b9/contrib/init/systemd/docker.service#L35

szaimen:
> Also some things:
> 
> * Will this break rootless mode? From my understanding setting these values requires root
> * Same issue still vor non-systemd hosts: they don't set this option https://github.com/moby/moby/blob/5062ed7823ba3db32c939a7fd93e0454568a93b9/contrib/init/systemd/docker.service#L35

Indeed, that is what https://github.com/nextcloud/all-in-one/pull/8633#issuecomment-5436268268 reported as well. So I guess this PR is stuck for  now.

PR #8642, opened 2026-08-30
eurooffice: persist sdkjs-plugins as a named volume

### LINKED ISSUE #8616: Euro-Office needs persistent volume for plugins

Currently, there is no way to install plugins, so that they survive deletion and re-creation of the container

A persistant volume should be provided where plugins can be installed. (var/www/euro-office/documentserver/sdkjs-plugins )

vasyugan:
Fixes #8616 by creating persistent volume for Euro-Office plugins

- Resolves: #8616

## Summary

The EuroOffice DocumentServer image ships the `sdkjs-plugins` directory
read-only (`dr-xr-xr-x`, owned by `ds:ds`), and it isn't backed by a
volume. Any plugin installed into that directory is lost whenever the
eurooffice container is recreated (e.g. on update).

This PR:
- Adds a named volume (`nextcloud_aio_eurooffice_plugins`) mounted into
  `/var/www/euro-office/documentserver/sdkjs-plugins` in
  `php/containers.json`, following the same pattern as the existing
  `nextcloud_aio_eurooffice_data` volume added in #8512.
- Fixes ownership/permissions for that directory in the Dockerfile so
  the `ds` user (which the document server process actually runs as)
  can write to it. Mounting the volume alone is not sufficient, since
  Docker preserves the read-only permissions of the original directory
  when initializing a new named volume.

### Testing performed
Built the eurooffice image locally and tested outside of AIO's
orchestration, directly against the named volume:
1. Created the named volume and started the built image with it mounted.
2. Confirmed the `ds` user can write to the directory (`touch` succeeds,
   `ls -la` shows `drwxr-xr-x` instead of `dr-xr-xr-x`).
3. Removed and recreated the container against the same volume
   (`docker rm` + `docker run`, simulating an AIO container update) and
   confirmed the test file written in step 2 survived.

## AI (if applicable)
- [x] the creation of this pull request was aided by AI. 
Used Claude to help identify the fix (following the pattern from #8512)
and draft the containers.json/Dockerfile changes. I reviewed and
verified all code myself, and performed all testing (local image
build, volume mount, permission check, container recreation test)
manually on my own machine — none of that was run by the AI.

### DIFF
--- a/Containers/eurooffice/Dockerfile
+++ b/Containers/eurooffice/Dockerfile
@@ -4,6 +4,10 @@ FROM ghcr.io/euro-office/documentserver:v9.3.3
 # USER root is probably used
 
 COPY --chmod=775 healthcheck.sh /healthcheck.sh
+# Make sdkjs-plugins writable so it can be mounted as a persistent named
+# volume (upstream image ships it read-only, dr-xr-xr-x, owned by ds:ds)
+RUN chown ds:ds /var/www/euro-office/documentserver/sdkjs-plugins \
+    && chmod 755 /var/www/euro-office/documentserver/sdkjs-plugins
 
 HEALTHCHECK --start-period=60s --retries=9 CMD ["/healthcheck.sh"]
 LABEL com.centurylinklabs.watchtower.enable="false" \
--- a/php/containers.json
+++ b/php/containers.json
@@ -796,6 +796,11 @@
           "source": "nextcloud_aio_eurooffice_data",
           "destination": "/var/www/euro-office/Data",
           "writeable": true
+        },
+  	{
+	  "source": "nextcloud_aio_eurooffice_plugins",
+          "destination": "/var/www/euro-office/documentserver/sdkjs-plugins",
+          "writeable": true
         }
       ],
       "secrets": [


szaimen (changes requested):
Hi, thanks for the PR!

I let Claude Opus 5.0 review this PR and here is the output:

Mounting a named volume over `sdkjs-plugins` will freeze the image's built-in
plugins at whatever version was present when the volume was first created.

`sdkjs-plugins` isn't a data directory — it's the code directory that ships the
built-in plugins. EuroOffice is a fork of ONLYOFFICE Document Server and keeps
the same layout (just under `/var/www/euro-office/`), so the upstream docs apply:

- [Adding new plugins to server editors](https://helpcenter.onlyoffice.com/docs/installation/docs-add-plugin.aspx)
  documents `sdkjs-plugins` as the directory *inside the installation* that you
  drop plugin folders into.
- [ONLYOFFICE Docs on-premises](https://api.onlyoffice.com/docs/plugin-and-macros/tutorials/installing/onlyoffice-docs-on-premises/)
  notes that "after app updating the default plugins will appear again" — i.e.
  upstream expects this directory to be *rewritten by the image on update*.
- The upstream community thread
  [Installing plugins in docker](https://community.onlyoffice.com/t/installing-plugins-in-docker/848)
  hits exactly this: "the existing content of that directory as it is shipped
  with the docker image will be obscured, as the docker documentation says."
  The suggested workaround there is to manually re-download all the default
  plugins into the mount — which tells you this path isn't meant to be mounted over.

The mechanics: Docker seeds a named volume from the image only on *first*
creation, while the volume is still empty. On every subsequent eurooffice image
update the volume is non-empty, so nothing is re-seeded and the stale plugins
shadow the ones in the new image. Built-in plugin fixes and additions would never
reach existing users, and there's no upgrade path short of users manually
deleting the volume.

I don't think `nextcloud_aio_eurooffice_data` is a comparable precedent here:
`/var/www/euro-office/Data` holds generated keys and user-supplied fonts —
content the image never ships and never updates. `sdkjs-plugins` is the opposite:
image-owned content that we *want* replaced on every update.

If the goal is persisting user-installed plugins, could we either mount the
volume at a subdirectory (e.g. `sdkjs-plugins/<name>`, as upstream's own docker
example does) rather than over the whole directory?

vasyugan:
>     * The upstream community thread
>       [Installing plugins in docker](https://community.onlyoffice.com/t/installing-plugins-in-docker/848)
>       hits exactly this: "the existing content of that directory as it is shipped
>       with the docker image will be obscured, as the docker documentation says."
>       The suggested workaround there is to manually re-download all the default
>       plugins into the mount — which tells you this path isn't meant to be mounted over. s own docker example does) rather than over the whole directory?

Hi, thanks for looking into this. Would have been so nice.  Re-downloading plugins each time seems to presume some sort of package management, which, last time I check didn't exist, I had to download and install everything manually. So I have no idea how this would work.

I guess, one would need two separate directories, one for default plugins and one for custom plugins, like with the Nextcloud container itself. (although, currently /var/www/euro-office/documentserver/sdkjs-plugins is empty)

PR #8685, opened 2026-09-11
Rename "Nextcloud Office (Collabora)" by "Collabora Online"

chrip:
## Summary

Updates the office suite naming in the AIO interface so the two bundled
office suites are no longer both called "Nextcloud Office".

The Collabora container and its chooser card become **Collabora Online**
("by Collabora Productivity", matching the ONLYOFFICE card's "by Ascensio
System SIA"). The Euro-Office container's display name becomes **Nextcloud
Office**. ONLYOFFICE is untouched.

This is the AIO half of a coordinated rename. The other half is
nextcloud/richdocuments#6074.

## Why

AIO ships two office suites and, before this change, labelled both of them
"Nextcloud Office" — the Collabora container in `containers.json` and the
Euro-Office chooser card. That made the suite picker ambiguous at the exact
point where a user has to choose between them.

The two halves are at different stages:

- **Euro-Office** already ships as "Nextcloud Office" — `appinfo/info.xml`
  on `Euro-Office/eurooffice-nextcloud` `main` says so today. AIO's label is
  catching up to the app it installs, not leading it.
- **Collabora** does not yet. `nextcloud/richdocuments` `main` still says
  "Nextcloud Office (Collabora)"; the rename lives in the open PR
  nextcloud/richdocuments#6074.

### Merge ordering

If this merges before nextcloud/richdocuments#6074, the AIO chooser will say
"Collabora Online" while the app it installs still identifies itself as
"Nextcloud Office (Collabora)" inside Nextcloud, until the next richdocuments
release reaches users.

I consider that window acceptable — it is cosmetic, self-correcting, and
strictly less confusing than the current state where both suites share one
name — but it is a deliberate choice rather than an oversight, so flagging it
here. Happy to hold this until richdocuments#6074 lands if maintainers prefer.

## Testing

- [ ] Office suite chooser: all three cards show the correct titles and vendor lines
- [ ] Titles, checkmarks and feature lists line up across all three cards
- [ ] Collabora and ONLYOFFICE cards render unchanged from `main`
- [ ] Chooser at narrow viewport widths, including where the grid collapses to one column
- [ ] Container list shows the new display names
- [ ] Dictionaries / additional options section with Collabora selected
- [ ] Switching between office suites still works and only one stays selectable

Assisted-by: ClaudeCode:opus-5

### DIFF
--- a/community-containers/languagetool/readme.md
+++ b/community-containers/languagetool/readme.md
@@ -1,9 +1,9 @@
-## LanguageTool for Nextcloud Office
-This container bundles [LanguageTool](https://github.com/languagetool-org/languagetool) for Nextcloud Office which adds spell checking functionality to Nextcloud Office.
+## LanguageTool for Collabora Online
+This container bundles [LanguageTool](https://github.com/languagetool-org/languagetool) for Collabora Online which adds spell checking functionality to Collabora Online.
 
 ### Notes
-- Make sure to have Nextcloud Office enabled via the AIO interface
-- After adding this container via the AIO Interface, while all containers are still stopped, you need to scroll down to the `Additional Nextcloud Office options` section and enter `--o:languagetool.enabled=true --o:languagetool.base_url=http://nextcloud-aio-languagetool:8010/v2`.
+- Make sure to have Collabora Online enabled via the AIO interface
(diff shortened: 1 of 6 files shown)


szaimen (changes requested):
Hi, lets wait for https://github.com/nextcloud/richdocuments/pull/6074 first

